PR #25, opened 2015-01-16
Inner LightCouch classes moved to another package + some small changes

rwitzel:


mariobriggs:
@rwitzel can you comment on the reasons for 'moved : classes of lightcouch driver to cloudant-specific package'. 
Personally i want to remove some of the cloudant wrappers for respective lightcouch model classes... becuase of Database.bulk() and Replicator.findAll() since it is very inefficient to be making some many copies of wrapper objects. Then i would simply copy across the implementation of those methods from lightcouch and remove the redirect. That seems more beneficial to do.

rwitzel:
The LightCouch classes of java-cloudant caused conflicts with classes of the original LightCouch driver in my classpath. It should be strictly avoided to add third party classes to own artifacts because this can cause severe problems due to duplicate classes in different versions in the classpath. This is why I have moved the classes to another package.

For me it is difficult to comment on your suggested refactoring because I have not looked at the code in detail so that I don't know what behaviour ist added by Java-Cloudant.

But I assume that (1) Cloudant databases provides additional APIs, and (2) original CouchDB APIs are not changed. Therefore, as a developer I would use LightCouch for all CouchDB APIs und a Cloudant class for Cloudant-specific APIs. Therefore, I did not expect any LightCouch code in java-cloudant. But as I said, I have not looked at the code in detail.

All that said, copying code is like forking and this should be avoided.

mariobriggs:
@rwitzel thanks for throwing light on the problem you are going after (classpath conflicts with original lightcouch). That makes sense and I agree that is a problem. One of the reasons we kept our copy for now was because we additionally refactored the lightcouch one, separating the 'Server' functionality versus the 'Database' functionality. But with the problem you highlight, I agreed we need to address it. thanks once again

ricellis:
This is superseded by #84 which moves the lightcouch classes to a cloudant specific package for the current master.
Please note the following regarding the other small changes:
* 079d284 - the source level is currently Java 1.6 so the `@Override` annotations remain on interface method implementations, but the 1.7 `<>` diamonds were already replaced in 4ec960b94835e79d36bcb9216c8677704c206d83
* 816e153 - We’re working to avoid exposing LightCouch in our APIs, as we’re hoping to reduce the duplication of code and classes between the LC and Cloudant packages over time so I haven't incorporated this change.
* 708581a - the maven pom has been removed and has been generated automatically using the version number in `build.gradle` since 5e9d19711b8549c537d7fd6a0fa086803c204c0f

PR #38, opened 2015-02-17
test: rename the test properties to sample names

srl295:
Following the advice in README.md, we don't want to accidentally
check in the test properties file with account credentials!

Renamed the files to "sample" names and updated the docs.


mariobriggs:
thanks. this is a nice one, appreciate it

PR #48, opened 2015-05-12
46770 character encoding bug

brynh:
Fixes FB#46770.

This fixes the specific bug identified by the test case included in the bug report: Instantiating an InputStreamReader using the platform's default charset can lead to corruption of non ASCII characters. The fix forces the InputStreamReader to use UTF-8. 

Additionally, all other usages of InputStreamReader have been changed to use UTF-8 instead of the platform default charset.

Note that UTF-8 is required to be supported by every implementation of the Java platform - see under "Standard charsets" here:
http://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html

*TESTING*
The test case from the bug report has been modified to fit the style used by the existing tests and incorporated into the test suite.  It was confirmed using the modified test that prior to the code fixes the test failed when the platform's default charset was set to use ISO08859-1 and that once fixed the test passes.

reviewer @mikerhodes
reviewer @rhyshort

### DIFF
--- a/src/main/java/com/cloudant/client/api/Database.java
+++ b/src/main/java/com/cloudant/client/api/Database.java
@@ -14,6 +14,7 @@ import static org.lightcouch.internal.URIBuilder.buildUri;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
 import java.net.URI;
 import java.util.ArrayList;
@@ -235,7 +236,7 @@ public class Database {
 		 InputStream stream = null; 
 		 try {
 			 stream = getStream(client.executeRequest(createPost(uri, body, "application/json")));
-			 Reader reader = new InputStreamReader(stream);
+			 Reader reader = new InputStreamReader(stream, "UTF-8");
 			 JsonArray jsonArray = new JsonParser().parse(reader)
 						.getAsJsonObject().getAsJsonArray("docs");
 			List<T> list = new ArrayList<T>();
@@ -245,6 +246,9 @@ public class Database {
 				list.add(t);
 			}
 			return list;
+		 } catch (UnsupportedEncodingException e) {
+			// This should never happen as every implementation of the java platform is required to support UTF-8.
+			throw new RuntimeException(e);
 		 }
 		 finally {
 			 close(stream);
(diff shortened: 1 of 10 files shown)


src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -245,6 +246,9 @@ public void createIndex(String indexDefinition) {
 				list.add(t);
 			}
 			return list;
+		 } catch (UnsupportedEncodingException e) {
+			// This should never happen as every implementation of the java platform is required to support UTF-8.
+			return null;
```
rhyshort: Just for discussion: I'm wondering if we should just throw a RuntimeException since we are sure it is never going to happen on java platforms, what are your thoughts @brynh

src/test/java/com/cloudant/tests/UnicodeTest.java
```diff
@@ -0,0 +1,347 @@
...
+
+	// According to JSON (ECMA-404, section 9 "Strings"):
+	// - All Unicode characters except those that must be escaped
+	//   (U+0000..U+001F, U+0022, U+005C) may be placed in a string.
+	// - All Unicode characters may be included as Unicode escapes
+	//   (after conversion to UTF-16).
+
+	public static final String TESTSTRING = "Gr\u00fc\u00dfe \u65e5\u672c\u8a9e \uD834\uDD1E.";
+	public static final String TESTSTRING_ESCAPED = "Gr\\u00fc\\u00dfe \\u65e5\\u672c\\u8a9e \\uD834\\uDD1E.";
+
```
rhyshort: These constants should be at the top of the class definition

brynh: Addressed in b81a678. Also made them private rather than public.

src/main/java/com/cloudant/client/api/Database.java
brynh: On balance I think it's probably a good idea. I was going on the basis that as it should never happen it doesn't really matter. However, it's probably a bit cleaner to throw a RuntimeException and would give us a chance of identifying the issue from a stack trace if it ever did happen on a non-compliant/buggy platform.

rhyshort: Okay, so what I think I'd do then, is keep the comments about the exception should never be thrown and just have `throw new RuntimeException(e)` instead of `return null`

brynh: Agreed. Addressed in b8ec44a.

rhyshort:
Looks good to me +1 :+1:

mikerhodes:
This looks good. The changes from b81a678 and b8ec44a need merging appropriately into earlier commits before merging.

mikerhodes:
:+1: Looking good.

PR #49, opened 2015-05-15
46986 queries do not encode +

brynh:
Fixes FB#46986.

Ensures the characters `+`, `=` and `&` are correctly percent encoded when they appear in the query part of a URL.

Replacement of the characters is done in `URIBuilder.build`

*TESTING*
Tests added to ensure correct percent encoding of `+`, `=` and `&`.

reviewer @mikerhodes
reviewer @tomblench

### DIFF
--- a/src/main/java/org/lightcouch/internal/URIBuilder.java
+++ b/src/main/java/org/lightcouch/internal/URIBuilder.java
@@ -32,98 +32,98 @@ import org.lightcouch.Params;
  * 
  */
 public class URIBuilder {
-	private String scheme;
-	private String host;
-	private int port;
-	private String path = "";
-	private String pathToEncode = "";
-	/* The final query */
-	private final StringBuilder query = new StringBuilder();
-	/* key=value params */
-	private final List<String> qParams = new ArrayList<String>();
+    private String scheme;
+    private String host;
+    private int port;
+    private String path = "";
+    private String pathToEncode = "";
+    /* The final query */
+    private final StringBuilder query = new StringBuilder();
+    /* key=value params */
+    private final List<String> qParams = new ArrayList<String>();
 
-	public static URIBuilder buildUri() {
-		return new URIBuilder();
-	}
+    public static URIBuilder buildUri() {
+        return new URIBuilder();
+    }
 
-	public static URIBuilder buildUri(URI uri) {
-		URIBuilder builder = URIBuilder.buildUri().scheme(uri.getScheme())
-				.host(uri.getHost()).port(uri.getPort()).path(uri.getPath());
-		return builder;
-	}
+    public static URIBuilder buildUri(URI uri) {
+        URIBuilder builder = URIBuilder.buildUri().scheme(uri.getScheme())
+                .host(uri.getHost()).port(uri.getPort()).path(uri.getPath());
+        return builder;
+    }
 
-	public URI build() {
-		try {
-			for (int i = 0; i < qParams.size(); i++) {
-				String amp = (i != qParams.size() - 1) ? "&" : "";
-				query.append(qParams.get(i) + amp);
-			}
-			String q = (query.length() == 0) ? null : query.toString();
-			return new URI(scheme, null, host, port, path, q, null);
-		} catch (URISyntaxException e) {
-			throw new IllegalArgumentException(e);
-		}
-	}
-	
-	public URI buildEncoded() {
-		for (int i = 0; i < qParams.size(); i++) {
-			String amp = (i != qParams.size() - 1) ? "&" : "";
-			query.append(qParams.get(i) + amp);
-		}
-		try {
-			String q = (query.length() == 0) ? "" : "?" + query;
-			String uri = String.format("%s://%s:%s%s%s%s", new Object[]{scheme, host, port, path, pathToEncode, q});
-			return new URI(uri);
-		} catch (URISyntaxException e) {
-			throw new IllegalArgumentException(e);
-		} 
-	}
+    public URI build() {
+        try {
+            for (int i = 0; i < qParams.size(); i++) {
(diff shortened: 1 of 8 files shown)


tomblench:
This uses hard tabs for indentation, which is an Eclipse default and seems to have been the standard for this library so far.

Can you convert the classes you have modified to use 4 space indentation and we will look into fixing up the whole codebase in another PR.

src/main/java/org/lightcouch/internal/URIBuilder.java
```diff
@@ -72,7 +88,7 @@ public URI buildEncoded() {
 		}
 		try {
 			String q = (query.length() == 0) ? "" : "?" + query;
-			String uri = String.format("%s://%s:%s%s%s%s", new Object[]{scheme, host, port, path, pathToEncode, q});
+			String uri = String.format("%s://%s:%s%s%s%s", new Object[]{scheme, host, port, path, encodedPath, q});
```
tomblench: I know this wasn't in your change but can we use the varargs syntax here instead (remove `new Object[]{}`)?

src/main/java/org/lightcouch/internal/URIBuilder.java
```diff
@@ -54,17 +54,33 @@ public static URIBuilder buildUri(URI uri) {
...
 	public URI build() {
 		try {
+			StringBuilder queryBuilder = new StringBuilder(encodeQuery(query.toString()));
+
 			for (int i = 0; i < qParams.size(); i++) {
 				String amp = (i != qParams.size() - 1) ? "&" : "";
-				query.append(qParams.get(i) + amp);
+				String param;
+				String queryParam = qParams.get(i);
+				int firstEqualsIndex = queryParam.indexOf('=');
```
tomblench: I'm not sure if this is going to work properly. It assumes they don't want the first `=` encoding but they want all of the rest encoded. Why don't we investigate removing the `public URIBuilder query(String query)` flavour and then it will force API users to give their keys and values separately. If this isn't possible we should keep this flavour only for internal use and/or with a health warning to potential users.

src/main/java/org/lightcouch/internal/URIBuilder.java
brynh: Addressed in 956aeda.

src/main/java/org/lightcouch/internal/URIBuilder.java
brynh: Addressed in aa78892 & 136cc50:
* aa78892 deprecates the methods that don't keep keys and values separate from each other;
* 136cc50 stores the query parameters internally as key/value pairs. `Params` now stores a list of `Param` rather than `String`, where a `Param` is a key/value pair.  The `Param` class has a method `toURLEncodedString` which handles the URL encoding of the query parameters.

mikerhodes:
We shouldn't have changed the whitespacing for the affected files in this PR, but looks like we can't go back on that now. It's difficult to review.

brynh:
With regard to the whitespace changes, that was done due to tomblench's comment on this PR. Unfortunately going back on that now would be difficult. However, it's often possible to ignore whitespace changes depending how you're reviewing, it is a hidden feature in github's web interface: adding `?w=0` to the end of the URL of most github diff views ignores whitespace changes.

src/test/java/com/cloudant/tests/UpdateHandlerTest.java
```diff
@@ -18,63 +18,65 @@
...
+        db.syncDesignDocsWithDb();
+    }
+
+    @AfterClass
+    public static void tearDownClass() {
+        dbClient.shutdown();
+    }
+
+    @Test
+    public void updateHandler_queryString() {
```
tomblench: Can we check that the special characters, + = etc are being correctly encoded here by using them in newValue?

src/test/java/com/cloudant/tests/SearchTests.java
```diff
@@ -1,208 +1,244 @@
...
+        uriBuilder.host(cloudantAccount + ".cloudant.com");
+        uriBuilder.port(443);
+        uriBuilder.path("/animaldb/_design/views101/_search/animals");
+        uriBuilder.query("include_docs", true);
+        uriBuilder.query("q", query);
+        URI uri = uriBuilder.build();
+
+        String uriBaseString = account.getBaseUri().toASCIIString();
+
+		String expectedUriString = uriBaseString + "animaldb/_design/views101/_search/animals?include_docs=true&q=" + expectedResult;
```
tomblench: Indenting

src/test/java/com/cloudant/tests/SearchTests.java
```diff
@@ -1,208 +1,244 @@
...
+        assertEquals("omnivore", it.next()); // diet in reverse order
+        assertEquals("herbivore",it.next());
+    }
+
+
+
+    @Test
+    public void drillDownTest() {
+        // do a faceted search for drilldown
+                Search srch = db.search("views101/animals");
```
tomblench: Indenting

tomblench:
+1 after my latest round of minor comments

src/test/java/com/cloudant/tests/UpdateHandlerTest.java
brynh: Addressed in 5b19d7f

src/test/java/com/cloudant/tests/SearchTests.java
brynh: Addressed in fbac6e8

src/test/java/com/cloudant/tests/SearchTests.java
brynh: Addressed in fbac6e8

mikerhodes:
:+1: Perhaps work with @tomblench to squash?

PR #54, opened 2015-06-01
46809 evaluate cert trust chain

brynh:
Fixes FB#46809.

Currently, connections made via HTTPS disable hostname verification and certificate chain validation. This PR changes the behaviour so that hostname verification and certificate chain validation are enabled by default. A `disableSSLAuthentication` option has been added to allow hostname verification and certificate chain validation to be explicitly disabled/enabled.

*TESTING*
Tests have been added to check that connection to the Cloudant service both with and without hostname verification and certificate chain validation work correctly. Further tests use a simple local HTTPS server running with an invalid certificate to verify that the connection fails when hostname verification and certificate chain validation is enabled and succeeds when disabled.

reviewer @mikerhodes
reviewer @rhyshort


rhyshort:
:+1: LGTM

src/main/java/com/cloudant/client/api/model/ConnectOptions.java
```diff
@@ -60,8 +71,12 @@ public String getProxyHost() {
...
-	
-	
-	
-	
+
+	/** @return true if hostname verification and certificate chain validation are
+	 *  set to disabled or false otherwise.
+	 *  @see #disableSSLAuthentication(boolean) */
+	public boolean isSSLAuthenticationDisabled() {
+		return disableSSLAuthentication;
```
mikerhodes: Rename the ivar to match the accessor?

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
```diff
@@ -0,0 +1,286 @@
...
+    private static String DUMMY_USERNAME = "username";
+    private static String DUMMY_PASSWORD = "password";
+
+    private static boolean localServerReady;
+    private static final Object lock = new Object();
+
+    /**
+     * A very simple HTTPS Server that runs on the localhost and has a simple certificate that won't
+     * work for hostname verification.
+     * <p>
```
mikerhodes: Either self-close the tag `<p/>` or wrap the paragraphs correctly. Seems `<p/>` is the javadoc way.

mikerhodes:
Do we need doc updates here?

One thing is that we're changing a default in a breaking way, we at least need a big notice in the changelog.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
brynh: I believe `<p>` is correct for Javadoc. It's one of those weird things that's inconsistent with normal HTML/XML formatting - see the 7th bullet under *Format of a Doc Comment* http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format

src/main/java/com/cloudant/client/api/model/ConnectOptions.java
brynh: Addressed in a93f44a

brynh:
Made doc updates to document the breaking change in 127ee2a.

brhaible:
A comment about the API: The class ConnectOptions, so far, is a Java Bean, that is, all its properties follow the getter/setter pattern (getFoo/isFoo for the getter, setFoo for the setter) of Java Beans https://docs.oracle.com/javase/tutorial/javabeans/writing/properties.html. To continue to use this pattern, I would suggest to rename the method 'disableSSLAuthentication' to 'setSSLAuthenticationDisabled'. (It may sound like a monster word, but such setter method names are commonplace in Java, because of the Java Beans pattern.)

mikerhodes:
+1. Appropriate squashing needed before merge.

brhaible:
Bryn,
Belonging to the method rename ('disableSSLAuthentication' -> 'setSSLAuthenticationDisabled') that you did after my comment, the documentation is no longer up-to-date. Please pull a doc update from https://github.com/brhaible/java-cloudant/commits/46809-adjust-documentation

brynh:
Well spotted - thanks Bruno.

brhaible:
Bryn,
The wording in the CHANGES.md
    
      To disable these additional checks when the client connects to
      the database,  `setSSLAuthenticationDisabled(true)` can be called
      on the `ConnectOptions` object passed to the `CloudantClient`
      constructor.
    
sounds like you can invoke the CloudantClient constructor with a
ConnectOptions object and then afterwards call setSSLAuthenticationDisabled
on it. But this will not have the desired effect, because the
ConnectOptions object is never accessed again by the CloudantClient
class, after the constructor terminated. I would therefore suggest
to change the wording:
    
      To disable these additional checks when the client connects to
      the database,  `setSSLAuthenticationDisabled(true)` can be called
      on the `ConnectOptions` object before you pass it to the `CloudantClient`
      constructor.

Please pull a doc update from https://github.com/brhaible/java-cloudant/commits/46809-adjust-documentation

PR #56, opened 2015-06-08
47056 query previous page with descending view

brynh:
Fixes FB#47056

When using a descending view, `queryPreviousPage` didn't go to the previous page. Also, when calling `queryNextPage` after `queryPreviousPage`, the paging was incorrect.  This PR fixes these paging issues.

A new flag `pagingForwards` has been added to track changes in paging direction, so we know when the paging direction has changed.

`queryNextPage` and `queryPreviousPage` have been refactored to reduce code duplication as there were only minor differences between the methods.

*TESTING*
Tests have been added to verify the paging is correct in ascending and descending views, with multiple changes of direction of paging.

reviewer @mikerhodes
reviewer @rhyshort

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,12 +6,13 @@
   `setSSLAuthenticationDisabled(true)` can be called on the `ConnectOptions`
   object before you pass it to the `CloudantClient` constructor.
 - [NEW] New API for deleting databases, `CloudantClient.deleteDB(String name)`
-- [Deprecated] Deprecated `CloudantClient.deleteDB(String name, String confirm)`
-  API.
+- [FIX] Fixed querying of next/previous page in a descending view.
 - [FIX] Fixed handling of non-ASCII characters when the platform's
   default charset is not UTF-8.
 - [FIX] Fixed encoding of `+`, `=` and `&` characters when they are used
   in the query part of a URL.
+- [Deprecated] Deprecated `CloudantClient.deleteDB(String name, String confirm)`
+  API.
 
 # 1.0.1 (2015-02-04)
 
--- a/src/main/java/org/lightcouch/Params.java
+++ b/src/main/java/org/lightcouch/Params.java
@@ -92,4 +92,25 @@ public class Params {
     public void addAll(Params params) {
         this.params.addAll(params.params);
     }
+
+    /**
+     * If this object already contains a parameter whose name matches {@code name}, replace the
+     * existing value with {@code value} for the first instance of parameter {@code name}.
+     * Otherwise, add a parameter with the new {@code name} and {@code value}.
+     * @param name The name of the parameter to replace/add.
+     * @param value The new value of the parameter.
+     */
+    public void replaceOrAdd(String name, String value) {
+        boolean found = false;
+        for (Param param : params) {
+            if (param.getKey().equals(name)) {
+                param.setValue(value);
+                found = true;
+                break;
+            }
+        }
+        if (!found) {
+            addParam(name, value);
+        }
+    }
 }
--- a/src/main/java/org/lightcouch/View.java
+++ b/src/main/java/org/lightcouch/View.java
@@ -109,7 +109,18 @@ public class View {
 	private Boolean includeDocs;
 	private Boolean inclusiveEnd;
 	private Boolean updateSeq;
-	
+
+	private enum PagingDirection {
+		FORWARD,
+		BACKWARD
+	}
+
+	/** This will be FORWARD initially or if
+	 * {@link #queryPage(PagingDirection, int, String, String, String, String, Class)} has been
+	 * called, it will match the last {@code direction} passed to
+	 * {@link #queryPage(PagingDirection, int, String, String, String, String, Class)}.*/
+	private PagingDirection pagingDirection = PagingDirection.FORWARD;
+
 	private CouchDatabaseBase dbc;
 	private Gson gson;
 	private URIBuilder uriBuilder;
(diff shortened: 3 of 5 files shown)


src/main/java/org/lightcouch/internal/URIBuilder.java
```diff
@@ -115,10 +115,37 @@ public URIBuilder pathToEncode(String path) {
...
+     * @param value The value to add/update.
+     * @param update set to true to update an existing query parameter matching {@code name},
+     *               or false to add a new one. Note that if this is true and there is no
+     *               parameter matching {@code name}, the parameter will be added.
+     * @return The updated {@link URIBuilder} object.
+     */
+    public URIBuilder query(String name, Object value, boolean update) {
+        if (name != null && value != null) {
+            if (update) {
+                this.qParams.updateOrAdd(name, value.toString());
```
rhyshort: Couldn't we just use this rather than have the update flag?

brynh: I don't know enough about queries with Cloudant to know if it ever would make sense to have multiple query parameters with the same name?  If it does, then you wouldn't want to update, but would want to add in those cases. I kind of assumed that there could be cases when it's necessary to have multiple query parameters with the same name and different values partly because `Params` stores the parameters internally using a list (whereas I think it would be more sensible to use a map if the keys are unique).  If we're sure that this is the way to go, we should also change `Params` to use a map.

rhyshort: I'm pretty certain Cloudant/Couch doesn't use the same name for query parameters, thoughts @mikerhodes

mikerhodes: I'm not really sure what `update` means in this context, it's ambiguous -- does `replace` or `append` make more sense as the variable name (along with appropriate logic changes)? Off the top of my head I'm not sure whether any view query params take lists, but I suggest we stick with adding items to the list and maintain the param to the method.

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- [FIX] Fixed querying of next/previous page in a descending view.
```
mikerhodes: Ordering of changes: NEW, CHANGE, FIX, DEPRECATED. Could you move the FIX down?

src/main/java/org/lightcouch/Params.java
```diff
@@ -92,4 +92,25 @@ public Param get(int index) {
...
+            if (param.getKey().equals(name)) {
+                param.setValue(value);
+                found = true;
+                break;
+            }
+        }
+        if (!found) {
+            addParam(name, value);
+        }
+    }
```
mikerhodes: Seems like "update" is "replace" to me, more explicit.

brynh: Fixed in 5ff00c0.

CHANGES.md
brynh: Fixed in f3efdf7.

src/main/java/org/lightcouch/View.java
```diff
@@ -309,78 +313,42 @@ public boolean queryForBoolean() {
...
+
 	/**
 	 * @return The previous page.
 	 */
 	private <T> Page<T> queryPreviousPage(int rowsPerPage, String currentStartKey, 
 			String currentStartKeyDocId, String startKey, String startKeyDocId, Class<T> classOfT) {
+		return queryPage(true, rowsPerPage, currentStartKey, currentStartKeyDocId, startKey, startKeyDocId, classOfT);
+	}
+
+	private <T> Page<T> queryPage(boolean previousPage, int rowsPerPage, String currentStartKey,
```
mikerhodes: Maybe create an enum for direction to replace `previousPage`?

src/main/java/org/lightcouch/View.java
```diff
@@ -109,7 +109,11 @@
 	private Boolean includeDocs;
 	private Boolean inclusiveEnd;
 	private Boolean updateSeq;
-	
+
+	/** This will be true initially or if the most recent call to {@link #queryNextPage} or
+	 * {@link #queryPreviousPage} was {@link #queryNextPage}, and false otherwise. */
+	private boolean pagingForwards = true;
```
mikerhodes: Here too, pagingDirection and either enum or defined forward/backward constants.

src/main/java/org/lightcouch/internal/URIBuilder.java
```diff
@@ -115,10 +115,37 @@ public URIBuilder pathToEncode(String path) {
...
+     * Add the given {@code name} and {@code value} to the query parameters or update
+     * the existing query parameter matching {@code name} with the new {@code value}.
+     * @param name The name to add/update.
+     * @param value The value to add/update.
+     * @param update set to true to update an existing query parameter matching {@code name},
+     *               or false to add a new one. Note that if this is true and there is no
+     *               parameter matching {@code name}, the parameter will be added.
+     * @return The updated {@link URIBuilder} object.
+     */
+    public URIBuilder query(String name, Object value, boolean update) {
```
mikerhodes: This `update` var name needs changing to `replace` too.

mikerhodes:
It looks like your tests paginate all the way forward or backwards. It'd also be good to see a test that goes partway forward, then partway backwards, then forward again to the end (or similar). Essentially not assuming reading all pages available in a given direction.

src/main/java/org/lightcouch/View.java
brynh: Addressed in c6eb443.

src/main/java/org/lightcouch/View.java
brynh: Addressed in c6eb443.

src/main/java/org/lightcouch/internal/URIBuilder.java
brynh: Addressed in 8a3481a.

brynh:
Added additional tests that don't page all the way through the available pages before changing paging direction.

mikerhodes:
Looks good now, +1 for squashing.

rhyshort:
+1

PR #57, opened 2015-06-18
Add get/setSecureSSLSocketFactory configuration option.

brhaible:
A custom SSLSocketFactory can be set to enhance security in specific environments.
More details in FB#48311.
reviewer @mikerhodes 
reviewer @ricellis

### DIFF
--- a/src/main/java/org/lightcouch/CouchDbClient.java
+++ b/src/main/java/org/lightcouch/CouchDbClient.java
@@ -25,6 +25,7 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
 
 import org.apache.http.Consts;
 import org.apache.http.HttpHost;
@@ -205,6 +206,8 @@ public class CouchDbClient extends  CouchDbClientBase {
 
 		if("https".equals(props.getProtocol())) {
 			if (props.isSSLAuthenticationDisabled()) {
+				// No SSL authentication.
+				// No need for a custom SSLSocketFactory in this case.
 				SSLContext sslcontext = SSLContexts.custom()
 						.loadTrustMaterial(null, new TrustStrategy() {
 							public boolean isTrusted(X509Certificate[] chain, String authType)
@@ -216,8 +219,18 @@ public class CouchDbClient extends  CouchDbClientBase {
 				return registry.register("https", new SSLConnectionSocketFactory(sslcontext,
 						SSLConnectionSocketFactory.ALLOW_ALL_HOSTNAME_VERIFIER)).build();
 			} else {
-				return registry.register("https", new SSLConnectionSocketFactory(SSLContexts.createDefault(),
-						SSLConnectionSocketFactory.STRICT_HOSTNAME_VERIFIER)).build();
+				// With SSL authentication enabled.
+				// A custom SSLSocketFactory can be set to
+				// enhance security in specific environments.
+				SSLSocketFactory factory = props.getSecureSSLSocketFactory();
+				if (factory != null) {
+					return registry.register("https", new SSLConnectionSocketFactory(factory,
+							SSLConnectionSocketFactory.STRICT_HOSTNAME_VERIFIER)).build();
+				} else {
+					// Use the default SSL configuration and truststore of the JRE.
+					return registry.register("https", new SSLConnectionSocketFactory(SSLContexts.createDefault(),
+							SSLConnectionSocketFactory.STRICT_HOSTNAME_VERIFIER)).build();
+				}
 			}
 		} else {
 			return registry.register("http", PlainConnectionSocketFactory.INSTANCE).build();
--- a/src/main/java/org/lightcouch/CouchDbClient.java
+++ b/src/main/java/org/lightcouch/CouchDbClient.java
@@ -222,7 +222,7 @@ public class CouchDbClient extends  CouchDbClientBase {
 				// With SSL authentication enabled.
 				// A custom SSLSocketFactory can be set to
 				// enhance security in specific environments.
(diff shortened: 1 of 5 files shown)


src/main/java/org/lightcouch/CouchDbProperties.java
```diff
@@ -192,4 +195,27 @@ public boolean isSSLAuthenticationDisabled() {
...
+	 * Specifies the SSLSocketFactory to use when connecting to CouchDB
+	 * over a <code>https</code> URL.
+	 * @param factory An SSLSocketFactory, or <code>null</code> for the
+	 *                default SSLSocketFactory of the JRE.
+	 * @see #getSecureSSLSocketFactory()
+	 */
+	public CouchDbProperties setSecureSSLSocketFactory(SSLSocketFactory factory) {
+		this.secureSSLSocketFactory = factory;
+		return this;
+	}
```
mikerhodes: Why are these called "setSecureSSLSocketFactory" and not just setSSLSocketFactory? I'm not sure the socket factory need be more secure...

src/main/java/org/lightcouch/CouchDbClient.java
```diff
@@ -205,6 +206,8 @@ private PoolingHttpClientConnectionManager createConnectionManager(
 
 		if("https".equals(props.getProtocol())) {
 			if (props.isSSLAuthenticationDisabled()) {
+				// No SSL authentication.
+				// No need for a custom SSLSocketFactory in this case.
```
mikerhodes: If it doesn't make sense to both disable SSL authentication and set a custom SSL factory this needs at least documenting in the public documentation, if not including checks for in the setters in `ConnectOptions`.

src/main/java/org/lightcouch/CouchDbProperties.java
brhaible: > Why are these called "setSecureSSLSocketFactory" and not just setSSLSocketFactory?
The name 'setSSLSocketFactory' would imply that this SSLSocketFactory is used always. But we don't want it to be used in the "insecure" mode - because the use-case of the custom SSLSocketFactory is to increase security (e.g. by using only the TLSv1.2 protocol, or by using a truststore that contains very few certificate authorities), and the point of the insecure mode is to avoid all kinds of SSL configuration hassles in situations where security does not matter.

> I'm not sure the socket factory that is set need be more secure...
It does not need to be. The method name should not imply "set a more secure socket factory" but is meant to imply "set a socket factory for the secure mode".

src/main/java/org/lightcouch/CouchDbClient.java
brhaible: <i>&gt; this needs at least documenting in the public documentation</i>
Yes, thanks for noticing this. I've added a commit "Javadoc tweak" to this branch.

<i>&gt; if not including checks for in the setters in ConnectOptions.</i>
Such checks would not be in line with Java Bean coding conventions. Java Bean specification implies that the various properties of a Java Bean are *independent* - even if afterwards their effects/semantics are interdependent  -, which implies that an application can set the different properties in arbitrary order. This guarantee that the order in which you call the set... methods is arbitrary is not only important for GUI applications that shows a property dialog, but also for the normal programmer who does not expect that the order matters. So,
<pre>options.setSSLAuthenticationDisabled(true);
options.setSecureSSLSocketFactory(myFactory);
options.setSSLAuthenticationDisabled(false);
</pre>
should have the same effect as
<pre>options.setSSLAuthenticationDisabled(true);
options.setSSLAuthenticationDisabled(false);
options.setSecureSSLSocketFactory(myFactory);
</pre>
and the last of these code snippets definitely should not produce an exception.

mikerhodes: Thanks for the notes w.r.t. bean-like expectations, I wasn't aware of those.

Thanks for adding the docs.

src/main/java/org/lightcouch/CouchDbProperties.java
mikerhodes: Okay, let's go with `setSecureModeSSLSocketFactory`. A mouthful, but avoids the incorrect implication.

mikerhodes:
I wondered for a while if we could avoid the mutually exclusive options somehow by combining the custom SSL factory with the secure mode flag. However, as far as I can see, it's rather hard to avoid the two options here as the `SSLSocketFactory` needs to be combined with the `ALLOW_ALL_HOSTNAME_VERIFIER` option in the registry builder to get the "insecure" mode.

Perhaps we can indicate the two options are inter-related via naming, `setAuthenticatedModeSSLSocketFactory`?

ricellis:
I'm +1 after the method renaming.

brhaible:
> Perhaps we can indicate the two options are inter-related via naming, setAuthenticatedModeSSLSocketFactory?
Agree. This naming provides a better indication that the methods are related, than 'setSecureSSLSocketFactory'.

Done: https://github.com/brhaible/java-cloudant/commits/48311-Support-custom-SSLSocketFactory

Now, what are the next steps to do before this can be merged? Should be combine the 3 commits into a single one (in my fork under a different branch name, or in the official repository with the same branch name)? Should I deal with the conflict in CHANGES.md, or can you do it?

ricellis:
Squash your 3 commits into 1 in a branch in the official repository. If you resolve the conflict then it should be able to merge automatically.

brhaible:
> Squash your 3 commits into 1 in a branch in the official repository.

Done. The branch is here: https://github.com/cloudant/java-cloudant/tree/48311-Support-custom-SSLSocketFactory-merged

I resolved the conflicts in CHANGES.md and in the source code files (by applying the new codestyle to my changes).

ricellis:
Squash your 3 commits into 1 in a branch in the official repository. If you resolve the conflict then it should be able to merge automatically.

ricellis:
Moved to #62

PR #58, opened 2015-06-26
Added User-Agent to requests

ricellis:
*What*
Add a User-Agent header to requests of the form:
java-cloudant/1.0.2 [Java (amd64; Windows 7; 6.1) Oracle Corporation; 1.8.0_45; 1.8.0_45-b14]

*How*
* Generate a client.properties file as part of the build containing the agent name and version
* At runtime read the client.properties and combine with JVM system properties
* Add the string to the client request headers

*Testing*
Added a test that validates the code is generating strings of the expected format

reviewer @emlaver 
reviewer @mikerhodes

### DIFF
--- a/src/main/java/org/lightcouch/CouchDbClient.java
+++ b/src/main/java/org/lightcouch/CouchDbClient.java
@@ -50,12 +50,15 @@ import org.apache.http.protocol.BasicHttpContext;
 import org.apache.http.protocol.HttpContext;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
 import java.net.URLDecoder;
 import java.security.KeyManagementException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
+import java.util.Properties;
 
 import javax.net.ssl.SSLContext;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/lightcouch/CouchDbClient.java
```diff
@@ -95,6 +98,41 @@
...
+            ua = properties.getProperty("user.agent.name", ua);
+            version = properties.getProperty("user.agent.version", version);
+        } catch (Exception ex) {
+            //swallow exception and keep using default values
+        } finally {
+            if (propStream != null) {
+                try {
+                    propStream.close();
+                } catch (IOException e) {
+                    //can't do anything else
```
emlaver: Is it worth trying to add an error message in the response if an exception occurs?

src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -101,4 +103,19 @@ public void cookieNegativeTest() {
         }
     }
 
+    /**
+     * Assert that the User-Agent header is of the expected form.
+     * It would be nice to test that the header is actually added to requests, but we can't
+     * assert that without making assumptions about the underlying client or asserting server-side.
```
mikerhodes: Note @brynh's use of a mock HTTP server in the certificate tests if you want to do the full thing here.

mikerhodes:
+1

src/main/java/org/lightcouch/CouchDbClient.java
ricellis: If there is an error trying to parse the client.properties we will use a default User-Agent value of "java-cloudant/unknown ..." on the request. I don't think this value should have any impact on the HTTP response at all because a user could change the User-Agent string anyway.
Is that the response you meant?
Or do you feel we should log an error client side?

My view is that the user probably doesn't care if the User-Agent string is set correctly or not they just want the client to work so we should just do our best to set it correctly and proceed.

src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: Thanks @mikerhodes - added a new test that checks the value on an actual request in 878c96c

src/main/java/org/lightcouch/CouchDbClient.java
emlaver: Ok makes sense, thanks.

emlaver:
+1

PR #59, opened 2015-06-26
Intermittent failures of Replicator and Replication tests

emlaver:
*What*
Investigate the intermittent failures of replicator and replication test classes.

*How*
The tests were disabled as part of case 47450.  They required a timing fix which was resolved with Thread.sleep().  Also, the replicator documents used in each test are now removed when the test is complete.

*Testing*
Since these classes are test cases, no additional tests are required.

FogBugz: 47545

reviewer @ricellis 
reviewer @mikerhodes

### DIFF
--- a/src/test/java/com/cloudant/tests/ReplicatorTest.java
+++ b/src/test/java/com/cloudant/tests/ReplicatorTest.java
@@ -9,20 +9,17 @@ import com.cloudant.client.api.Database;
 import com.cloudant.client.api.model.ReplicatorDocument;
 import com.cloudant.client.api.model.Response;
 import com.cloudant.client.api.model.ViewResult;
+import com.cloudant.tests.util.Utils;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.UUID;
 
-
-@Ignore
 public class ReplicatorTest {
 
     private static Properties props;
@@ -58,27 +55,34 @@ public class ReplicatorTest {
     }
 
     @Test
-    public void replication() {
-        account.replicator()
+    public void replication() throws Exception {
+        Response response = account.replicator()
                 .createTarget(true)
                 .source(db1URI)
                 .target(db2URI)
                 .save();
 
+        // find and remove replicator doc
+        Utils.waitForReplicatorToComplete(account, response.getId());
+        Utils.removeReplicatorTestDoc(account, response.getId());
     }
 
     @Test
-    public void replication_filteredWithQueryParams() {
+    public void replication_filteredWithQueryParams() throws Exception {
         Map<String, Object> queryParams = new HashMap<String, Object>();
         queryParams.put("somekey1", "value 1");
 
-        account.replicator()
+        Response response = account.replicator()
                 .createTarget(true)
                 .source(db1URI)
                 .target(db2URI)
                 .filter("example/example_filter")
                 .queryParams(queryParams)
                 .save();
+
+        // find and remove replicator doc
+        Utils.waitForReplicatorToComplete(account, response.getId());
+        Utils.removeReplicatorTestDoc(account, response.getId());
     }
 
     @Test
@@ -95,28 +99,21 @@ public class ReplicatorTest {
                 .createTarget(true)
                 .save();
 
-        Thread.sleep(1000);
+        // we need the replication to finish before continuing
+        Utils.waitForReplicatorToStart(account, response.getId());
 
         // find all replicator docs
         List<ReplicatorDocument> replicatorDocs = account.replicator()
                 .findAll();
         assertThat(replicatorDocs.size(), is(not(0)));
 
-        // find replicator doc
-        ReplicatorDocument replicatorDoc = account.replicator()
-                .replicatorDocId(response.getId())
-                .find();
-
-        // cancel a replication
-        account.replicator()
-                .replicatorDocId(replicatorDoc.getId())
-                .replicatorDocRev(replicatorDoc.getRevision())
-                .remove();
+        // find and remove replicator doc
+        Utils.removeReplicatorTestDoc(account, response.getId());
     }
 
     @Test
     public void replication_conflict() throws Exception {
-        String docId = generateUUID();
+        String docId = Utils.generateUUID();
         Foo foodb1 = new Foo(docId, "title");
         Foo foodb2 = null;
 
@@ -124,11 +121,12 @@ public class ReplicatorTest {
 
         db1.save(foodb1);
 
-        account.replicator().source(db1URI)
+        Response response = account.replicator().source(db1URI)
                 .target(db2URI).replicatorDocId(docId)
                 .save();
 
-        Thread.sleep(1000); //we need the replication to finish before continuing
+        // we need the replication to finish before continuing
+        Utils.waitForReplicatorToComplete(account, response.getId());
 
         foodb2 = db2.find(Foo.class, docId);
         foodb2.setTitle("titleY");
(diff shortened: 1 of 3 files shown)


mikerhodes:
To me, `Thread.sleep()` is almost always wrong. For these tests, I'm assuming the intermittent failures were against Cloudant/Couch 2.0. Therefore it looks like they may have been failing because the `find` calls were coming too soon after the `put` calls and for certain busy clusters, perhaps things were not replicating between the nodes quickly enough (i.e., eventual consistency issues).

If this is the case, instead of waiting an arbitrary amount of time _every time_, you should be retrying the call(s) below the doc inserts a few times. Then you can set the timeout lower (say ramping up the gaps between retries: 0.05, 0.1, 0.2, 0.4, 0.8, 1, 2 seconds) which will stop these tests taking a second unnecessarily while still covering all but the most egregious of problems.

ricellis:
I agree with @mikerhodes hard coded `Thread.sleep()` values always make me uneasy. We should loop until the replication has completed or we reach some test timeout.

Also what do you think about putting `assertThat(removeResponse.getError(), is(nullValue()));` in the `removeReplicatorTestDoc` method and making it void instead of returning the remove response? The check appears to be the same in all cases so moving it into the method reduces the amount of code duplication.

emlaver:
After some difficulty getting the travis build to pass, I've added @mikerhodes and @ricellis comments in to commits 7481953 and c532b6d.  I did some squashing to remove code that I used to print test details during the travis build.  
There is a new method that checks whether the while loop should continue or if it should timeout.  The default timeout constant used by Thread.Sleep is 100 ms. In the method hasTimedOut, I've set maximum for the timeout constant to be 3.2 seconds.  After 3.2 seconds, the loop will exit and the replication document variable will be null.  Let me know if this constant should be increased.

src/test/java/com/cloudant/tests/ReplicatorTest.java
```diff
@@ -138,15 +135,20 @@ public void replication_conflict() throws Exception {
         foodb1.setTitle("titleZ");
         db1.update(foodb1);
 
-        account.replicator().source(db1URI)
+        Response secondResponse = account.replicator().source(db1URI)
                 .target(db2URI).save();
 
-        Thread.sleep(1000);
+        // we need the replication to finish before continuing
+        Utils.waitForReplicatorToComplete(account, response);
```
ricellis: Doesn't this one need to be `waitForReplicatorToComplete(account, secondResponse)`?

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -33,4 +44,61 @@ public static String getHostName(String account) {
...
+
+        //Reset timeout to default 50 milliseconds
+        TIMEOUT_CONSTANT = 100;
+
+        return replicatorDoc;
+    }
+
+    private static boolean hasTimedOut() {
+        long currentTime = System.currentTimeMillis();
+        if(TIMEOUT_CONSTANT <= 3200 && currentTime < (currentTime + TIMEOUT_CONSTANT)) {
```
ricellis: The check `currentTime < (currentTime + TIMEOUT_CONSTANT))` is always going to be true if `TIMEOUT_CONSTANT` is >0. It is also a bit confusing for me that the `TIMEOUT_CONSTANT` is not a constant. How would you feel about removing this `hasTimedOut` method and changing the while loop in `waitForReplicatorToComplete` to something like:
```java
        long startTime = System.currentTimeMillis();
        long timeout = startTime + TIMEOUT_CONSTANT;
        //initial wait of 100 ms
        long delay = 100;
        while (!finished && System.currentTimeMillis() < timeout) {
            //Sleep before finding replication document
            Thread.sleep(delay);

            replicatorDoc = account.replicator()
                    .replicatorDocId(replicatorResponse.getId())
                    .find();

            //If replicator state is completed or triggered, return true
            //If false, continue in the while loop until replicator is complete

            if (replicatorDoc != null && replicatorDoc.getReplicationState() != null
                    && (replicatorDoc.getReplicationState().equalsIgnoreCase("completed")
                    || (replicatorDoc.getReplicationState().equalsIgnoreCase("triggered")
                    && replicatorDoc.getContinuous()))) {
                finished = true;
            }
            //double the delay for the next iteration
            delay *= 2;
        }
```

emlaver:
Added @ricellis comments and other changes in d2b91c1 and 71a585f.

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -1,13 +1,25 @@
...
 
 import java.io.InputStream;
 import java.util.Properties;
+import java.util.UUID;
 
 public class Utils {
+
+    //Timeout in milliseconds
+    private static long TIMEOUT_CONSTANT = 3200;
+    private static long timeout;
```
ricellis: This static timeout is not needed anymore.

ricellis:
+1

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -33,4 +44,58 @@ public static String getHostName(String account) {
...
+
+        Response removeResponse = account.replicator()
+                .replicatorDocId(replicatorDoc.getId())
+                .replicatorDocRev(replicatorDoc.getRevision())
+                .remove();
+
+        assertThat(removeResponse.getError(), is(nullValue()));
+    }
+
+    public static ReplicatorDocument waitForReplicatorToComplete(CloudantClient account, Response
```
mikerhodes: `removeReplicatorTestDoc` and `waitForReplicatorToComplete` should just take the document ID I think, as that's the only attribute that's used -- prefer passing the smallest possible amount of information to a method so it's clear what its dependencies are.

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -1,13 +1,24 @@
...
 import org.apache.commons.logging.Log;
 
 import java.io.InputStream;
 import java.util.Properties;
+import java.util.UUID;
 
 public class Utils {
+
+    //Timeout in milliseconds
+    private static long TIMEOUT_CONSTANT = 3200;
```
mikerhodes: Just call the variable `TIMEOUT_MILLISECONDS` then you can avoid the need to comment the units above. Also, remove `CONSTANT` given that's implied. Also the variable should be final.

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -33,4 +44,58 @@ public static String getHostName(String account) {
...
+            replicatorDoc = account.replicator()
+                    .replicatorDocId(replicatorResponse.getId())
+                    .find();
+
+            //If replicator state is completed or triggered, return true
+            //If false, continue in the while loop until replicator is complete
+            if (replicatorDoc != null && replicatorDoc.getReplicationState() != null
+                    && (replicatorDoc.getReplicationState().equalsIgnoreCase("completed")
+                    || (replicatorDoc.getContinuous() != null && replicatorDoc.getContinuous()
+                    && replicatorDoc.getReplicationState().equalsIgnoreCase("triggered")))) {
```
mikerhodes: The method name isn't accurate given it only waits for continuous replications to start.

I'd split this up into `waitForReplicationToStart` and `waitForReplicationToComplete` so you don't have this nasty-to-understand if statement and incorrect method naming.

src/test/java/com/cloudant/tests/util/Utils.java
emlaver: The account variable is required to access the replicator.  The only way to not have the account passed to this method is to move the method outside of the Utils package and into this class.

src/test/java/com/cloudant/tests/util/Utils.java
emlaver: Fixed in 49ed003

src/test/java/com/cloudant/tests/util/Utils.java
emlaver: Fixed in 49ed003

src/test/java/com/cloudant/tests/util/Utils.java
emlaver: Fixed in 49ed003

mikerhodes: I actually meant to swap `replicatorResponse` for just the doc ID, I think leaving the account as being passed in is fine. For some reason I missed the `account` param on first read.

emlaver: Fixed in 8826b94

mikerhodes:
+1. Squash before merging.

Issue #39, opened 2015-02-18
startKey/endKey with complex keys


srl295:
I think there's a bug with the startKey and endKey functions as well as key()

Even in the `ViewsTest` there is this snippet in `byComplexKey()`
```
	public void byComplexKey() {
		int[] complexKey = new int[] { 2011, 10, 15 };
		List<Foo> foos = db.view("example/by_date")
				.key(complexKey)
//				.key(2011, 10, 15) 
```

I think the problem is that one `(Object...)` function is calling another `(Object...)` function.  I can't remember how to fix this yet, but want to capture.

I had to do similar code to the above workaround in my own project.

mariobriggs:
a test case with expected results and received results would be good. For instance we had usage issues like #29

srl295:
@mariobriggs thanks. I'm trying to make a test case.

ricellis:
The `ViewsTest.byComplexKey()` works with `view.key(2011,10,15)` since release 1.1.2.

PR #63, opened 2015-07-16
View's queryPage doesn't seem to handle multi-value values

emlaver:
*What:*
View’s queryPage doesn’t seem to handle multi-value values (FB 48355)
The queryPage method doesn’t handle views that emit keys such as [“uuid”, 100].

#39 DB view key varargs needs parameter typing (FB 48532)
The key, start key, and end key methods are not able to handle complex keys.  One example is a complex key with multiple integer values i.e. key(2011 10, 15).

*How:*
The queryView method in View’s queryPage had a hard cast of String.  This would cause an exception if an integer value existed in a key from the result rows. To fix this:
The String parameter was changed to Object class to allow for other data types. 
If the key in the result rows is an array of multiple values (i.e. startkey=[“uuid”,100]), the array will be converted into one string and then be added as a property for the page parameters.

View’s query had a similar problem where multiple number value keys were not property handled when building the URI.  During testing, if a string and integer value existed in a key array, then the URI would be properly created. If the key only contained an array of numbers, then it would throw an exception. 
The fix required a statement in getKeyAsJson to check whether the multi value key array was exactly one value and an instance of the Number class.

*Testing:*
An existing test case under ViewsTest (byComplexKey) was modified to use the new multi value change.
New test cases were added to ViewsTest:
Assert that result list is not empty using multi value keys in both View’s query and queryPage method.
Run tests with multi value key arrays from one to three values

reviewer @ricellis 
reviewer @mikerhodes

### DIFF
--- a/src/main/java/org/lightcouch/View.java
+++ b/src/main/java/org/lightcouch/View.java
@@ -94,10 +94,10 @@ public class View {
     private static final String REDUCE_JS = "reduce.js";
 
     // view fields
-    private String key;
-    private String startKey;
+    private JsonElement key;
+    private JsonElement startKey;
     private String startKeyDocId;
-    private String endKey;
+    private JsonElement endKey;
     private String endKeyDocId;
     private Integer limit;
     private String stale;
@@ -117,9 +117,9 @@ public class View {
 
     /**
      * This will be FORWARD initially or if
-     * {@link #queryPage(PagingDirection, int, String, String, String, String, Class)} has been
+     * {@link #queryPage(PagingDirection, int, JsonElement, String, JsonElement, String, Class)} has been
      * called, it will match the last {@code direction} passed to
-     * {@link #queryPage(PagingDirection, int, String, String, String, String, Class)}.
+     * {@link #queryPage(PagingDirection, int, JsonElement, String, JsonElement, String, Class)}.
      */
     private PagingDirection pagingDirection = PagingDirection.FORWARD;
 
@@ -307,9 +307,9 @@ public class View {
             return queryPage(PagingDirection.FORWARD, rowsPerPage, null, null, null, null,
                     classOfT);
         }
-        String currentStartKey;
+        JsonElement currentStartKey;
         String currentStartKeyDocId;
-        String startKey;
+        JsonElement startKey;
         String startKeyDocId;
         String action;
         try {
@@ -320,9 +320,9 @@ public class View {
                 log.debug("Paging Param Decoded = " + json);
             }
             final JsonObject jsonCurrent = json.getAsJsonObject(CURRENT_KEYS);
-            currentStartKey = jsonCurrent.get(CURRENT_START_KEY).getAsString();
+            currentStartKey = jsonCurrent.get(CURRENT_START_KEY);
             currentStartKeyDocId = jsonCurrent.get(CURRENT_START_KEY_DOC_ID).getAsString();
-            startKey = json.get(START_KEY).getAsString();
+            startKey = json.get(START_KEY);
             startKeyDocId = json.get(START_KEY_DOC_ID).getAsString();
             action = json.get(ACTION).getAsString();
         } catch (Exception e) {
@@ -337,9 +337,9 @@ public class View {
         }
     }
 
-    private <T> Page<T> queryPage(PagingDirection direction, int rowsPerPage, String
+    private <T> Page<T> queryPage(PagingDirection direction, int rowsPerPage, JsonElement
             currentStartKey,
-                                  String currentStartKeyDocId, String startKey, String
+                                  String currentStartKeyDocId, JsonElement startKey, String
                                           startKeyDocId, Class<T> classOfT) {
         // set view query params
         limit(rowsPerPage + 1);
@@ -361,8 +361,9 @@ public class View {
         // init page, query view
         final Page<T> page = new Page<T>();
         final List<T> pageList = new ArrayList<T>();
-        final ViewResult<String, String, T> vr = queryView(String.class, String.class, classOfT);
-        final List<ViewResult<String, String, T>.Rows> rows = vr.getRows();
+
+        final ViewResult<JsonElement, String, T> vr = queryView(JsonElement.class, String.class, classOfT);
+        final List<ViewResult<JsonElement, String, T>.Rows> rows = vr.getRows();
         final int resultRows = rows.size();
         final int offset = vr.getOffset();
         final long totalRows = vr.getTotalRows();
@@ -373,7 +374,8 @@ public class View {
         final JsonObject currentKeys = new JsonObject();
         final JsonObject jsonNext = new JsonObject();
         final JsonObject jsonPrev = new JsonObject();
-        currentKeys.addProperty(CURRENT_START_KEY, rows.get(0).getKey());
+
+        currentKeys.add(CURRENT_START_KEY, rows.get(0).getKey());
         currentKeys.addProperty(CURRENT_START_KEY_DOC_ID, rows.get(0).getId());
         for (int i = 0; i < resultRows; i++) {
             // set keys for the next page
@@ -381,7 +383,7 @@ public class View {
                 boolean isLastPage = resultRows <= rowsPerPage;
                 if (!isLastPage) {
                     page.setHasNext(true);
-                    jsonNext.addProperty(START_KEY, rows.get(i).getKey());
+                    jsonNext.add(START_KEY, rows.get(i).getKey());
                     jsonNext.addProperty(START_KEY_DOC_ID, rows.get(i).getId());
                     jsonNext.add(CURRENT_KEYS, currentKeys);
                     jsonNext.addProperty(ACTION, NEXT);
@@ -397,7 +399,7 @@ public class View {
                 || (direction == PagingDirection.BACKWARD && offset == totalRows - rowsPerPage - 1);
         if (!isFirstPage) {
             page.setHasPrevious(true);
-            jsonPrev.addProperty(START_KEY, currentStartKey);
+            jsonPrev.add(START_KEY, currentStartKey);
             jsonPrev.addProperty(START_KEY_DOC_ID, currentStartKeyDocId);
             jsonPrev.add(CURRENT_KEYS, currentKeys);
             jsonPrev.addProperty(ACTION, PREVIOUS);
@@ -428,8 +430,8 @@ public class View {
      * @param key The key value, accepts a single value or multiple values for complex keys.
      */
     public View key(Object... key) {
-        this.key = getKeyAsJson(key);
-        uriBuilder.query("key", this.key);
+        this.key = getKeyAsJsonElement(key);
+        uriBuilder.query("key", gson.toJson(this.key));
         return this;
     }
 
@@ -438,8 +440,8 @@ public class View {
      *                 keys.
      */
     public View startKey(Object... startKey) {
-        this.startKey = getKeyAsJson(startKey);
-        uriBuilder.query("startkey", this.startKey);
+        this.startKey = getKeyAsJsonElement(startKey);
+        uriBuilder.query("startkey", gson.toJson(this.startKey));
         return this;
     }
 
@@ -453,8 +455,8 @@ public class View {
      * @param endKey The end key value, accepts a single value or multiple values for complex keys.
      */
     public View endKey(Object... endKey) {
-        this.endKey = getKeyAsJson(endKey);
-        uriBuilder.query("endkey", this.endKey);
+        this.endKey = getKeyAsJsonElement(endKey);
+        uriBuilder.query("endkey", gson.toJson(this.endKey));
         return this;
     }
 
@@ -552,7 +554,9 @@ public class View {
      * @return
      */
     public View keys(Object... keys) {
-        this.allDocsKeys = String.format("{%s:%s}", gson.toJson("keys"), getKeyAsJson(keys));
+        JsonObject keysObject = new JsonObject();
+        keysObject.add("keys", getKeyAsJsonElement(keys));
+        this.allDocsKeys = gson.toJson(keysObject);
         return this;
     }
 
@@ -582,8 +586,13 @@ public class View {
         return this;
     }
 
-    private String getKeyAsJson(Object... key) {
-        return (key.length == 1) ? gson.toJson(key[0]) : gson.toJson(key); // single or complex key
+    private JsonElement getKeyAsJsonElement(Object... key) {
+        // single or complex key
+        if(key.length == 1) {
+            return gson.toJsonTree(key[0]);
+        } else {
+            return gson.toJsonTree(key).getAsJsonArray();
+        }
     }
 
     /**
--- a/src/test/java/com/cloudant/tests/ViewsTest.java
+++ b/src/test/java/com/cloudant/tests/ViewsTest.java
@@ -30,13 +30,13 @@ import com.cloudant.client.api.View;
 import com.cloudant.client.api.model.Page;
 import com.cloudant.client.api.model.ViewResult;
 import com.cloudant.test.main.RequiresDB;
+import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
-import org.lightcouch.DocumentConflictException;
 import org.lightcouch.NoDocumentException;
 
 import java.util.Arrays;
@@ -49,7 +49,8 @@ public class ViewsTest {
 
     private static Database db;
     private CloudantClient account;
-
+    private static String[] testViews = {"example/doc_title", "example/creator_created",
+            "example/creator_boolean_total"};
 
     @Before
     public void setUp() {
@@ -65,7 +66,6 @@ public class ViewsTest {
         account.deleteDB("lightcouch-db-test");
         account.shutdown();
     }
-
 
     @Test
     public void queryView() {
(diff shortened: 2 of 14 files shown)


src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -594,6 +731,91 @@ private static void init() {
...
+            keyViews.put("created", mapFunctions);
+
+            designDoc.setViews(keyViews);
+
+            db.save(designDoc);
+
+            db.syncDesignDocsWithDb();
+
+        } catch (DocumentConflictException e) {
+        }
```
mikerhodes: Conflicts should probably fail the test?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -105,12 +108,146 @@ public void byStartAndEndKey() {
...
+
+
+        query = db.view("keyarray_example/creator_created_boolean")
+                .startKey(new Object[]{true, "uuid", 1})
+                .endKey(new Object[]{true, "uuid", 2000})
+                .includeDocs(true);
+
+        result = query.query(Object.class);
+
+        assertThat(result.size(), is(2));
```
mikerhodes: I assume a tear-down is cleaning up the database for you?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -105,12 +108,146 @@ public void byStartAndEndKey() {
...
+            page = query.queryPage(30, null, Object.class);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
+        assertThat(page.getResultList().size(), is(3));
+    }
+
+    @Test
+    public void byStartAndEndThreeKeyArray() {
```
mikerhodes: Add java doc for test purpose.

src/main/java/org/lightcouch/View.java
```diff
@@ -373,7 +373,19 @@ public boolean queryForBoolean() {
...
+        Object getKey = rows.get(0).getKey();
+        //Handle multi value key array
+        if(getKey instanceof ArrayList) {
+            String keyAsString = "[";
+            for(Object key : (ArrayList<Object>)getKey) {
+                keyAsString += key.toString() + ",";
+            }
+            //Final key value as string with closing bracket
+            getKey = keyAsString.substring(0,keyAsString.length()-1).concat("]");
+        }
```
mikerhodes: Seems like you are trying to make a JSON array here. As far as I can see, this will make invalid JSON for anything other than numbers and boolean values, unless the keys are already pre-encoded as JSON (i.e., escaped and surrounded in quotes).

Why not use GSON here rather than manual concatenation?

mikerhodes:
> The queryPage method doesn’t handle views that emit values such as [“uuid”, 100].

Do you mean "emit keys such as" here rather than "values"?

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Yes, the tearDown method is deleting the whole test database after all tests are completed.

emlaver:
Yes, I meant "keys".  Fixed.

src/main/java/org/lightcouch/View.java
ricellis: Yeah it would be cleaner if it can be a JsonArray rather than trying to manipulate the string

ricellis:
I think it might be worth having an additional test that uses a non-String, non-array key type to make sure that path works with these changes. Maybe something like `byStartAndEndIntKey()`

emlaver:
@mikerhodes @ricellis I've added your comments in the latest commit.  I removed the manual concatenation code.  Now in getKeyAsJson, booleans and numbers are be passed as an object array to toJson.  With that, there's no extra code needed (just use toString) to convert the Object array to string for the page parameters.

mikerhodes:
I suggest adding tests where pagination is required: it looks like the changes [here](https://github.com/cloudant/java-cloudant/pull/63/files#diff-462176273836fe6f0e80f32ba1f1aeadR386) affect pagination as they're in an `if` check for `!isLastPage`.

emlaver:
@mikerhodes The message will be separated during squashing. I'll also add a test case for pagination.

Just FYI, I compared the debug test results in the console from master branch with this branch.  The values for paging parameters were the same.

src/main/java/org/lightcouch/View.java
```diff
@@ -583,7 +585,14 @@ public View tempView(MapReduce mapReduce) {
...
     private String getKeyAsJson(Object... key) {
-        return (key.length == 1) ? gson.toJson(key[0]) : gson.toJson(key); // single or complex key
+        // single or complex key
+        if (key.length == 1 && (key[0] instanceof Boolean || key[0] instanceof Number)) {
+            return gson.toJson(key);
+        } else if (key.length == 1) {
+            return gson.toJson(key[0]);
+        } else {
+            return gson.toJson(key);
+        }
```
mikerhodes: Why does `[1]` and `[true]` get special treatment here?

src/main/java/org/lightcouch/View.java
```diff
@@ -381,7 +383,7 @@ public boolean queryForBoolean() {
                 boolean isLastPage = resultRows <= rowsPerPage;
                 if (!isLastPage) {
                     page.setHasNext(true);
-                    jsonNext.addProperty(START_KEY, rows.get(i).getKey());
+                    jsonNext.addProperty(START_KEY, rows.get(i).getKey().toString());
```
mikerhodes: Why is `toString` required in this case and on line 377?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -104,13 +105,238 @@ public void byStartAndEndKey() {
...
+     */
+
+    @Test
+    public void byStartAndEndBooleanKey() {
+        multiValueKeyInit();
+
+        View query = db.view("keyarray_example/boolean")
+                .startKey(false)
+                .endKey(false)
+                .includeDocs(true);
```
mikerhodes: Do you need `includeDocs`, you don't look at the docs.

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -573,25 +799,138 @@ private Page checkPagesBackward(int currentPage, int numberOfPages, Page page, b
...
+
+        //Created single value map function for boolean key test
+        mapFunctions = new DesignDocument.MapReduce();
+
+        mapFunctions.setMap(
+                "function(doc) { emit([doc.boolean], null); }");
+
+        keyViews.put("boolean", mapFunctions);
+
+        designDoc.setViews(keyViews);
```
mikerhodes: You'd have much less code here if you just used a big JSON string for this ddoc -- does the library let you do that?

src/main/java/org/lightcouch/View.java
emlaver: @mikerhodes I realized after this addition above that the following start/end keys are valid and can effect the pagination: [1], ["1"], ["abc"].  I added a note in FB that I was in progress of fixing this.  The idea would be to check if the key is an array containing one value, and then manipulating the value before setting the final page parameters for start/end key.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: You're right, the includeDocs are not needed.  They were just copied from another text case.

src/main/java/org/lightcouch/View.java
emlaver: In order to fix the original problem with the queryView causing an exception, I changed the first parameter of the queryView constructor from String.class to Object.class (line 365).  Now, the `getKey()` method above will return an object.  I needed to call `toString` to get the string representation and set that to the start key.  

In my latest tests, I'm finding that key arrays with one string value are missing the quotes by the time it reaches this code.  For example, a key with "["abc"]" or "["1"]" would show as "[abc]" or "[1]".  Here is a gist of an if block that would add the required quotes:
https://gist.github.com/emlaver/81cc2db5cbfc0291ff8e

I'd like to replace the code in the gist with something more efficient.  I may need to manipulate the key in the `JsonToObject` method of CouchDBUtil.java.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: I was going to create map.js classes under the resources/example/views folder.  This is where the views for the rest of the test cases (like example/foo) exist.

emlaver:
I've made several changes:
- Move the views that were originally hard-coded in the ViewsTest class to the design-docs/example directory
- Use JsonElement object instead of String for start/end key in View class. Some benefits of this change:
   - No need to manipulate the JSON within a String object where the syntax is not recognized
   - Easy to work with a JsonArray object for single or multi-value key arrays
- Add additional test cases (and more to come) for multi-value keys in query, queryPage, and with pagination.
- Mike's comment to remove `includeDocs`

ricellis:
@emlaver is this PR still valid? I feel like it has been superseded by #71

emlaver:
@ricellis Commit 2129358 contains fixes you commented on for the javadoc in ViewsTest class.

emlaver:
@ricellis commit 15663de completely replaces ```getKeyAsJson``` with ```getKeyAsJsonElement```.  This also required the **key** class variable to be changed from a String to a JsonElement.

src/main/java/org/lightcouch/View.java
```diff
@@ -428,7 +430,7 @@ public boolean queryForBoolean() {
      * @param key The key value, accepts a single value or multiple values for complex keys.
      */
     public View key(Object... key) {
-        this.key = getKeyAsJson(key);
+        this.key = getKeyAsJsonElement(key);
         uriBuilder.query("key", this.key);
```
mikerhodes: After the change to using JsonElement rather than String, to fix the bug with complex keys, you are implicitly relying on JsonElement's `toString` method returning a valid JSON string here for us as part of the URL, which isn't part of the `toString` method's contract. 

Use `Gson.toJson` to guarantee you are sending in valid JSON in the URL. 

The same goes for `startKey` and `endKey`, below.

src/main/java/org/lightcouch/View.java
```diff
@@ -552,7 +554,7 @@ public View updateSeq(Boolean updateSeq) {
      * @return
      */
     public View keys(Object... keys) {
-        this.allDocsKeys = String.format("{%s:%s}", gson.toJson("keys"), getKeyAsJson(keys));
+        this.allDocsKeys = String.format("{%s:%s}", gson.toJson("keys"), getKeyAsJsonElement(keys));
```
mikerhodes: While you are here, fix this line to do something sensible:

```
JsonElement e = new JsonElement();
e.add("keys", getKeyAsJsonElement(keys));
gson.toJson(e);
```

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+
+        page = null;
+        try {
+            page = query.queryPage(30, null, Object.class);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
+        assertThat(page.getResultList().size(), is(3));
+    }
```
mikerhodes: This is two tests, split it into two methods.

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+
+        query = db.view("example/created")
+                .startKey(new Object[]{1})
+                .endKey(new Object[]{2000});
+
+
+        result = query.query(Object.class);
+
+        assertThat(result.size(), is(3));
+    }
```
mikerhodes: Again, two tests, split up. Particularly because the method name isn't correct for the first test.

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+
+        List<Object> result = query.query(Object.class);
+
+        assertThat(result.size(), is(3));
+
+        Page page = null;
+        try {
+            page = query.queryPage(30, null, Object.class);
+        } catch (Exception e) {
+            e.printStackTrace();
```
mikerhodes: Why isn't this a test failure?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+
+        page = null;
+        try {
+            page = query.queryPage(30, null, Object.class);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
+        assertThat(page.getResultList().size(), is(3));
+    }
```
mikerhodes: Seems like there are three tests here?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+    public void byStartAndEndThreeKeyArray() {
+        init();
+
+        View query = db.view("example/boolean_creator_created")
+                .startKey(false, "uuid", 1)
+                .endKey(false, "uuid", 2000);
+
+        List<Object> result = query.query(Object.class);
+
+        assertThat(result.size(), is(3));
```
mikerhodes: Why are we doing this in every test?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -528,6 +923,63 @@ private void checkPagination(boolean descending, int docCount, int docsPerPage,
...
+     *
+     * @param descending  true if the view is in descending order, and false otherwise.
+     * @param docCount    the total number of documents in the view.
+     * @param docsPerPage the maximum number of documents per page in the view.
+     * @param useNewView  true for creating a new view for the query, and false otherwise.
+     * @param pageToPages the list of page numbers to page to.
+     */
+    private void checkPaginationWithMultiValueKey(boolean descending, boolean useNewView,
+                                                  int docCount, int docsPerPage,
+                                                  int... pageToPages) {
```
mikerhodes: This shouldn't be a vararg method, it makes things very hard to follow at call sites.

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -552,6 +1004,35 @@ private Page checkPagesForward(int currentPage, int numberOfPages, Page page, bo
...
+     * @param view          the {@link View} object to use to perform the query (or null to create
+     *                         a new
+     *                      view for the query).
+     * @param viewCount     the current view in the array of multi-value key views. This value is
+     *                      used when view parameter is null.
+     * @return the last page in the view.
+     */
+    private Page checkPagesForwardMultiValueKey(int currentPage, int numberOfPages, Page page, boolean
+            descending, int docCount, int docsPerPage, View view, int viewCount) {
+        for (int i = 0; i < numberOfPages; ++i) {
```
mikerhodes: Why `++i` rather than the more common `i++`?

src/test/resources/design-docs/example/views/boolean/map.js
```diff
@@ -0,0 +1,3 @@
+function(doc) {
+  emit(doc.contentArray[0].boolean, null);
+}
```
mikerhodes: Be sure all newly added files have trailing new lines. (Set up your editor to always add them).

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -267,6 +481,15 @@ public void paginationAscendingReusingViewAllTheWayInEachDirectionTwice() {
     }
 
     /**
+     * Run the same check as paginationAscendingReusingViewAllTheWayInEachDirectionTwice
+     * test on views using an array of multi-value keys.
+     */
+    @Test
+    public void paginationAscendingReusingMVKeyViewAllTheWayInEachDirectionTwice() {
+        checkPaginationWithMultiValueKey(false, 6, 2, 3, 1, 3, 1);
```
mikerhodes: You could use a simple builder approach using inner classes to make these tests much more readable.

```java
public class CheckPaginationWithMultiValueKey {
 public b() {
   return new CheckPaginationWithMultiValueKey();
  }
  public descending(bool descending) {
    this.descending = descending;
    return this;
  }
  /* etc for other args */
  public runTest() {
    /* do current content of checkPaginationWithMultiValueKey method here */
  }
}

mikerhodes:
Looks like this does fix the bug, but introduces a couple of others that need fixing.

In addition, I'd like to see a number of style and readability fixes.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in e75a2f1

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in e75a2f1

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: This test case will pass but is not an array of integers for multi-value key testing.  Fixed in e75a2f1

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in e75a2f1

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: I've changed the start/end key to produce different results in all the multi-value key tests in e75a2f1.  What other test results should we check for?

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: @mikerhodes What if I changed constructor to ```checkPaginationWithMultiValueKey(int docCount, int docsPerPage, boolean descending, boolean useNewView, int... pageToPages)``` to separate the single int parameters from the int vararg?

emlaver: @mikerhodes Never-mind the comment above.  I will go with your builder pattern approach.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: @mikerhodes this was the design in the original code.  I believe this was used because pages need to start at 1.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in 7078797

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in 7078797

src/main/java/org/lightcouch/View.java
emlaver: Fixed in 0db11c3

src/main/java/org/lightcouch/View.java
emlaver: Fixed in 0db11c3

ricellis:
+1

src/test/resources/design-docs/example/views/boolean/map.js
emlaver: Fixed in 4de5a8d

mikerhodes:
Looks great. +1. Looks like it can be squashed to just one commit before merging?

After spending the time reviewing the PR, I think the commit's title should be something like "Fix paginated view queries that use complex keys", and remember the `BugzID: ` bit.

PR #65, opened 2015-07-21
Fix javadoc build errors for Java 8

ricellis:
*What*
Resolve doclint javadoc validation errors so java-cloudant can build on Java 8 with default settings

*How*
Removed invalid javadoc to pass Java 8 doclint validation
Fixed codestyle to not insert invalid <p/> elements during formatting.

*Testing*
Clean gradle builds with javadoc on Java 8

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -30,7 +30,6 @@ import org.lightcouch.CouchDbProperties;
 import org.lightcouch.Replication;
 import org.lightcouch.Replicator;
 import org.lightcouch.Response;
-import org.lightcouch.View;
 import org.lightcouch.internal.CouchDbUtil;
 
 import java.net.URI;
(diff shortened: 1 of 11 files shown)


rhyshort:
+1

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -48,9 +48,7 @@
  * <pre>
  * CloudantClient client = new CloudantClient("mycloudantaccount","myusername","mypassword");
  * </pre>
- * <p/>
```
tomblench: I think you need to fix up the paragraph tags throughout this javadoc block as they don't balance even after your changes. If it's a set of steps the user needs to go through then maybe a list (`<ul>`) is better.

tomblench: In general a bit more formatting here would make the output a bit nicer - I know you don't have time to fix all of the doc formatting but this is an important class.

ricellis: @tomblench - I've tried to make this better in 6f1350f, wdyt?

tomblench:
Much nicer formatting +1

PR #66, opened 2015-07-23
48534 permissions api

### LINKED ISSUE #23: Change setPermissions to use _api/v2/db/<dbname>/_security insteas of /dbname/_security
Per https://github.com/cloudant/nodejs-cloudant/issues/6#issuecomment-61628907 this will handle the merging into /dbname/_security

ricellis:
*What*

Update set/get permissions methods to use the new API.
Fixes #23

*How*

1. Replace calls to /$db/_security with the new endpoint /api/v2/db/$db/_security.
1. Update the Permissions type with the roles available for the new endpoint.
1. Document that these methods are only applicable to databases that support the new API and throw an UnsupportedOperationException if the API is not available.

*Testing*
* The existing permissions test still passes.
* Added a new test for the UnsupportedOperationException

reviewer @tomblench 
reviewer @rhyshort


src/test/java/com/cloudant/tests/DatabaseTest.java
```diff
@@ -71,6 +72,22 @@ public void permissions() {
...
+    /**
+     * Test that when called against a DB that is not a Cloudant service
+     * an UnsupportedOperationException is thrown
+     */
+    @Test
+    public void testPermissionsException() {
+        try {
+            Map<String, EnumSet<Permissions>> userPerms = db.getPermissions();
+            //ignore the test in cases where we don't get an exception
+            Assume.assumeTrue(false);
```
tomblench: I always find the logic for `Assume` confusing - not that it's your fault. Can we have a message here like "skipping test as it looks like the DB is a Cloudant service"

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -81,13 +82,25 @@
         super();
         this.client = client;
         this.db = db;
+        apiV2DBSecurityURI = buildUri(client.getBaseUri()).path("_api/v2/db/").path(db.getDbName()).path("/_security").build();
```
rhyshort: line length

.gitignore
```diff
@@ -11,3 +11,4 @@ src/test/resources/cloudant-2.properties
 *.iml
 build/
 .gradle/
+gradle.properties
```
rhyshort: I don't think we should be ignoring the global gradle properties, local.properties we should ignore though

ricellis: I thought it would reduce the risk of people committing credentials after running tests.

src/test/java/com/cloudant/tests/DatabaseTest.java
```diff
@@ -71,6 +72,22 @@ public void permissions() {
...
+        }catch(UnsupportedOperationException e){
```
rhyshort: You can do this with `@Test(expectedclass=UnsupportedOperationException)`

ricellis: Not with an Assume in place, because Assume throws an `AssumptionViolatedException` which then causes the test to fail because it wasn't the expected exception. I will be able to used `expected=` however, if we get rid of the Assume with more categories as suggested by @tomblench.

rhyshort: You wouldn't need to Assume using `expected` because you'd remove that code and if the exception wasn't thrown it would fail. which is what the Assume.assumeTrue is effectively doing

ricellis: Then the test would fail against the Cloudant service because there is no UnsupportedOperationException thrown, but the test wouldn't be skipped because there would be no assume code. It doesn't matter anyway because I can use expected if we skip the test through the extra categories.

src/test/java/com/cloudant/tests/DatabaseTest.java
ricellis: New test categories as discussed including updated running test information in: 8ea51b8 and 0679f26

src/main/java/com/cloudant/client/api/Database.java
ricellis: Fixed in 271a9de

.gitignore
ricellis: I fixed the ignore to be for local.properties in 5803642, but local.properties doesn't get picked up by gradle so isn't the right place to set test configuration. I have updated the instructions in 0679f26 to suggest using GRADLE_USER_HOME for the gradle.properties if you don't want to accidentally commit them.

src/test/java/com/cloudant/tests/DatabaseTest.java
ricellis: Fixed in 8ea51b8

rhyshort:
+1

src/test/java/com/cloudant/tests/DatabaseTest.java
```diff
@@ -71,6 +74,16 @@ public void permissions() {
 
     }
 
+    /**
+     * Test that when called against a DB that is not a Cloudant service
+     * an UnsupportedOperationException is thrown
+     */
+    @Test(expected = UnsupportedOperationException.class)
+    @Category({RequiresCouch.class, RequiresCloudantLocal.class})
```
tomblench: Is it possible to make the whole class `@Category(RequiresDB.class)` and then only add in the necessary others? My guess is that the categories compose in this way but I'm not sure.

ricellis: Done in f6eea47, but sadly this only had the effect of moving 1 test method annotation to the class.

tomblench:
+1

PR #67, opened 2015-07-24
49843 cloudant tests

ricellis:
*What*
Fixed Replicator and Index tests that were failing against the Cloudant service.

*How*
Remove extraneous prefix from IndexTest
Update documentation to make it clear that the _replicator DB must exist.

*Testing*
Fixes apply to existing tests.

reviewer @rhyshort


rhyshort:
+1

PR #69, opened 2015-07-28
Defaulted max connections to 6

ricellis:
*What*
Increased the default number of max connections to 6 (from 2).

*How*
Added a default value to CouchDbProperties. This makes more sense than adding it to ConnectOptions because that object is only used in non-default cases.

*Testing*
Code exercised by existing tests, no specific testing of the number of connections.

reviewer @rhyshort


rhyshort:
:+1: LGTM

PR #71, opened 2015-08-05
48355 querypage multi value

rhyshort:


src/main/java/org/lightcouch/View.java
```diff
@@ -438,14 +440,20 @@ public View key(Object... key) {
...
      */
     public View startKey(Object... startKey) {
-        this.startKey = getKeyAsJson(startKey);
-        uriBuilder.query("startkey", this.startKey);
+        this.startKey = getKeyAsJsonElement(startKey);
+        uriBuilder.query("startkey", this.startKey, true);
+        return this;
+    }
+
+    public View startKeyDocId(String startKeyDocId, boolean replace) {
```
ricellis: If we're hard-coding `replace=true` in the default method then adding this new method allows someone to specify `false` but is that necessary? Are there any circumstances where we would expect to have two of the same parameter for a view?

build.gradle
```diff
@@ -18,6 +18,7 @@ targetCompatibility = 1.6
 repositories {
     maven { url "http://oss.sonatype.org/content/repositories/snapshots" }
     maven { url "http://repo.maven.apache.org/maven2" }
+    mavenLocal();
```
ricellis: This is highlighting to me that our repository list is unusual, I think it should be:
```gradle
repositories {
    mavenLocal()
    mavenCentral()
}
```
If we're going to use the local repo we should check it in preference to central. I also don't think we need the snapshot repo url here as we should not normally build from snapshots and the uploads for snapshots are handled by a different repository element in the `mavenDeployer`.

src/main/java/org/lightcouch/View.java
```diff
@@ -337,12 +337,12 @@ public boolean queryForBoolean() {
...
 
-    private <T> Page<T> queryPage(PagingDirection direction, int rowsPerPage, String
+    private <T> Page<T> queryPage(PagingDirection direction, int rowsPerPage, JsonElement
             currentStartKey,
-                                  String currentStartKeyDocId, String startKey, String
+                                  String currentStartKeyDocId, JsonElement startKey, String
                                           startKeyDocId, Class<T> classOfT) {
         // set view query params
-        limit(rowsPerPage + 1);
+        limit(rowsPerPage + 1, true);
```
ricellis: I think this should just be `limit(rowsPerPage +1);` unless someone can convince me that there is a views case where we need to specify multiple limit parameters.

src/main/java/org/lightcouch/View.java
```diff
@@ -583,7 +603,22 @@ public View tempView(MapReduce mapReduce) {
     }
 
     private String getKeyAsJson(Object... key) {
-        return (key.length == 1) ? gson.toJson(key[0]) : gson.toJson(key); // single or complex key
```
ricellis: Can we get rid of this method and use `getKeyAsJsonElement()` (and appropriate `toJson` calls) everywhere?
Otherwise we appear to have two very similar methods, but only one that handles the array cases correctly.

src/main/java/org/lightcouch/View.java
```diff
@@ -583,7 +603,22 @@ public View tempView(MapReduce mapReduce) {
...
+        } else {
+            return gson.toJson(key);
+        }
+    }
+
+    private JsonElement getKeyAsJsonElement(Object... key) {
+        // single or complex key
+        // If the startKey is a single key value, grab the first value from the JSON tree array
+        if(key.length == 1) {
+            return gson.toJsonTree(key).getAsJsonArray().get(0);
```
ricellis: If there is only one object, is there an easier way to get the `JsonElement` besides moving through `JsonTree` and `JsonArray` before just getting the element?
i.e. doesn't  `gson.toJsonTree(key[0]);` work?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+
+        assertThat(result.size(), is(3));
+    }
+
+    /**
+     * Assert that passing a key with two values in query
+     * and queryView will produce a result list of all docs.
+     *
+     * Run the same assert above but with a key that is an
+     * object array containing only integers.
```
ricellis: Think this should be changed to only **numbers** since the code says `.startKey(1, 12.00)` which is a mix of int and float (but all numbers from a JSON perspective).

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -107,13 +107,227 @@ public void byStartAndEndKey() {
...
+        query = db.view("example/creator_created")
+                .startKey(new Object[]{"uuid", 1})
+                .endKey(new Object[]{"uuid", 2000});
+
+
+        result = query.query(Object.class);
+
+        assertThat(result.size(), is(3));
+
+        //Key array with only integers
```
ricellis: again only numbers

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -258,6 +472,8 @@ private static void checkPage(Page page, int docCount, int docsPerPage, int page
      * we expect. This test uses the same {@link View} instance for each page query.
      * Page forward to the last page, back to the first page, forward to the last page and back to
      * the first page.
+     *
+     * Run the same check above on an array of multi-value key views.
```
ricellis: Does this javadoc belong with the next test method?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -266,12 +482,19 @@ public void paginationAscendingReusingViewAllTheWayInEachDirectionTwice() {
         checkPagination(false, 6, 2, view, 3, 1, 3, 1);
     }
 
+    @Test
```
ricellis: Add javadoc? Something like "Run the same check as the test paginationAscendingReusingViewAllTheWayInEachDirectionTwice on views using an array of multi-type keys."

build.gradle
emlaver: The ```mavenLocal()``` method was added after a discussion for this custom release.  This does not need to be in this PR but is the a reason why it does not currently exist?

src/main/java/org/lightcouch/View.java
emlaver: @mikerhodes @ricellis What cases would we need ```false``` for in ```uriBuilder.query(key, value, false)```?  All the existing test cases and additional complex key test cases are passing with the true value.

src/main/java/org/lightcouch/View.java
emlaver: I agree that we don't need another method with a second boolean parameter.

src/main/java/org/lightcouch/View.java
emlaver: Yes, the two methods are similar.  I believe it shouldn't be difficult to make the required changes in the ```key(Object... key)``` and ```keys(Object... keys)``` to call ```getKeyAsJsonElement()```.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: I think removing this javadoc and adding a javadoc to each pagination multi-value key test is best.

emlaver:
Closing this PR.  All commits will be in https://github.com/cloudant/java-cloudant/pull/63.

build.gradle
ricellis: No reason that I am aware of, if it doesn't get included here I'll just update the gradle later.

src/main/java/org/lightcouch/View.java
ricellis: For the query URIs built by the `View` class I can see no cases where appending rather than replacing a parameter would be appropriate so we should always just use `true` to replace in this class.

PR #72, opened 2015-08-10
50101 remove db get

ricellis:
*What*
Remove the `GET` calls to a database as part of the instantiation of `Database` objects

*Why*
Some users repeatedly instantiate `Database` objects. The repeated `GET` calls reduce performance, but offer little benefit. Whilst doing the `GET` does validate that the backing DB exists, the DB could easily be deleted between the object instantiation and any subsequent calls anyway. As such it makes more sense to remove the call to `GET` and allow the `NoDocumentException` to be thrown when calls are made against a `Database` object for which the backing DB has been removed or does not exist, rather than at object instantiation time.

*How*
* Remove the dbUri `GET` calls from the `connect()` method
* Refactor `CouchDatabaseBase` to rename `connect()` to `create()` and call it only when `create=true`
* Wrap 412 responses from create `PUT` calls in a subclass of `CouchDbException` so that `create=true` behaviour remains consistent with the current behaviour, but without the need for the initial `GET` call to check DB existence.

*Testing*
* Added new tests for exception cases
    1. request against non-existent DB
    2. trying to create a DB that already exists
* Calls to `client.database(...)` are repeatedly exercised throughout existing tests

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+- [IMPROVED] Removed unconditional GET request when creating `Database` objects
+  This offered little protection to the developer while performing a hidden HTTP request.
+
 # 1.1.0 (2015-07-31)
 
 - [BREAKING CHANGE] Hostname verification and certificate chain
--- a/src/main/java/com/cloudant/client/api/Database.java
+++ b/src/main/java/com/cloudant/client/api/Database.java
@@ -64,6 +64,8 @@ import java.util.Set;
 /**
  * Contains a Database Public API implementation.
  *
+ * Methods may throw a {@link NoDocumentException} if the database does not exist.
+ *
  * @author Mario Briggs
  * @since 0.0.1
  */
--- a/src/main/java/org/lightcouch/CouchDatabaseBase.java
+++ b/src/main/java/org/lightcouch/CouchDatabaseBase.java
@@ -62,7 +62,9 @@ public abstract class CouchDatabaseBase {
         this.dbName = name;
         this.client = client;
         this.dbURI = buildUri(client.getBaseUri()).path(name).path("/").build();
-        connect(create);
+        if (create) {
+            create();
+        }
         this.design = new CouchDbDesign(this);
     }
 
@@ -478,22 +480,17 @@ public abstract class CouchDatabaseBase {
     }
 
 
-    private void connect(boolean create) {
-
-        InputStream getresp = null;
+    private void create() {
         HttpResponse putresp = null;
-
         try {
-            getresp = get(dbURI);
-        } catch (NoDocumentException e) { // db doesn't exist
-            if (!create) {
-                throw e;
-            }
             final HttpPut put = new HttpPut(dbURI);
             putresp = client.executeRequest(put);
             log.info(String.format("Created Database: '%s'", dbName));
+        } catch (PreconditionFailedException e) {
+            // The PreconditionFailedException is thrown if the database already existed.
+            // To preserve the behaviour of the previous version, suppress this type of exception.
+            // All other CouchDbExceptions will be thrown.
         } finally {
-            close(getresp);
             close(putresp);
         }
     }
--- a/src/main/java/org/lightcouch/PreconditionFailedException.java
+++ b/src/main/java/org/lightcouch/PreconditionFailedException.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2015 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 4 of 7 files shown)


CHANGES.md
```diff
@@ -1,3 +1,5 @@
+- [IMPROVED] Removed HTTP GET request when instantiating `Database` objects
```
rhyshort: I think we should word this more along the lines of 

`[IMPROVED] Improved performance by reducing the number of requests made when instantiating Database objects.`

rhyshort:
other than a tiny comment about the CHANGES.md +1

CHANGES.md
ricellis: I've blended the two statements in a4fc8da

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -64,6 +64,8 @@
 /**
  * Contains a Database Public API implementation.
  *
+ * Methods may throw a {@link NoDocumentException} if the database does not exist.
```
tomblench: Just checking, but is this not in a `@throws` clause because it's an unchecked exception?

src/main/java/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -478,22 +480,16 @@ public String getDbName() {
...
-            getresp = get(dbURI);
-        } catch (NoDocumentException e) { // db doesn't exist
-            if (!create) {
-                throw e;
-            }
             final HttpPut put = new HttpPut(dbURI);
             putresp = client.executeRequest(put);
             log.info(String.format("Created Database: '%s'", dbName));
+        } catch (PreconditionFailedException e) {
+            //suppress this exception thrown if the DB already existed
```
tomblench: Can you clarify this comment a bit? I don't really understand it

src/main/java/org/lightcouch/PreconditionFailedException.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package org.lightcouch;
+
+/**
+ * CouchDbException class for 412 precondition failed status codes
```
tomblench: is there a useful link to the couch docs we can give for this? I know that 412 is used for illegal DB names, anything else?

src/main/java/com/cloudant/client/api/Database.java
ricellis: Sadly, yes. The `CouchDbException` hierarchy are all descended from `RuntimeException`. I think we'd like to address that at some point in the future, but it would be a big breaking change that I don't think we're ready for yet.

src/main/java/org/lightcouch/CouchDatabaseBase.java
ricellis: Is this more clear?
```
// The PreconditionFailedException is thrown if the database already existed.
// To preserve the behaviour of the previous version, suppress this type of exception.
// All other CouchDbExceptions will be thrown.
```

tomblench: Yup, that's clearer

ricellis: Added in commit 91e9579

src/main/java/org/lightcouch/PreconditionFailedException.java
ricellis: As far as I can tell from a search of the CouchDB API reference, the only documented place that returns a 412 status code is for `PUT /{db}` where the DB already exists.
I have added this example to the javadoc with a link in b03928d
wdyt?

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+- [IMPROVED] Improved performance by reducing the number of HTTP requests made
+  when instantiating `Database` objects
```
mikerhodes: I would be specific here. This is a bit woolly. "Remove unneeded GET request to the database when creating `Database` objects with the create option set to `false`. It offered little protection as database could be deleted, credentials changed during the use of the object."

mikerhodes:
+1

CHANGES.md
ricellis: Thanks @mikerhodes. It doesn't just apply to the `false` case though, the GET is also removed in the `true` case. How about:
```
- [IMPROVED] Remove unneeded GET request to the database when creating `Database` objects.
  Irregardless of the chosen create option it offered little protection:
  - with create `false` the database could be deleted or credentials changed during the use
    of the object.
  - with create `true` a database with the same name could be created between the GET and PUT.
```

ricellis: Addressed in c70668c.

tomblench:
+1

PR #73, opened 2015-08-13
50264 queryPage throws an exception for Views that emit value as JSON

emlaver:
*What*:
The queryPage functionality throws an exception for map functions containing a JSON array or JSON object emitted value.

*How*:
In the View class, the method JsonToObject is called to convert the JSON element for value to the specified Java object class type.  The specified class type for value is always String as seen below where the ViewResult is initialized and calls queryPage (line 365 of org.lightcouch.View class):

```
final ViewResult<JsonElement, String, T> vr = queryView(JsonElement.class, String.class, classOfT);
```

This means that the JsonToObject method will always try and convert the Row's value to a String object. And, in the case of emitted values that are JSON objects or arrays, an exception of either "Expected String but was JSON_ARRAY” or "Expected String but was JSON_OBJECT” is thrown.

To fix this, the value passed in to queryView as a String is changed to JsonElement.  This way, string manipulation or toString will not be used. Also, gson.toJson is used to guarantee that we are sending valid JSON elements through the query functionality (including building the URL).

*Testing*:
Map functions under resources/example/views were changed to have an emit value as a JSON array and a JSON object.  These maps are queried in test cases to assert that the result list is not empty and no exception is thrown.

reviewer @ricellis 
reviewer @mikerhodes

### DIFF
--- a/src/test/java/com/cloudant/tests/ViewsTest.java
+++ b/src/test/java/com/cloudant/tests/ViewsTest.java
@@ -30,7 +30,9 @@ import com.cloudant.client.api.View;
 import com.cloudant.client.api.model.Page;
 import com.cloudant.client.api.model.ViewResult;
 import com.cloudant.test.main.RequiresDB;
+import com.google.gson.Gson;
 import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 import org.junit.After;
@@ -39,8 +41,11 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.lightcouch.NoDocumentException;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 
 @Category(RequiresDB.class)
@@ -50,7 +55,7 @@ public class ViewsTest {
     private static Database db;
     private CloudantClient account;
     private static String[] testViews = {"example/doc_title", "example/creator_created",
-            "example/creator_boolean_total"};
+            "example/creator_boolean_total", "example/created_boolean_creator"};
 
     @Before
     public void setUp() {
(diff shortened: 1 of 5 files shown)


src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -468,6 +468,89 @@ public void queryPageWithStartAndEndKeyUsingNewObjectWithIntegerBooleanStringArr
...
+        init();
+
+        View view = db.view("example/boolean_creator_created")
+                .startKey(false ,"uuid",1)
+                .endKey(false ,"uuid", 2000);
+
+        Page page = null;
+        try {
+            page = view.queryPage(30, null, Object.class);
+        } catch (Exception e) {
```
ricellis: I don't think it is good for these tests to swallow exceptions. Removing this try/catch will allow the test to fail if an exception is thrown.

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -468,6 +468,89 @@ public void queryPageWithStartAndEndKeyUsingNewObjectWithIntegerBooleanStringArr
...
+    @Test
+    public void queryPageWithComplexStartEndKeyAndJsonArrayValue() {
+        init();
+
+        View query = db.view("example/creator_boolean_total")
+                .startKey("uuid" ,false, 1)
+                .endKey("uuid" ,false, 2000);
+
+        Page page = null;
+        try {
```
ricellis: Remove try/catch again.

src/test/resources/design-docs/example/views/created_boolean_creator/map.js
```diff
@@ -1,4 +1,4 @@
 function(doc) {
   emit([doc.contentArray[0].created, doc.contentArray[0].boolean,
-   doc.contentArray[0].creator], null);
-}
+   doc.contentArray[0].creator], [doc]);
+}
```
ricellis: This one is missing the terminating new line.

emlaver:
@mikerhodes @ricellis The Page's result list does not include the emitted value in the map function.  If you look in View's queryPage method (line 340), you'll see that you won't find ```rows.get(i).getValue()```.  Should this value be added to the ```page``` variable? It would make testing a bit easier as I could just pull the JSON element from value.

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in 72b66e6

mikerhodes:
+1. I can see the awkwardness with getValue() not existing. It's a bit odd, but not for fixing now.

ricellis:
Agreed, let's merge this as-is for now.
+1

PR #75, opened 2015-08-14
Made URIBuilder query replace existing parameters

ricellis:
*What*
Replace rather than append existing parameters when using the `URIBuilder.query(String name, Object value)` method 

*How*
Call `true` not `false` for the replace argument.

*Testing*
All existing tests continue to pass.

reviewer @brynh


brynh:
+1

mikerhodes:
+1

PR #78, opened 2015-08-19
50264 50739 querypage

ricellis:
*What*
Re-introduce the changes from #73 which were reverted because of tests failing against Cloudant service clusters.
Fix the pagination issues arising from using the value of `offset`, which is invalid for cluster scenarios.

*How*
As per PR #73 use `JsonElement` instead of `String` for the value when querying views in `queryPage`
Remove usage of the `offset` value in JSON results from a view.
Refactor the pagination code to pass the page number in metadata so it doesn't need to be calculated from the offset.

*Testing*
All the existing pagination and new tests introduced in #73 now pass.

reviewer @brynh 
reviewer @mikerhodes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,15 @@
+# Unreleased
+- [FIXED] Fixed issue with pagination numbering when using `queryPage` with
+  a clustered DB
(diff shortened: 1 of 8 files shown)


mikerhodes:
+1

brynh:
+1

PR #80, opened 2015-08-26
Handle doc:null in Search and View's query results

emlaver:
*What*
The java-cloudant library appears to be exploding because it's having a problem de-serialising results of a search query with include_docs=true where one of the values for the "doc" field appears to be null.

*How*
Add logic to catch null values in JsonToObject method.  If the value of the doc is null, return null.

*Tests*
Cases added in SearchTests and ViewsTest.
Assert that the results contain the doc' null value and no ClassCastException occurs.

FogBugz: 50044

reviewer @ricellis 
reviewer @mikerhodes

### DIFF
--- a/src/main/java/org/lightcouch/View.java
+++ b/src/main/java/org/lightcouch/View.java
@@ -156,10 +156,12 @@ public class View {
             List<T> list = new ArrayList<T>();
             for (JsonElement jsonElem : jsonArray) {
                 JsonElement elem = jsonElem.getAsJsonObject();
+                T t = null;
                 if (Boolean.TRUE.equals(this.includeDocs)) {
-                    elem = jsonElem.getAsJsonObject().get("doc");
+                    t = JsonToObject(gson, elem, "doc", classOfT);
+                } else {
+                    t = this.gson.fromJson(elem, classOfT);
                 }
-                T t = this.gson.fromJson(elem, classOfT);
                 list.add(t);
             }
             return list;
--- a/src/main/java/org/lightcouch/internal/CouchDbUtil.java
+++ b/src/main/java/org/lightcouch/internal/CouchDbUtil.java
@@ -87,7 +87,16 @@ final public class CouchDbUtil {
     // JSON
 
     public static <T> T JsonToObject(Gson gson, JsonElement elem, String key, Class<T> classType) {
-        return gson.fromJson(elem.getAsJsonObject().get(key), classType);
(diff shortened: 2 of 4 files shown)


ricellis:
+1

mikerhodes:
I'm finding the testing of this a little odd. The `makeQuery` etc. methods are doing manual parsing of the result string to get back the result -- isn't exactly what we are trying to test is that the production code doing this parsing doesn't fail when passed the sample responses?

I would've expected to see something like https://github.com/cloudant/java-cloudant/blob/master/src/main/java/org/lightcouch/View.java#L190-L215 being broken out into a method we could test with the given input strings. 

@emlaver, @ricellis What am I missing here that means the tests are testing what we need them too?

ricellis:
Ultimately the thing being tested is the `CouchDbUtil.JsonToObject(...)` these blocks of JSON results are essentially unit test input for that method. The `makeQuery` etc methods are just creating the JsonElement input.
An alternative would be to invoke API methods (e.g. `queryView`) with some kind of mock or reflectively set `InputStream` containing the necessary JSON.

mikerhodes:
To me, if `JsonToObject` is what exploded on `null` we can just pass a properly constructed JSON string/inputstream to that method and make sure it doesn't. The current scaffolding doesn't add much I can see, if anything it disguises what was broken.

ricellis:
The tests as currently written would not expose this issue again if, for example, we changed the production impl in future to not use `CouchDbUtil.JsonToObject` any more. So I think Mike's point is we can simplify to a more unit test style to just test `JsonToObject` or we can change the scaffolding to ensure we are testing the production methods with the necessary JSON.

emlaver:
Yea, that makes sense.  I could either:
1) Cut the tests down to just use `{"doc" null }` and pass that through `JsonToObject`.
2) Build a mock string to be passed into the production query methods. Not sure how this string would look since it would be passed into `queryForStream`.

I would go with 1) since it gets to the root of the issue and does not require as much work.

mikerhodes:
I would do (1) too. Rich and I can discuss in the morning whether it's worth doing (2). Suggest doing that and I'll look in the morning.

ricellis:
+1

mikerhodes:
Seems good. When you squash up the commits, please add a note to the commit body on what the bug was, something like "Previously, blah method threw an exception if it was passed baz. Now we foo when passed a baz."

mikerhodes:
+1 other than above request for more informative commit message.

PR #83, opened 2015-09-09
50776 1 new view api

ricellis:
*What*
A series of PRs to provide a new API to access views. This description applies to the whole feature. Each subsequent part of the feature will be merged back into this branch before the merge to master.
This introduces the public API and javadoc and key types.

*Why*
Make improvements over the existing API:

* Prevent modification of request parameters during usage
* Pagination improvements (enhanced for loop, honor start/end key limits during reverse paging)
* Stronger typing of keys (in setter methods in addition to deserialization)
* Separate design doc and view name, in place of unusual view ID
* Fewer methods for querying
* Access to `_all_docs` more in line with the endpoint location, that is from `Database` instead of view 

*How*

1. New package-info.java and javadoc examples
1. New interfaces for `ViewRequest` and `ViewResponse`
1. `ViewRequest` provides access to complete response or a single value
1. `ViewRequest`s are typed (e.g. paginated, unpaginated, multiple)
1. Access requests via a `ViewRequestBuilder` for a given database, design doc and view
1. New interfaces to separate setters for different parameter types
1. Request builder types extend multiple setter interfaces to restrict request parameters to only the applicable ones for the request
1. Parameterized typing of builders based on key types
1. Request builders build an immutable view request object
1. `Key` and `Key.ComplexKey` classes to facilitate key typing and generation of complex key parameters

*Testing*

* Port of existing view tests to new API.
* New tests for:
    * Complex key (de)serialization
    * Multiple query request
    * New tests for paging enhancements (For loop, paging with key limits etc)
    * Validation assertions

reviewer @emlaver 
reviewer @rhyshort


src/test/java/com/cloudant/tests/ComplexKeySerializationTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+public class ComplexKeySerializationTest {
+
+    private static final Gson gson = new GsonBuilder().registerTypeAdapter(Key.ComplexKey.class,
+            new ComplexKeyDeserializer()).create();
+    public static final String STR_CMPLX_KEY_JSON = "[\"dog\",\"cat\",\"mouse\"]";
+    public static final String BOOL_CMPLX_KEY_JSON = "[true,false,true]";
+    public static final String NUM_CMPLX_KEY_JSON = "[1,12,15.0,99999999999999]";
+    public static final String MIXED_CMPLX_KEY_JSON = "[\"dog\",true,1,15.0,null]";
+
```
emlaver: Should we still keep the test for double quotes and spaces in strings? i.e. \" " test " \"

ricellis: Those tests are still there in `ViewsTest`, this unit test was more focused on the different applicable types.

emlaver:
@ricellis I executed `gradlew jacadocs` and see many `Tag @link: reference not found`.  Do you see these warnings and can they be fixed?

ricellis:
Good spot emlaver - yes, I also see them on this branch. It is a side effect of pulling apart into smaller PRs to facilitate the code review. Ultimately we will need to merge the other PRs back into this branch before this branch merges to master. If you build the javadoc on the branch at https://github.com/cloudant/java-cloudant/tree/50776-4-enable-api everything should be in order.

src/main/java/com/cloudant/client/api/views/Key.java
```diff
@@ -0,0 +1,152 @@
...
+
+        Class<T> getType() {
+            return type;
+        }
+    }
+
+    /**
+     * Provides methods for adding values to build a complex key array.
+     */
+    public interface ComplexKey {
```
rhyshort: Do we really need this interface? From what I can tell from reading the code and documentation, a complex key, is an array of json values, and The `ComplexKeyImpl` class handles that. From the current implemntation, I don't see what value a user could add, espcially since, the `toString` method is being overriden to provide the JSON string.

ricellis: It isn't intended for the user to implement this `ComplexKey` interface, its purpose is to provide strong typing of complex keys without exposing a particular JSON implementation through our APIs. Originally there were additional (non-user) methods in the ComplexKeyImpl, hence the need for the interface. You're right as it currently stands I think the interface could be replaced by a renamed ComplexKeyImpl.

rhyshort: Right so if it isn't intended for a user to implement it, we should just use a `final` concrete  class  that uses composition to achieve its end, because if a user can implement it, they will attempt to.

ricellis:
After some discussion these recommendations have been made. These changes are for this PR:

* Make the ComplexKey concrete and instead of `toString()` expose a `toJson()` method for the deserialization usage
* Remove unnecessary modifiers from interface methods
* API migration examples

These changes will be in the other PRs that make up this feature:

* handle `_design/` prefixing of design doc name
* Document the loss of `InputStream` querying option in the `CHANGES.md`
* Update the `README.md` with a link to [javadoc.io](http://www.javadoc.io/doc/com.cloudant/cloudant-client)

src/test/java/com/cloudant/tests/ComplexKeySerializationTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Test;
+
+public class ComplexKeySerializationTest {
+
+    private static final Gson gson = new GsonBuilder().registerTypeAdapter(Key.ComplexKey.class,
+            new Key.ComplexKeyDeserializer()).create();
+    public static final String STR_CMPLX_KEY_JSON = "[\"dog\",\"cat\",\"mouse\"]";
+    public static final String BOOL_CMPLX_KEY_JSON = "[true,false,true]";
+    public static final String NUM_CMPLX_KEY_JSON = "[1,12,15.0,99999999999999]";
+    public static final String MIXED_CMPLX_KEY_JSON = "[\"dog\",true,1,15.0,null]";
```
rhyshort: I wonder if its worth adding a test with a complex key with a negative integer value.

ricellis: Added to mixed complex case in e16cd3b

emlaver:
@ricellis This looks very good.  +1 when you get the mixed test passing.

PR #86, opened 2015-09-14
52064 http proxy

ricellis:
*What*
Always register both `http` and `https` socket factories no matter which protocol the database is using.

*Why*
A proxy could use a different protocol to the database.

*How*
Remove protocol `if` checks and always build a socket factory `Registry` with both `http` and `https`.

*Testing*
Existing tests continue to pass; a manual test of a `http` proxy with `https` database succeeded.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/src/main/java/org/lightcouch/CouchDbClient.java
+++ b/src/main/java/org/lightcouch/CouchDbClient.java
@@ -252,37 +252,37 @@ public class CouchDbClient extends CouchDbClientBase {
         RegistryBuilder<ConnectionSocketFactory> registry = RegistryBuilder
                 .<ConnectionSocketFactory>create();
 
-            if (props.isSSLAuthenticationDisabled()) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -3,7 +3,7 @@ apply plugin: 'maven'
 apply plugin: 'signing'
 
 group = 'com.cloudant'
-version = '1.2.0'
+version = '1.2.1-SNAPSHOT'
```
tomblench: Does this mean that somebody forgot to bump the version when the last release was done?

ricellis: We didn't have a maintenance branch when I made this branch, but then I added the commit to the maintenance branch and I thought it had vanished from this one.

tomblench:
+1

rhyshort:
+1

PR #88, opened 2015-09-17
ReplicationTest improvements

ricellis:
*What*
Improve the `ReplicationTest` and `ReplicatorTest`

*Why*
There are two problems with these tests:

1. The test `com.cloudant.tests.ReplicationTest#replicatorDB` is incorrectly using a replicator document instead of triggering a replication. This is a duplication of the equivalent test method in `ReplicatorTest`.
2. The tests `com.cloudant.tests.Replicat*Test#replication_conflict` nearly always fails when running against the Cloudant service. When running against a cluster trying to instantly read back a document that was just saved may not be successful depending on which node the request gets routed to.

*How*
* Rename the `replicatorDB` test to `replicateDatabaseUsingReplicationTrigger` and change the content to use `replication` not `replicator` (which is already covered in the `ReplicatorTest` class)
* Improve resiliency of `replication_conflict` to eventual consistency by adding a `findDocumentWithRetries` to try again a few times to read back the document if it initially is not found.

*Testing*
These are test modifications that pass on a local CouchDB and against the Cloudant service.

reviewer @brynh 
reviewer @tomblench

### DIFF
--- a/src/test/java/com/cloudant/tests/ReplicationTest.java
+++ b/src/test/java/com/cloudant/tests/ReplicationTest.java
@@ -16,14 +16,14 @@ package com.cloudant.tests;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 
 import com.cloudant.client.api.CloudantClient;
 import com.cloudant.client.api.Database;
 import com.cloudant.client.api.model.ReplicationResult;
 import com.cloudant.client.api.model.ReplicationResult.ReplicationHistory;
-import com.cloudant.client.api.model.ReplicatorDocument;
-import com.cloudant.client.api.model.Response;
 import com.cloudant.client.api.model.ViewResult;
 import com.cloudant.test.main.RequiresDB;
 import com.cloudant.tests.util.Utils;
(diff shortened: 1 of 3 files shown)


src/test/java/com/cloudant/tests/ReplicationTest.java
```diff
@@ -103,33 +102,21 @@ public void replication_filteredWithQueryParams() {
     }
 
     @Test
-    public void replicatorDB() throws Exception {
+    public void replicationDB() throws Exception {
```
tomblench: The name is a bit terse and not very descriptive - `replicateDatabase` would be a bit better.

src/test/java/com/cloudant/tests/ReplicationTest.java
```diff
@@ -103,33 +102,21 @@ public void replication_filteredWithQueryParams() {
...
-
-        // find all replicator docs
-        List<ReplicatorDocument> replicatorDocs = account.replicator()
-                .findAll();
-        assertThat(replicatorDocs.size(), is(not(0)));
+        ReplicationResult result = account.replication().source(db1URI)
+                .target(db2URI).createTarget(true).trigger();
 
-        Utils.removeReplicatorTestDoc(account, response.getId());
+        assertTrue("The replication should complete ok", result.isOk());
```
tomblench: Is it possible to use `findDocumentWithRetries` to check that all of the documents have replicated successfully to the remote side?

ricellis: So this replication trigger request is synchronous and shouldn't complete until the replication has finished. The `isOk()` is returning the value of the [CouchDB reponse](http://docs.couchdb.org/en/1.6.1/api/server/common.html#single-replication) `ok` field, which should indicate the success of the replication. If the docs don't sync correctly that would suggest a core db problem, but I added a doc to the replication for some validation anyway in b6f0060.
Also as discussed I removed the obsolete version check.

src/test/java/com/cloudant/tests/ReplicationTest.java
ricellis: Changed to `replicateDatabaseUsingReplicationTrigger` in b6f0060.

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -129,4 +130,31 @@ public static ReplicatorDocument waitForReplicatorToReachStatus(CloudantClient a
     public static String generateUUID() {
         return UUID.randomUUID().toString().replace("-", "");
     }
+
+    /**
+     * When testing against a cluster requests can be routed to different nodes. This means after
+     * a save a document may not appear for all subsequent requests until an internal replication
+     * has finished between the nodes. This method tries to find a document up to maxRetries
+     * times. Even if a document is found it only means it has propagated to the node the request
+     * reached not that it has reached all nodes.
```
brynh: I think this comment should specify that the retries happen at 0.5s intervals as that's a significant part of why retrying might make it more likely to result in success and helps users realise that it might take a while to return with a high retry count.

ricellis: Added more explanation in f797174.

brynh:
+1

tomblench:
+1

PR #92, opened 2015-09-22
50046 dbcopy option

ricellis:
*What*
Cloudant provides a `dbcopy` option to store key/value reduce results in a database. The current `DesignDocument.MapReduce` API should provide a way to set this parameter.

*How*
Add `setDbCopy` method (and getter) to `MapReduce` taking a database name string.
Expose the other options from `DesignDocument` and `MapReduce` API.

*Testing*
Added a new test (in the `RequiresCloudant` category) that sets the option and validates that the db is copied as expected.

reviewer @emlaver
reviewer @brynh


emlaver:
+1

brynh:
+1

PR #95, opened 2015-09-23
52506 all docs

ricellis:
*What*
Use custom interfaces for `_all_docs` requests and responses.

*Why*
Feedback from #90 suggested that it was confusing to have `View*` interfaces to handle `_all_docs` requests.

*How*
Add interfaces for `AllDocsRequest` and `AllDocsResponse` that expose simpler methods for handling an `_all_docs` response.
Since `_all_docs` returns view like structures anyway the implementation is still backed by `View*` objects, but additionally complexity is hidden from the end user (e.g. typing as this is constant for `_all_docs`).

*Testing*
Updated `ViewsTest#allDocs()` to use new methods.

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 # Unreleased
+- [NEW] - Requests for the `_all_docs` endpoint are made via `Database#getAllDocsRequestBuilder()`
+  instead of using a view.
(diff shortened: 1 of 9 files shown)


rhyshort:
+1

emlaver:
Looks good +1

PR #96, opened 2015-09-25
52580 changes test

ricellis:
*What*
Use uniquely named databases for `ChangeNotificationsTest` to try and resolve intermittent `NoDocumentException` failures in the travis checks, which appear to be caused by overlap of a single database.

*How*
Create JUnit `ExternalResource` types for both the `CloudantClient` and `Database` objects. This allows `before` and `after` methods to be applied at either the test method or test class level by using the `@Rule` and `@ClassRule` annotations respectively. It has the advantage of keeping the setup and
teardown code for these resources in a single place, allowing for easier enforcement of test best practice (such as using a unique name for a `Database`).
Refactor the `ChangeNotificationsTest` to use these resources.

*Testing*
Existing `ChangeNotificationsTest` pass.

reviewer @tomblench 
reviewer @emlaver

### DIFF
--- a/src/test/java/com/cloudant/tests/ChangeNotificationsTest.java
+++ b/src/test/java/com/cloudant/tests/ChangeNotificationsTest.java
@@ -20,19 +20,21 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 
 import com.cloudant.client.api.Changes;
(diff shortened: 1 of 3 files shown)


emlaver:
+1

tomblench:
+1

PR #98, opened 2015-09-29
48522 unicode tests

ricellis:
*What*
Refactor the unicode tests for the new HTTP API.

*How*
Replace usage of the Apache HTTP classes in the test methods.
Fix print lines with logging.
Update the test to use the new JUnit rules.

*Testing*
The tests pass after the refactor.

reviewer @emlaver 
reviewer @tomblench

### DIFF
--- a/src/test/java/com/cloudant/tests/UnicodeTest.java
+++ b/src/test/java/com/cloudant/tests/UnicodeTest.java
@@ -16,32 +16,26 @@ package com.cloudant.tests;
 
 import static org.junit.Assert.assertEquals;
 
-import com.cloudant.client.api.CloudantClient;
 import com.cloudant.client.api.Database;
 import com.cloudant.client.api.model.Index;
 import com.cloudant.client.api.model.IndexField;
-import com.cloudant.client.api.model.IndexField.SortOrder;
 import com.cloudant.client.api.views.Key;
+import com.cloudant.http.Http;
+import com.cloudant.http.HttpConnection;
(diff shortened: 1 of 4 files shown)


src/test/java/com/cloudant/tests/UnicodeTest.java
```diff
@@ -94,21 +83,26 @@ public void tearDown() {
...
         // -the-character-set-of-the-page-with-a-head
-        ContentType contentType;
-        try {
-            contentType = ContentType.getOrDefault(entity);
-        } catch (UnsupportedCharsetException e) {
-            throw new RuntimeException("The output of " + uri + " is in charset " + e
-                    .getCharsetName() + ", which is unsupported.",
-                    e);
+
+        String contentType = connection.getConnection().getHeaderField("Content-Type");
```
emlaver: Can we not use `connection.getConnection().getContentType()`?

emlaver: There's also `connection.getConnection().getContentEncoding()`.

ricellis: Yes, `getContentType()` refers to the remote object so we can use that - fixed in 3b5db94.
`getContentEncoding()` does not get charset information for text types, it is used for entire body compression schmes (e.g. `gzip`) so we can't use that one.

tomblench:
+1

emlaver:
Looks good, +1

PR #99, opened 2015-10-01
52852 timeout defaults

ricellis:
*What*
Add default socket and connection timeouts

*Why*
Having an infinite timeout can cause the client to hang by blocking the connection pool if a response is not received.

*How*
Set 5 minute timeouts as defaults in `CouchDbProperties`.
Upgrade Apache HttpClient from 4.3.3 to 4.3.6 to get a handshake socket timeout fix.

*Testing*
Existing tests pass against local and the Cloudant service.

reviewer @rhyshort 
reviewer @brynh


rhyshort:
+1

brynh:
+1

PR #104, opened 2015-10-07
2nd PR: Replace Apache HttpClient with com.cloudant.http.Http class

emlaver:
*What*
This is the second part for replacing the http dependency from Apache’s httpclient to Java’s HttpURLConnection.  Migrating to HttpUrlConnection is required for us to use java-cloudant within sync-android.

*How*
- Replace HttpPut and RESTful commands of Apache’s httpclient with sync-android’s Http (using Java’s HttpUrlConnection)
- Replace HttpResponse with HttpConnection’s InputStream response
- Remove imports for org.apache.http
- Remove any methods that were only specific to Apache’s httpclient (i.e. connection manager)
- Remove CouchDb Android class
- Remove shutdown method from CloudantClient and CouchDbClient
- Add URI’s getUserInfo to URIBuilder and CouchDbProperties.  This is needed for setting auth property in HttpConnection
- Use sync-android’s executeToInputStream for executing HTTP calls and properly handling the response
- Add Rich's commit for intermittent problems with Replicator and Replication tests (Rich’s branch 52593-replicate-tests).
- Remove shutdown comment in javadoc
- Fix return javadoc for HTTP API
- Remove shutdown method from test cases
- Ignore SSL test cases - will enable in next PR with Rich’s interceptor branch (53085-ssl-setters)
- Ignore cookieNegativeTest - enable in next PR with cookie interceptor fixes
- Ignore Replicator tests. Currently, they succeed for CouchDB but several fail for Cloudant database.       
  Will be enabled with the PR from 52593-replicate-tests.

*Testing*
Use existing test cases. Test cases from sync-android in next PR (i.e. HttpTest)

reviewer @ricellis 
reviewer @rhyshort


src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -236,7 +237,30 @@ public CloudantClient(String account, String authCookie, ConnectOptions connectO
...
+
+        doInit(h.get("scheme"),
+                h.get("hostname"),
+                new Integer(h.get("port")).intValue(),
+                h.get("username"),
+                h.get("password"),
+                null,
+                null, //authcookie
+                interceptor
+        );
```
ricellis: I don't think we should have any API exposure of the `CookieInterceptor` here. We should also leave the constructor argument changes for the other ticket.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -247,7 +271,8 @@ public CloudantClient(String account, String authCookie, ConnectOptions connectO
      */
     public ApiKey generateApiKey() {
         URI uri = buildUri(getBaseUri()).path("_api/v2/api_keys").build();
-        return client.executeRequest(createPost(uri, "", ""), ApiKey.class);
+        HttpConnection connection = Http.POST(uri, "application/json");
```
ricellis: We should keep the usage of `HttpConnection` in as few classes as possible. We will need to set interceptors on every created `HttpConnection` and also potentially other parameters (e.g. `Proxy`). As such it would be beneficial to have all usages in one place.
So a better solution here might be to call a `post(uri)` method (e.g. the one on `CouchDatabaseBase`).

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -388,24 +403,19 @@ public String serverVersion() {
...
+        InputStream response = client.executeRequest(request);
         return response;
 
     }
 
-
-    /**
-     * Shuts down the connection manager used by this client instance.
-     */
-    public void shutdown() {
```
ricellis: I think we should leave the `shutdown()` method in place even if it is currently a no-op. It is already part of the API and it preserves it for future use.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -484,10 +503,10 @@ Response put(URI uri, Object object, boolean newEntity, int writeQuorum, Gson gs
                 assertNotEmpty(id, "id");
                 assertNotEmpty(rev, "rev");
             }
-            final HttpPut put = new HttpPut(buildUri(uri).pathToEncode(id).query("w",
-                    writeQuorum).buildEncoded());
-            CouchDbUtil.setEntity(put, json.toString(), "application/json");
-            response = client.executeRequest(put);
+            HttpConnection putConnection = Http.PUT(buildUri(uri).pathToEncode(id).query("w",
```
ricellis: Another instance where a `put()` method might be better. Although thinking about it more maybe we will be ok as we can fix up the `executeRequest` method to add the interceptors/proxy etc anyway.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -530,17 +553,31 @@ Response put(URI uri, Object object, boolean newEntity, int writeQuorum, Gson gs
      */
     private void doInit(String scheme, String hostname, int port,
                         String loginUsername, String password, ConnectOptions connectOptions,
-                        String authCookie) {
+                        String authCookie, CookieInterceptor interceptor) {
```
ricellis: Take this `CookieInterceptor` out of the constructor for now and create it from the credentials in this method. Obviously we'll need to expose setters for the interceptors later.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -211,10 +207,10 @@ private JsonObject getPermissionsObject() {
      * @return List of shards
      */
     public List<Shard> getShards() {
-        HttpResponse response = null;
-        HttpGet get = new HttpGet(buildUri(db.getDBUri()).path("/_shards").build());
+        InputStream response = null;
+        //HttpGet get = new HttpGet(buildUri(db.getDBUri()).path("/_shards").build());
```
ricellis: Remove commented out code.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -211,10 +207,10 @@ private JsonObject getPermissionsObject() {
...
         try {
-            response = client.executeRequest(get);
+            response = client.get(buildUri(db.getDBUri()).path("/_shards").build());
```
ricellis: Does this `/` before `_shards` need removing? In other places they seem to have been.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -258,10 +254,10 @@ public void createIndex(String indexName, String designDocName, String indexType
      */
     public void createIndex(String indexDefinition) {
         assertNotEmpty(indexDefinition, "indexDefinition");
-        HttpResponse putresp = null;
+        InputStream putresp = null;
         URI uri = buildUri(getDBUri()).path("_index").build();
         try {
-            putresp = client.executeRequest(createPost(uri, indexDefinition, "application/json"));
+            putresp = client.executeRequest(createPost(uri, null, "application/json"));
```
ricellis: Should this `null` be `indexDefinition`?

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -957,10 +953,14 @@ private String getFindByIndexBody(String selectorJson,
         return finalbody.toString();
     }
 
-    HttpResponse executeRequest(HttpRequestBase request) {
+    InputStream executeRequest(HttpConnection request) {
```
ricellis: This method should probably be removed, it seems like most places call `client.executeRequest()` anyway and if they don't they probably should.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -957,10 +953,14 @@ private String getFindByIndexBody(String selectorJson,
...
         return client.executeRequest(request);
     }
 
+    <T> T executeRequest(HttpConnection request, Class<T> classType) {
```
ricellis: Also should be removed.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -388,24 +403,19 @@ public String serverVersion() {
...
 
     }
 
-
-    /**
-     * Shuts down the connection manager used by this client instance.
-     */
-    public void shutdown() {
-        client.shutdown();
+    public <T> T executeRequest(HttpConnection request, Class<T> classType) {
```
ricellis: Why do we need this parameterized version? Isn't the `InputStream` version sufficient? If it isn't then shouldn't we just have the parameterized one (optionally being typed to `InputStream`).

src/main/java/com/cloudant/client/api/Search.java
```diff
@@ -119,9 +119,9 @@
     public InputStream queryForStream(String query) {
         key(query);
         URI uri = uriBuilder.build();
-        HttpGet get = new HttpGet(uri);
-        get.addHeader("Accept", "application/json");
-        return getStream(db.executeRequest(get));
+        HttpConnection get = Http.GET(uri);
+        get.getConnection().addRequestProperty("Accept", "application/json");
```
ricellis: I think the form `get.requestProperties.add(...)` is preferred.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -89,11 +56,8 @@
...
  * <p/>
  * <p/>
- * <p>At the end of a client usage; it's useful to call: {@link #shutdown()} to ensure proper
- * release of resources.
  *
  * @author Ahmed Yehia
- * @see CouchDbClientAndroid
  * @since 0.0.2
  */
 public class CouchDbClient extends CouchDbClientBase {
```
ricellis: As there is essentially nothing left here perhaps we should move the remaining `database()` method to `CouchDbClientBase` and then delete this class and rename `CouchDbClientBase` to `CouchDbClient`.
Wdyt?

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -506,6 +525,10 @@ Response put(URI uri, Object object, boolean newEntity, int writeQuorum, Gson gs
                 h.put("scheme", uri.getScheme());
                 h.put("hostname", uri.getHost());
                 h.put("port", new Integer(uri.getPort()).toString());
+                if(uri.getUserInfo() != null && !uri.getUserInfo().isEmpty()) {
```
ricellis: Why is this added here? I thought this already happened in `HttpConnection`'s execute method.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -77,85 +63,49 @@
...
+
         this.gson = GsonHelper.initGson(new GsonBuilder()).create();
         final String path = props.getPath() != null ? props.getPath() : "";
-        this.baseURI = buildUri().scheme(props.getProtocol()).host(props.getHost()).port(props
-                .getPort()).path("/").path(path).build();
-
-        // for the authentication using previous session's cookie . do not get the cookie
-        if (props.getAuthCookie() == null) {
-            getCookie(props);
+        //Add username and password if authentication info exists
```
ricellis: Is this temporary until we have the `CookieInterceptor` enabled?

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -77,85 +63,49 @@
...
-    /**
-     * @return {@link HttpContext} instance for HTTP request execution.
-     */
-    abstract HttpContext createContext();
-
-    /**
-     * Shuts down the connection manager used by this client instance.
-     */
-    abstract void shutdown();
+    // Client(s) provided implementation
```
ricellis: If  `database` method is moved into here this comment can be deleted.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -307,20 +241,29 @@ public Replicator replicator() {
...
-     * <p><b>Note</b>: The response must be closed after use to release the connection.
+     * <p><b>Note</b>: The stream must be closed after use to release the connection.
      *
-     * @param request The HTTP request to execute.
-     * @return {@link HttpResponse}
+     * @param connection The HTTP request to execute.
+     * @return Class type of object T (i.e. {@link Response}
      */
-    public HttpResponse executeRequest(HttpRequestBase request) {
+    public <T> T executeRequest(HttpConnection connection, Class<T> classType) {
```
ricellis: Are the only acceptable types here `InputStream` and `Response`?

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -330,64 +273,92 @@ public HttpResponse executeRequest(HttpRequestBase request) {
...
-     * @return {@link InputStream}
+     * @return Class type of object T (i.e. {@link Response}
      */
-    InputStream get(URI uri) {
-        HttpGet get = new HttpGet(uri);
-        get.addHeader("Accept", "application/json");
-        return get(get);
+    public <T> T get(URI uri, Class<T> classType) {
+        HttpConnection connection = Http.GET(uri);
+        if(classType.isInstance(HttpConnection.class)) {
```
ricellis: What is going on here? This seems very strange as the get/put/post/delete methods would be expected to actually do a request, but this branch appears to convert a URI into a `HttpConnection`, so why not just do `Http.GET(uri)` in the caller?

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -330,64 +273,92 @@ public HttpResponse executeRequest(HttpRequestBase request) {
...
+            return (T)connection;
+        } else {
+            //return executeToJsonObject(connection, classType);
```
ricellis: remove commented out code

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -330,64 +273,92 @@ public HttpResponse executeRequest(HttpRequestBase request) {
...
+        return executeRequest(connection);
+    }
+
+    /**
+     * Performs a HTTP PUT request, saves or updates a document.
+     *
+     * @return Input stream with response
+     */
+    InputStream put(URI uri) {
+        return put(uri, null);
```
ricellis: Improve javadoc to say this defaults to "application/json" content type and then pass "application/json" String, instead of `null`.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -330,64 +273,92 @@ public HttpResponse executeRequest(HttpRequestBase request) {
...
+    /**
+     * Performs a HTTP PUT request with content type, saves or updates a document.
+     *
+     * @return Input stream with response
+     */
+    InputStream put(URI uri, String contentType) {
+        HttpConnection connection = null;
+        if(contentType != null) {
+            connection = Http.PUT(uri, contentType);
+        } else {
```
ricellis: remove this `if/else` based on https://github.com/cloudant/java-cloudant/pull/104/files?diff=split#r41374990

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -330,64 +273,92 @@ public HttpResponse executeRequest(HttpRequestBase request) {
...
+     * @return {@link Response}
      */
-    HttpResponse head(URI uri) {
-        return executeRequest(new HttpHead(uri));
+    Response put(URI uri, InputStream instream, String contentType) {
+        HttpConnection connection;
+        try {
+            connection = Http.PUT(uri, contentType);
+
+            StringWriter writer = new StringWriter();
```
ricellis: Remove this `StringWriter` - `HttpConnection#setRequestBody(InputStream)` would be better.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -512,17 +479,114 @@ public Gson getGson() {
...
      *
-     * @param request  The HTTP request to execute.
-     * @param Class<T> class of domain object to return
-     * @return <T> T
+     * @param connection  The HTTP connection to execute.
+     * @return <T> T json object
      */
-    public <T> T executeRequest(HttpRequestBase request, Class<T> classType) {
-        HttpResponse response = null;
+    public InputStream executeRequest(HttpConnection connection) {
```
ricellis: Why do we need this method? Can't the `executeToInputStream` on line 495 just be public?

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
```diff
@@ -512,17 +479,114 @@ public Gson getGson() {
...
+        return executeToInputStream(connection);
+    }
+
+    // - if 2xx then return stream
+    // - map 404 to NoResourceException
+    // - if there's a couch error returned as json, un-marshall and throw
+    // - anything else, just throw the IOException back, use the cause part of the exception?
+
+    // it needs to catch eg FileNotFoundException and rethrow to emulate the previous exception handling behaviour
+    private InputStream executeToInputStream(HttpConnection connection) throws CouchDbException {
```
ricellis: I think there would be some value to extracting this into a class of its own and potentially with a response object (of a new type) instead of `InputStream`, but I think that can happen later.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbProperties.java
```diff
@@ -229,4 +251,39 @@ public CouchDbProperties setAuthenticatedModeSSLSocketFactory(SSLSocketFactory f
         this.authenticatedModeSSLSocketFactory = factory;
         return this;
     }
+
+    public Map<String, String> getCustomHeaders() {
```
ricellis: custom headers don't need to be supported here (we'll just use interceptors if we need a custom header).

src/main/java/com/cloudant/client/org/lightcouch/CouchDbProperties.java
```diff
@@ -229,4 +251,39 @@ public CouchDbProperties setAuthenticatedModeSSLSocketFactory(SSLSocketFactory f
...
+        return customHeaders;
+    }
+
+    public CouchDbProperties setCustomHeaders(Map<String, String> customHeaders) {
```
ricellis: custom headers don't need to be supported here either

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
```diff
@@ -211,11 +199,13 @@ public static String streamToString(InputStream in) {
     /**
      * Closes the response input stream.
      *
-     * @param response The {@link HttpResponse}
+     * @param response The {@link HttpConnection}
      */
-    public static void close(HttpResponse response) {
+    public static void close(InputStream response) {
         try {
```
ricellis: As IOUtils has been imported why not just use `IOUtils.closeQuietly()`?

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
```diff
@@ -227,7 +217,9 @@ public static void close(HttpResponse response) {
      */
     public static void close(Closeable c) {
         try {
-            c.close();
+            if(c != null) {
```
ricellis: closeQuietly again?

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
```diff
@@ -257,36 +248,37 @@ public static String getAsString(HttpResponse response, String e) {
...
-     * @param response The {@link HttpResponse}
+     * @param response The response from {@link com.cloudant.http.HttpConnection}
+     * @param klazz
      * @return {@link Response}
      */
-    public static <T> List<T> getResponseList(HttpResponse response, Gson gson, Class<T> klazz,
+    public static <T> List<T> getResponseList(InputStream response, Gson gson, Class<T> klazz,
                                               Type typeofT) throws CouchDbException {
-        InputStream instream = getStream(response);
+        //InputStream instream = getStream(response);
```
ricellis: Remove commented out code

src/main/java/com/cloudant/client/org/lightcouch/internal/URIBuilder.java
```diff
@@ -38,17 +38,26 @@
...
 
     public static URIBuilder buildUri() {
         return new URIBuilder();
     }
 
     public static URIBuilder buildUri(URI uri) {
-        URIBuilder builder = URIBuilder.buildUri().scheme(uri.getScheme())
-                .host(uri.getHost()).port(uri.getPort()).path(uri.getPath());
+        URIBuilder builder;
+        if(uri.getUserInfo() != null && !uri.getUserInfo().isEmpty()) {
```
ricellis: I'm really confused as to why the user info stuff appears in so many places. I'm pretty sure this is going to go away when the `CookieInterceptor` is enabled though right?

src/main/java/com/cloudant/client/org/lightcouch/internal/URIBuilder.java
```diff
@@ -106,6 +127,11 @@ public URIBuilder path(String path) {
         return this;
     }
 
+    public URIBuilder userInfo(String userInfo) {
+        this.userInfo += userInfo;
```
ricellis: Should this really be appending?
I feel like this should just do a set.

src/test/java/com/cloudant/tests/AttachmentsTest.java
```diff
@@ -51,7 +51,6 @@ public void setUp() {
     @After
     public void tearDown() {
         account.deleteDB("lightcouch-db-test");
-        account.shutdown();
```
ricellis: A bunch of these test changes won't be necessary if the `shutdown()` method remains in place.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: What if I move this put method to `CouchDbClientBase` class?

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Yes - currently the only return types are InputStream and Response.  Maybe it's better to just have two methods?

src/main/java/com/cloudant/client/api/Database.java
emlaver: `Search` class uses this in the `queryForStream` method and does not have a `client` object.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Yes, that's a good idea.  I wasn't completely clear of the purpose of this class and it seemed to add obscurity.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Yes, I agree.

src/main/java/com/cloudant/client/org/lightcouch/internal/URIBuilder.java
emlaver: The reason I added getUserInfo is to support the full URI with username:password in HttpConnection's execute.  If getUserInfo is null, it won't add the proper authentication and none of the HTTP requests will work:
https://github.com/cloudant/java-cloudant/blob/856de65030fdfbff6fb731b75daa4ceb77647d8b/src/main/java/com/cloudant/http/HttpConnection.java#L190

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Partial fix in e6dd315.  Left a note for tomorrow to discuss if we should move the method into `CouchDbClient` class.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Added back. Fixed in e6dd315.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/Search.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Yes - I left a note on another userInfo comment pointing to HttpConnection's execute method that uses userInfo to create authentication.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClientBase.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbProperties.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbProperties.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
emlaver: Fixed in e6dd315.

src/main/java/com/cloudant/client/org/lightcouch/internal/URIBuilder.java
emlaver: I temporarily need a way to set the username and password into the URI.  If not, the proper authentication will not be added in HttpConnection's execute.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
ricellis: Yeah, I believe the structure was such that the Android and Java sub-classes of `CouchDbClientBase` could each provide some specific behaviour whilst sharing a bunch of code from the abstract base class. Hopefully we can keep all the Android specifics in the HTTP classes so we don't need the division.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -128,10 +127,6 @@
  * <h2>Replication</h2>
  * Replication {@link Replication account.replication()} and {@link Replicator account.replicator()}
  *
- * <h2>Shutting down</h2>
- * At the end of a client usage; it's useful to call {@link #shutdown() client.shutdown()} to
- * ensure proper release of resources.
- *
```
rhyshort: This should be reverted too, even though shutdown is now a No-op. Its better to have people still call so they are used to doing it in the event we need to make a change.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -1,333 +1,592 @@
...
+     * @param properties An object containing configuration properties.
+     * @see {@link CouchDbProperties}
+     */
+    public CouchDbClient(CouchDbProperties properties) {
+        this(new CouchDbConfig(properties));
+    }
+
+    /**
+     * Shuts down the connection manager used by this client instance.
+     */
```
rhyshort: JavaDoc should be changed to reflect that we don't use a connection manager anymore

src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -66,7 +67,6 @@ public void setUp() {
     @After
     public void tearDown() {
         account.shutdown();
-        cookieBasedClient.shutdown();
```
rhyshort: I don't think this should be removed, since looking at the setup code the `cookieBasedClient` isn't the same as account.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
```diff
@@ -162,9 +164,7 @@ public static void tearDownClass() {
 
     @After
     public void tearDown() {
-        if (dbClient != null) {
-            dbClient.shutdown();
-        }
+        dbClient.shutdown();
```
rhyshort: do we still need the if null check?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -68,7 +68,6 @@ public void setUp() {
     @After
     public void tearDown() {
         account.deleteDB("lightcouch-db-test");
-        account.shutdown();
```
rhyshort: Should this have been reverted too?

PR #105, opened 2015-10-08
2nd PR: Replace Apache HttpClient with com.cloudant.http.Http class

emlaver:
*What*
This is the second part for replacing the http dependency from Apache’s httpclient to Java’s HttpURLConnection.  Migrating to HttpUrlConnection is required for us to use java-cloudant within sync-android.

*How*
- Replace HttpPut and RESTful commands of Apache’s httpclient with sync-android’s Http (using Java’s HttpUrlConnection)
- Replace HttpResponse with HttpConnection’s InputStream response
- Remove imports for org.apache.http
- Remove any methods that were only specific to Apache’s httpclient (i.e. connection manager)
- Remove CouchDb Android class
- Remove shutdown method from CloudantClient and CouchDbClient
- Add URI’s getUserInfo to URIBuilder and CouchDbProperties.  This is needed for setting auth property in HttpConnection
- Use sync-android’s executeToInputStream for executing HTTP calls and properly handling the response
- Add Rich's commit for intermittent problems with Replicator and Replication tests (Rich’s branch 52593-replicate-tests).
- Remove shutdown comment in javadoc
- Fix return javadoc for HTTP API
- Remove shutdown method from test cases
- Ignore SSL test cases - will enable in next PR with Rich’s interceptor branch (53085-ssl-setters)
- Ignore cookieNegativeTest - enable in next PR with cookie interceptor fixes
- Ignore Replicator tests. Currently, they succeed for CouchDB but several fail for Cloudant database.       
  Will be enabled with the PR from 52593-replicate-tests.

*Testing*
Use existing test cases. Test cases from sync-android in next PR (i.e. HttpTest)

reviewer @ricellis 
reviewer @rhyshort


src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -128,10 +123,6 @@
  * <h2>Replication</h2>
  * Replication {@link Replication account.replication()} and {@link Replicator account.replicator()}
  *
- * <h2>Shutting down</h2>
- * At the end of a client usage; it's useful to call {@link #shutdown() client.shutdown()} to
- * ensure proper release of resources.
- *
```
rhyshort: This should be reverted too, even though shutdown is now a No-op. Its better to have people still call so they are used to doing it in the event we need to make a change.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -1,333 +1,589 @@
...
+     * @see {@link CouchDbProperties}
+     */
+    public CouchDbClient(CouchDbProperties properties) {
+        this(new CouchDbConfig(properties));
+    }
+
+    /**
+     * Shuts down the connection manager used by this client instance.
+     */
+    public void shutdown() {
```
rhyshort: JavaDoc should be changed to reflect that we don't use a connection manager anymore

src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -66,7 +67,6 @@ public void setUp() {
     @After
     public void tearDown() {
         account.shutdown();
-        cookieBasedClient.shutdown();
```
rhyshort: I don't think this should be removed, since looking at the setup code the cookieBasedClient isn't the same as account.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
```diff
@@ -162,9 +164,7 @@ public static void tearDownClass() {
 
     @After
     public void tearDown() {
-        if (dbClient != null) {
-            dbClient.shutdown();
-        }
+        dbClient.shutdown();
```
rhyshort: do we still need the if null check?

src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -68,7 +68,6 @@ public void setUp() {
     @After
     public void tearDown() {
         account.deleteDB("lightcouch-db-test");
-        account.shutdown();
```
rhyshort: Should this have been reverted too?

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
emlaver: I'm not sure, I would think not.  @ricellis What do you think?

src/test/java/com/cloudant/tests/ViewsTest.java
emlaver: Fixed in 55cb722.

src/test/java/com/cloudant/tests/CloudantClientTests.java
emlaver: Fixed in 55cb722.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Fixed in 55cb722.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in 55cb722.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
ricellis: Strangely this test does not configure the `dbClient` in a `@Before` method, but it does appear to new up `dbClient` in each test. I do think that there is the possibility for a test to fail without setting `dbClient` though so I think the null check is worth having at least until we refactor this test with the JUnit `Rule`.

src/main/java/com/cloudant/client/api/Replication.java
```diff
@@ -47,7 +47,7 @@
 public class Replication {
     private com.cloudant.client.org.lightcouch.Replication replication;
 
-	/*public Replication(CouchDbClientBase client) {
+	/*public Replication(CouchDbClient client) {
```
ricellis: Why is this commented out?
Does it need removing or is this a TODO?

src/main/java/com/cloudant/client/org/lightcouch/Changes.java
```diff
@@ -215,7 +212,6 @@ private void setNextRow(Row nextRow) {
     }
 
     private void terminate() {
-        httpGet.abort();
```
ricellis: Have you seen any issues with this? Do we need to consider whether we can cancel a `HttpConnection` or something?

src/test/java/com/cloudant/tests/ReplicateBaseTest.java
```diff
@@ -0,0 +1,55 @@
+/*
```
ricellis: If feature-http-replacement has already been rebased on master then I guess you can skip this commit (or maybe skip it anyway as it didn't seem to help your issue).

src/main/java/com/cloudant/client/api/Replication.java
emlaver: Needs to be removed.

src/main/java/com/cloudant/client/org/lightcouch/Changes.java
emlaver: I haven't seen any issues with this.  Is it worth closing the `CouchDatabaseBase` object?

src/test/java/com/cloudant/tests/ReplicateBaseTest.java
emlaver: Your `52593-replicate-tests` branch did help - all tests pass on travis that I had originally ignored.

src/main/java/com/cloudant/client/api/Replication.java
emlaver: Fixed in ddc7aee.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
emlaver: Fixed in ddc7aee.

src/main/java/com/cloudant/client/org/lightcouch/Changes.java
ricellis: I did a bit of checking and closing the `InputStream` should be sufficient for this, so I think it is ok.

rhyshort:
+1 after my last little comment is fixed

ricellis:
Yep +1 after removing that note. This will need some squashing before merging.

PR #106, opened 2015-10-08
Fix and improve TOC

snowch:
Some of the TOC entries were broken, also the advanced options were a bit hidden before so adding them to the main TOC.

reviewer @ricellis
reviewer @rhyshort


ricellis:
+1

rhyshort:
+1 merging

PR #108, opened 2015-10-12
53035 http proxy

ricellis:
*What*
Add proxy support

*How*
Add `ConnectOptions` for proxy url, user and password.
Add a proxy auth interceptor.
Open connections using a `java.net.Proxy` when a proxy has been set.

*Testing*
Added a new test case that checks requests are sent to the proxy address and that the auth header is ok.

reviewer @rhyshort

### DIFF
--- a/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/src/main/java/com/cloudant/http/HttpConnection.java
@@ -10,15 +10,18 @@
 
 package com.cloudant.http;
 
+import com.cloudant.http.interceptors.BasicAuthInterceptor;
+
 import org.apache.commons.io.IOUtils;
 
 import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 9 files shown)


src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -332,4 +334,28 @@ public void disconnect() {
...
+        if ("http".equals(proxyAddress.getProtocol()) || "https".equals(proxyAddress.getProtocol
+                ())) {
+            this.proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyAddress.getHost(),
+                    proxyAddress.getPort()));
+        } else {
+            throw new IllegalArgumentException("Only HTTP type proxies are supported");
+        }
+    }
+
+    public static String getEncodedAuth(String userInfo) throws IOException {
```
rhyshort: Delete this method

rhyshort:
+1 after that minor comment is addressed

src/main/java/com/cloudant/http/HttpConnection.java
ricellis: Fixed in e511f35

emlaver:
+1

PR #110, opened 2015-10-13
53034 timeout interceptor

ricellis:
*What*
Enable timeout customization for `HttpConnection`

*How*
Created a `TimeoutCustomizationInterceptor`.
Changed `ConnectOptions` for timeouts.
Constructed a `TimeoutCustomizationInterceptor` from the options and added it to the list of interceptors.

*Testing*
New tests for read and connect timeouts.

reviewer @emlaver 
reviewer @brynh

### DIFF
--- a/src/test/java/com/cloudant/tests/CloudantClientTests.java
+++ b/src/test/java/com/cloudant/tests/CloudantClientTests.java
@@ -21,30 +21,44 @@ import static org.junit.Assert.assertTrue;
 import com.cloudant.client.api.CloudantClient;
 import com.cloudant.client.api.Database;
 import com.cloudant.client.api.model.ApiKey;
+import com.cloudant.client.api.model.ConnectOptions;
 import com.cloudant.client.api.model.Membership;
 import com.cloudant.client.api.model.Task;
 import com.cloudant.client.org.lightcouch.CouchDbException;
 import com.cloudant.client.org.lightcouch.NoDocumentException;
 import com.cloudant.http.AgentHelper;
+import com.cloudant.http.interceptors.TimeoutCustomizationInterceptor;
 import com.cloudant.test.main.RequiresCloudant;
 import com.cloudant.test.main.RequiresCloudantService;
 import com.cloudant.test.main.RequiresDB;
 import com.cloudant.tests.util.SingleRequestHttpServer;
+import com.cloudant.tests.util.TestLog;
 
-import org.apache.commons.logging.Log;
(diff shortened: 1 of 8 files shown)


src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -208,4 +223,109 @@ public void existingDatabaseCreateException() {
...
+        }
+    }
+
+    /**
+     * Checks that the read timeout works. The test sets a read timeout of 0.25 s and the mock
+     * server waits 0.5 s before continuing to send data down the stream so the timeout should be
+     * triggered.
+     */
+    @Test(expected=SocketTimeoutException.class)
+    public void readTimeout() throws Throwable {
```
brynh: As `SocketTimeoutException` can be thrown for various timeout reasons, I wonder if we should add a couple of additional tests very similar to `connectionTimeout()` and `readTimeout()`, which instead check the converse of these - i.e. we don't get a `SocketTimeoutException` if the connect/reads happen within the timeout. This would give a little extra confidence we're getting the `SocketTimeoutException`s for the reasons we expect to get them in these tests.

ricellis: A reasonable suggestion - I was also concerned about the `SocketTimeoutException` ambiguity and I was originally checking the content of the exception message for appropriate info (i.e. `connect timed out` and `read timed out`). Sadly I discovered that the OpenJDK used on Travis always says `Read timed out`. It is a shame they don't have different exception classes for this.
I'll try and add some negative cases.

ricellis: @brynh what about addressing this by setting a default timeout on our test client instances to be 1 minute? This is smaller than the 5 minute default (which is probably a good thing for test runs anyway), but long enough that it should expose if there are other socket timeouts going on during the test run besides the ones we're manufacturing for these tests.

brynh: @ricellis OK, sounds like a reasonable approach I think, and I'm not sure how much added value we really get from doing anything more complex. As you said above, it's just a shame there's no way to differentiate between the errors that works for all JVMs.

ricellis: Changed default timeouts for tests in 470204f.

brynh:
+1

emlaver:
+1

PR #111, opened 2015-10-14
Improve responseToInputStream exception handling

emlaver:
*What*
Improve CouchDbClient's `responseToInputStream` exception handling.

*How*
Throw `PreconditionFailedException` when response status code is 412.

In else block, if `fromJson` deserialization fails:
Catch the exception and create a new `CouchDbException` with the response message and code. 

Or, if deserialization returns null:
Create a new `CouchDbException` in `finally` block.

reviewer @ricellis
reviewer @tomblench

### DIFF
--- a/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
+++ b/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
@@ -30,10 +30,10 @@ import com.cloudant.http.HttpConnectionResponseInterceptor;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
 import com.google.gson.JsonParser;
 import com.google.gson.reflect.TypeToken;
 
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -93,11 +93,11 @@ public class CouchDbClient {
         this.requestInterceptors = new ArrayList<HttpConnectionRequestInterceptor>();
         this.responseInterceptors = new ArrayList<HttpConnectionResponseInterceptor>();
 
-        if(props.getRequestInterceptors() != null) {
+        if (props.getRequestInterceptors() != null) {
             this.requestInterceptors.addAll(props.getRequestInterceptors());
         }
 
-        if(props.getResponseInterceptors() != null) {
+        if (props.getResponseInterceptors() != null) {
             this.responseInterceptors.addAll(props.getResponseInterceptors());
         }
 
@@ -284,7 +284,7 @@ public class CouchDbClient {
                 throw e;
             }
         } finally {
-            closeQuietly(is);
+            close(is);
         }
     }
 
@@ -365,9 +365,9 @@ public class CouchDbClient {
 
     /**
      * Performs a HTTP PUT request, saves or updates a document.
-     * @param object Object for updating request
-     * @param newEntity If true, saves a new document. Else, updates an existing one.
      *
+     * @param object    Object for updating request
+     * @param newEntity If true, saves a new document. Else, updates an existing one.
      * @return {@link Response}
      */
     public Response put(URI uri, Object object, boolean newEntity) {
@@ -376,9 +376,9 @@ public class CouchDbClient {
 
     /**
      * Performs a HTTP PUT request, saves or updates a document.
-     * @param object Object for updating request
-     * @param newEntity If true, saves a new document. Else, updates an existing one.
      *
+     * @param object    Object for updating request
+     * @param newEntity If true, saves a new document. Else, updates an existing one.
      * @return {@link Response}
      */
     public Response put(URI uri, Object object, boolean newEntity, int writeQuorum) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -516,47 +516,30 @@ public InputStream executeToInputStream(HttpConnection connection) throws CouchD
...
+                    ex = getGson().fromJson(new InputStreamReader(es),
+                            CouchDbException.class);
+                } catch(Exception e) {
+                    ex = new CouchDbException(response, code);
+                } finally {
+                    if (ex == null) {
                         ex = new CouchDbException(response, code);
                     }
-
+                    ex.setStatusCode(code);
```
ricellis: Should this just be moved to after [line 533](https://github.com/cloudant/java-cloudant/pull/111/files#diff-d4cc4d9c9440a68731ce18c3b9f0555cR533)?
The other branches of `ex` (lines 535, 538) already construct with the response `code` so I don't see the need to set it again for those two cases - am I missing something?

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -516,47 +516,30 @@ public InputStream executeToInputStream(HttpConnection connection) throws CouchD
...
-                        //Create new Exception from scratch
+                CouchDbException ex = null;
+                try {
+                    es = connection.getConnection().getErrorStream();
+                    ex = getGson().fromJson(new InputStreamReader(es),
+                            CouchDbException.class);
+                } catch(Exception e) {
+                    ex = new CouchDbException(response, code);
+                } finally {
+                    if (ex == null) {
```
tomblench: I don't think this block is needed because you already set the value for `ex` in the `catch` block

ricellis: I guess it depends if anything in the `try` can generate something other than an `Exception` i.e. a subclass of `Throwable` or `Error`.

tomblench: I thought about that - so just catch `Throwable` then

ricellis: Yes, I agree that works better.

ricellis:
The code style also needs to be applied to this - I noticed a bunch of changes when I was using the class in another branch.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -516,47 +516,30 @@ public InputStream executeToInputStream(HttpConnection connection) throws CouchD
...
-                    try {
-                        ex = getGson().fromJson(errorJsonMessage,
-                                CouchDbException.class);
-                        ex.setStatusCode(code);
-                    } catch (NullPointerException e) {
-                        //Create new Exception from scratch
+                CouchDbException ex = null;
+                try {
+                    es = connection.getConnection().getErrorStream();
+                    ex = getGson().fromJson(new InputStreamReader(es),
```
ricellis: we should wrap this in a not `null` check for the error stream

emlaver: Fixed in 593f0c5.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Fixed in 593f0c5 - replaced exception with `JsonParseException`.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Fixed in 593f0c5.

ricellis:
How about:
```java
             else {
                CouchDbException ex = new CouchDbException(response, code);
                es = connection.getConnection().getErrorStream();
                if (es != null) {
                    try {
                        ex = getGson().fromJson(new InputStreamReader(es),
                                CouchDbException.class);
                        ex.setStatusCode(code);
                    } catch (JsonParseException e) {
                        //supress and just throw ex momentarily
                    }
                }
                throw ex;
            }
```

emlaver:
@ricellis Yes, the code flow is easier to read.  Fixed in 673005d.

tomblench:
Can you use `closeQuietly` here? https://github.com/cloudant/java-cloudant/pull/111/files#diff-d4cc4d9c9440a68731ce18c3b9f0555cR548

Also what is the difference between `closeQuietly` and `com.cloudant.client.org.lightcouch.internal.CouchDbUtil.close`?

ricellis:
+1

emlaver:
@tomblench There isn't much difference - `CouchDbUtil.close` uses `IOUtils`.  What if I remove the `closeQuietly` and just call the util `close` method?

tomblench:
@emlaver good idea

emlaver:
@tomblench Fixed in b619185.

tomblench:
+1

PR #112, opened 2015-10-16
3rd PR: Add cookie interceptor from sync-android's com.cloudant.http package

emlaver:
*What*
This is the third part of the PR series for replacing the http dependency from Apache’s httpclient to Java’s HttpURLConnection. This part focuses on updating and enabling the cookie interceptor.  

*How*
- Add cookie interceptor in CloudantClient
- Remove CouchDbProperties constructor with username and password
- Use request and response interceptors for adding cookie
- Remove getters for username and password
- Remove cookieNegativeTest
- Add HttpTest class with cookie auth test case
- Add category RequiresCloudant to testCookieAuthWithoutRetry test case
- Remove all calls to URI's userInfo from lightcouch classes
- Remove getUserInfo in HttpConnection
- Remove current auth cookie code
- Improved error handling when PR 111 is complete
- Update code style

*Testing*
Add cookie authentication test from sync-android’s HttpTest.

reviewer @ricellis
reviewer @tomblench

### DIFF
--- a/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -32,10 +32,12 @@ import com.cloudant.client.api.views.Key;
 import com.cloudant.client.org.lightcouch.Changes;
 import com.cloudant.client.org.lightcouch.CouchDbClient;
 import com.cloudant.client.org.lightcouch.CouchDbDesign;
+import com.cloudant.client.org.lightcouch.CouchDbException;
 import com.cloudant.client.org.lightcouch.CouchDbProperties;
 import com.cloudant.client.org.lightcouch.Replication;
 import com.cloudant.client.org.lightcouch.Replicator;
 import com.cloudant.client.org.lightcouch.Response;
+import com.cloudant.http.CookieInterceptor;
 import com.cloudant.http.HttpConnection;
 import com.cloudant.http.interceptors.ProxyAuthInterceptor;
 import com.cloudant.http.interceptors.SSLCustomizerInterceptor;
@@ -125,7 +127,7 @@ import java.util.Map;
  *
  * <h2>Replication</h2>
  * Replication {@link Replication account.replication()} and {@link Replicator account.replicator()}
- * 
+ *
  * @author Mario Briggs
  * @since 0.0.1
  */
@@ -159,8 +161,7 @@ public class CloudantClient {
                 new Integer(h.get("port")).intValue(),
                 loginUsername,
                 password,
-                null, //connectoptions
-                null //authcookie
+                null //connectoptions
         );
     }
 
@@ -187,8 +188,7 @@ public class CloudantClient {
                 new Integer(h.get("port")).intValue(),
                 loginUsername,
                 password,
-                connectOptions,
-                null //authcookie
+                connectOptions
         );
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -517,18 +511,29 @@ public InputStream executeToInputStream(HttpConnection connection) throws CouchD
 
         try {
             JsonObject errorJsonMessage = null;
-            if(cause != null && connection.getConnection().getErrorStream() != null) {
+            if (cause != null && connection.getConnection().getErrorStream() != null) {
```
ricellis: This is all removed by #111 right?

ricellis:
This is quite confusing because of the expected changes from #111 with exception handling in `CouchDbClient`. I think I'd like to take another look after the merge with those changes, otherwise this seems OK to me.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Yes, the PR was rebased in a9648e1.

emlaver:
@ricellis I've merged in #111 to the feature branch and rebased.

src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.experimental.categories.Category;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.URL;
+
+public class HttpTest {
+
+    private CloudantClient account;
```
ricellis: As this is a new test it might be nice to use the JUnit `ExternalResource`. If we eventually refactor all the tests to use it then any changes to how we obtain clients for tests can be made in just one place. I also don't think this needs a new client before each test (although there is only one test at the moment) so might be worth using the `ClassRule` e.g.
```java
@ClassRule
public static CloudantClientResource clientResource = new CloudantClientResource();
private CloudantClient account = clientResource.get();
```
and the `setUp` and `tearDown` methods could then be removed.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -498,20 +492,17 @@ Response put(URI uri, Object object, boolean newEntity, int writeQuorum) {
...
-                        String loginUsername, String password, ConnectOptions connectOptions,
-                        String authCookie) {
+                        String loginUsername, String password, ConnectOptions connectOptions) {
 
-        CouchDbProperties props;
-        if (authCookie == null) {
-            props = new CouchDbProperties(scheme, hostname, port,
-                    loginUsername, password);
+        CouchDbProperties props = new CouchDbProperties(scheme, hostname, port);
+        if (loginUsername != null && password != null) {
```
tomblench: Should we warn/error if user is null but password isn't or vice-versa?

emlaver: Yea, we could use `CouchDbException` and throw a message like:
"Invalid username and/or password.  Please check the credentials and try again."

src/test/java/com/cloudant/tests/HttpTest.java
emlaver: Fixed in 71cbec2

ricellis:
I think this could do with new tests for the new exception cases - I've got some I can add to branch 48522-service-test-fixes which needs to merge on this anyway.
Otherwise +1

tomblench:
+1 after the fix we discussed.

PR #114, opened 2015-10-16
Added default ports

### LINKED ISSUE #113: Default ports are not used when creating client instance from a URL
Creating a `CloudantClient` instance from a URL without specifying a port results in the port being set to `-1` which is invalid e.g.
```java
new CloudantClient("http://192.0.2.0", "user", "pass");
```
It is possible to workaround by specifying the port e.g.
```java
new CloudantClient("http://192.0.2.0:80", "user", "pass");
```
It shouldn't be necessary to specify the default port.

ricellis:
*What*
Set the port to a default value if it is not supplied in the URL
Fixes #113

*How*
Check if the port value was supplied in the URL and if not get the default for the protocol.

*Testing*
Added new test to assert default ports for `http` and `https` URLs.

reviewer @emlaver 
reviewer @rhyshort


rhyshort:
+1

emlaver:
+1

PR #115, opened 2015-10-16
48522 refactor test server

ricellis:
*What*
Refactor the simple http server we use for some tests.

*Why*
The use cases had diverged, resulting in a few different implementations, but with a lot of common code. With new interceptors the `SingleHttpRequestServer` was not great for testing because there could be multiple requests.

*How*
Create a `SimpleHttpServer` and subclass `SimpleHttpsServer` that can serially service single HTTP requests and provide an `serverAction` method that can be overridden for test specific behaviour. Also make the server an `ExternalResource` so it can optionally be used as  JUnit `Rule`. Also improve over the previous server by dynamically allocating a port which may avoid issues on some test systems.

*Testing*
Existing tests refactored to use the new `SimpleHttpServer` and continue to pass.

reviewer @emlaver 
reviewer @brynh


src/test/java/com/cloudant/tests/util/SimpleHttpServer.java
```diff
@@ -0,0 +1,210 @@
...
+            log.fine("Server stopping");
+        } catch (Exception e) {
+            log.log(Level.SEVERE, SimpleHttpServer.class.getName() + " exception", e);
+        } finally {
+            semaphore.release();
+        }
+    }
+
+    /**
+     * Wait up to 2 mintues for the server, longer than that and we give up on the test
```
brynh: Typo "mintues" -> "minutes"

ricellis: Fixed in 2ded07a

src/test/java/com/cloudant/tests/util/SimpleHttpServer.java
```diff
@@ -0,0 +1,210 @@
...
+        BufferedWriter w = new BufferedWriter(new OutputStreamWriter(os));
+        //note the HTTP spec says the status line must be followed by a CRLF
+        //and an empty line must separate headers from optional body
+        //so basically we need 2 CRLFs
+        w.write("HTTP/1.0 200 OK\r\n\r\n");
+        w.flush();
+    }
+
+    /**
+     * Read the input stream lines into the specified List
```
brynh: There is no list specified in the args. Maybe this should say "Read the input stream lines into {@link #lines}"

src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -266,67 +276,40 @@ public void connectionTimeout() throws Throwable {
...
-                    startupLatch.countDown();
-                    //close resources
-                    IOUtils.closeQuietly(w);
-                    IOUtils.closeQuietly(socket);
-                    IOUtils.closeQuietly(serverSocket);
-                    cleanupLatch.countDown();
-                }
+            protected void serverAction(InputStream is, OutputStream os) throws Exception {
+                //sleep for longer than the read timeout
+                Thread.sleep(READ_TIMEOUT * 2);
```
brynh: The test says "the mock server waits 0.5 s before continuing to send data down the stream", but there is now no data sent by the server. Should this call `super.serverAction(is, os)` after the `Thread.sleep`? I guess in reality it doesn't really matter if the mock server actually sends data or not, but either the comment should be changed or the code so it matches the description.

ricellis: Yeah there is no need for the server to actually do anything, fixed the comment in 2557419

src/test/java/com/cloudant/tests/util/SimpleHttpServer.java
ricellis: The field was private so I updated the comment to point to the `getLastInputRequestLines` method in 2557419.

brynh:
+1

emlaver:
+1

PR #116, opened 2015-10-20
48522 service test fixes

ricellis:
*What*
A series of fixes for the tests to ensure they pass against the Cloudant service

*How*
Ensure interceptors are invoked by using `client.executeRequest` instead of calling `execute` directly on `HttpConnection` objects.
Fix the `Proxy-Authorization` header and make sure the test actually does an assert.
Remove redundant code from the SSL tests.
Fixed intermittent issues in the replication tests.
New tests for not specifying both user and password.

*Testing*
The tests pass against local CouchDB and the Cloudant service.

reviewer @emlaver 
reviewer @tomblench

### DIFF
--- a/src/main/java/com/cloudant/http/interceptors/BasicAuthInterceptor.java
+++ b/src/main/java/com/cloudant/http/interceptors/BasicAuthInterceptor.java
@@ -33,7 +33,7 @@ public class BasicAuthInterceptor implements HttpConnectionRequestInterceptor {
         this(userinfo, "Authorization");
     }
 
-    public BasicAuthInterceptor(String userinfo, String authHeader) {
+    protected BasicAuthInterceptor(String userinfo, String authHeader) {
         this.encodedAuth = encodedCreds(userinfo);
         this.authHeader = authHeader;
     }
@@ -41,7 +41,7 @@ public class BasicAuthInterceptor implements HttpConnectionRequestInterceptor {
     @Override
     public HttpConnectionInterceptorContext interceptRequest(HttpConnectionInterceptorContext
                                                                      context) {
-        context.connection.requestProperties.put("Authorization", String.format("Basic %s",
+        context.connection.requestProperties.put(authHeader, String.format("Basic %s",
                 encodedAuth));
         return context;
     }
--- a/src/test/java/com/cloudant/tests/HttpProxyTest.java
+++ b/src/test/java/com/cloudant/tests/HttpProxyTest.java
@@ -14,12 +14,12 @@
 
 package com.cloudant.tests;
 
-import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.createPost;
-import static junit.framework.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static junit.framework.TestCase.assertEquals;
+import static junit.framework.TestCase.assertTrue;
 
 import com.cloudant.client.api.CloudantClient;
 import com.cloudant.client.api.model.ConnectOptions;
+import com.cloudant.http.Http;
 import com.cloudant.http.interceptors.ProxyAuthInterceptor;
 import com.cloudant.tests.util.SimpleHttpServer;
 
@@ -50,12 +50,12 @@ public class HttpProxyTest {
         //create a client with a bogus address (TEST-NET)
         String mockProxyUser = "alpha";
         String mockProxyPass = "alphaPass";
-        CloudantClient client = new CloudantClient("http://192.0.2.0", "", "", new ConnectOptions()
+        CloudantClient client = new CloudantClient("http://192.0.2.0", new ConnectOptions()
                 .setProxyURL(new URL(server.getUrl()))
                 .setProxyUser(mockProxyUser)
                 .setProxyPassword(mockProxyPass));
 
-        client.executeRequest(createPost(client.getBaseUri(), null, "application/json"));
+        client.executeRequest(Http.GET(client.getBaseUri()));
         //if it wasn't a 20x then an exception should have been thrown by now
 
         //assert that the request had the expected proxy auth header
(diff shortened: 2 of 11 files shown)


emlaver:
LGTM +1

src/main/java/com/cloudant/http/interceptors/BasicAuthInterceptor.java
```diff
@@ -41,7 +41,7 @@ public BasicAuthInterceptor(String userinfo, String authHeader) {
     @Override
     public HttpConnectionInterceptorContext interceptRequest(HttpConnectionInterceptorContext
                                                                      context) {
-        context.connection.requestProperties.put("Authorization", String.format("Basic %s",
+        context.connection.requestProperties.put(authHeader, String.format("Basic %s",
```
tomblench: I know that the `authHeader` field wasn't added by this PR - but why would you ever want to change the value of this string?

src/test/java/com/cloudant/tests/HttpProxyTest.java
```diff
@@ -50,12 +50,12 @@ public void proxyConfiguration() throws Exception {
         //create a client with a bogus address (TEST-NET)
         String mockProxyUser = "alpha";
         String mockProxyPass = "alphaPass";
-        CloudantClient client = new CloudantClient("http://192.0.2.0", "", "", new ConnectOptions()
+        CloudantClient client = new CloudantClient("http://192.0.2.0:80", new ConnectOptions()
```
tomblench: Why do we have to specify port 80? might be worth a comment if the reason isn't obvious

src/main/java/com/cloudant/http/interceptors/BasicAuthInterceptor.java
ricellis: The subclass that does proxy auth needs to make it `Proxy-Authorization` otherwise the code is identical so it seemed to make sense to share it. Maybe there is a more elegant way to change the value, or maybe the constructor that takes the value should be `protected` instead of public?

src/test/java/com/cloudant/tests/HttpProxyTest.java
ricellis: Good spot, that was working around a problem where default ports were not used (resulting in -1 for the port, that is now fixed so I can update).

src/test/java/com/cloudant/tests/util/Utils.java
```diff
@@ -66,13 +73,27 @@ public static String getHostName(String account) {
...
-        //Grab replicator doc revision using HTTP GET command
+        //Grab replicator doc revision using HTTP HEAD command
         String replicatorDb = "_replicator";
         URI uri = URI.create(account.getBaseUri() + replicatorDb + "/"
                 + replicatorDocId);
         HttpConnection head = Http.HEAD(uri);
-        String revision = head.execute().getConnection().getHeaderField("ETag");
 
+        //add a response interceptor to allow us to retrieve the ETag revision header
+        final AtomicReference<String> revisionRef = new AtomicReference<String>();
```
tomblench: I'm not familiar with the `AtomicReference` class but is it needed? I thought the method in the anonymous `HttpConnectionResponseInterceptor` class would capture the `revisionRef` variable and that would be sufficient.

src/main/java/com/cloudant/http/interceptors/BasicAuthInterceptor.java
tomblench: Ah no that makes sense - didn't realise you used that header as well.

src/test/java/com/cloudant/tests/HttpProxyTest.java
```diff
@@ -63,12 +63,10 @@ public void proxyConfiguration() throws Exception {
...
             if (line.contains("Proxy-Authorization")) {
                 foundProxyAuthHeader = true;
-                Matcher m = Pattern.compile("Proxy-Authorization\\s*:\\s*(.*)", Pattern
+                Matcher m = Pattern.compile("Proxy-Authorization: Basic (.*)", Pattern
                         .CASE_INSENSITIVE).matcher(line);
                 assertTrue("The Proxy-Authorization header should match the pattern", m.matches());
-                assertEquals("There should be 2 groups, one for the header key and one for the " +
-                        "value", 2, m.groupCount());
-
+                assertEquals("There should be 1 group for the value", 1, m.groupCount());
```
tomblench: Why did the group count change? I thought you got one group for the whole matching string and another for the bracketed subgroup, hence the original 2 - is that wrong?

src/test/java/com/cloudant/tests/util/Utils.java
ricellis: Honestly this was just for conciseness to get a getter/setter for free on a final field so the `HttpConnectionResponseInterceptor` instance used here could remain anonymous.
The alternative is the significantly more verbose declaration of an inner (local) class that is a named subclass of `HttpConnectionResponseInterceptor` that includes a getter method to return the `revision` string, which I'm willing to do it if it is preferred or happy to receive suggestions for other alternatives.

src/test/java/com/cloudant/tests/HttpProxyTest.java
ricellis: Yes, this surprised me also. You do indeed get 1 group for the whole matching string and another for the capturing group. However, a closer inspection of the javadoc revealed that the `groupCount()` method excludes the whole match. So although `group(0)` returns the whole match and `group(1)` returns the value `groupCount()` returns 1.

tomblench: OK, did this test fail before? What's the setup for doing test against a proxy?

src/main/java/com/cloudant/http/interceptors/BasicAuthInterceptor.java
ricellis: Made the constructor protected anyway in 5f432e9.

src/test/java/com/cloudant/tests/HttpProxyTest.java
ricellis: Fixed in a1b062f.

src/test/java/com/cloudant/tests/HttpProxyTest.java
ricellis: The test did not fail before because the `Proxy-Authentication` header wasn't being added due to the bug in `BasicAuthInterceptor` that didn't use the header key from the subclass. This made the test bypass all the matching code and the lack of an assertion on the boolean `foundProxyAuthHeader` meant it didn't fail even though the header was missing. These changes mean the test would now fail for the lack of header or broken regex.
The test uses the `SimpleHttpServer` as a proxy to verify that the requests are routed to the proxy and contain the expected auth header, but it doesn't actually do any real proxying of requests.

tomblench: OK good to know we have tests to guard against any future regressions.

tomblench:
+1

PR #120, opened 2015-10-27
48522 5 tidy up

ricellis:
*What*
Complete the Apache HttpClient replacement work

*How*
* Removed Apache HttpClient dependency from gradle
* Removed Apache commons-logging
* Added support for deserializing the error stream into `CouchDbException` subclasses, which in some cases adds additional useful info to the exception.
* Removed the code in `HttpConnection` that was disabling the JVM persistent http connections

*Testing*
* Minor test code cleanups
* Added new test for `Expect: 100-continue` behaviour
* Changed test JVM to `openjdk7`

reviewer @emlaver 
reviewer @alfinkel

### DIFF
--- a/src/main/java/com/cloudant/client/api/Database.java
+++ b/src/main/java/com/cloudant/client/api/Database.java
@@ -20,7 +20,6 @@ import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.createPost
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getAsString;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getResponse;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getResponseList;
-import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.setEntity;
 import static com.cloudant.client.org.lightcouch.internal.URIBuilder.buildUri;
 
 import com.cloudant.client.api.model.DbInfo;
@@ -58,9 +57,6 @@ import com.google.gson.JsonParser;
 import com.google.gson.JsonPrimitive;
 import com.google.gson.reflect.TypeToken;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
@@ -74,6 +70,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.logging.Logger;
 
 
 /**
@@ -86,7 +83,7 @@ import java.util.Set;
  */
 public class Database {
 
-    static final Log log = LogFactory.getLog(Database.class);
+    static final Logger log = Logger.getLogger(Database.class.getCanonicalName());
     private CouchDatabase db;
     private CloudantClient client;
     private final URI apiV2DBSecurityURI;
@@ -261,7 +258,7 @@ public class Database {
             if (result.equalsIgnoreCase("created")) {
                 log.info(String.format("Created Index: '%s'", indexDefinition));
             } else {
-                log.warn(String.format("Index already exists : '%s'", indexDefinition));
+                log.warning(String.format("Index already exists : '%s'", indexDefinition));
             }
         } finally {
             close(putresp);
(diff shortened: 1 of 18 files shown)


src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -86,7 +83,7 @@
  */
 public class Database {
 
-    static final Log log = LogFactory.getLog(Database.class);
+    static final Logger log = Logger.getLogger(Database.class.getName());
```
alfinkel: I think you might want to use `Database.class.getCanonicalName()` here instead of `Database.class.getName()` since it seems to be more appropriate for logging operations.  See:  http://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav.

src/main/java/com/cloudant/client/api/Search.java
```diff
@@ -85,7 +83,7 @@
  */
 public class Search {
 
-    private static final Log log = LogFactory.getLog(Search.class);
+    private static final Logger log = Logger.getLogger(Search.class.getName());
```
alfinkel: Same comment as https://github.com/cloudant/java-cloudant/pull/120/files#r43167135

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -43,7 +41,7 @@
  */
 public abstract class CouchDatabaseBase {
 
-    static final Log log = LogFactory.getLog(CouchDatabase.class);
+    static final Logger log = Logger.getLogger(CouchDatabase.class.getName());
```
alfinkel: Same comment as https://github.com/cloudant/java-cloudant/pull/120/files#r43167135

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -60,7 +58,7 @@
 
 public class CouchDbClient {
 
-    static final Log log = LogFactory.getLog(CouchDbClient.class);
+    static final Logger log = Logger.getLogger(CouchDbClient.class.getName());
```
alfinkel: Same comment as https://github.com/cloudant/java-cloudant/pull/120/files#r43167135

src/main/java/com/cloudant/client/org/lightcouch/CouchDbConfig.java
```diff
@@ -28,7 +26,7 @@
  * @author Ahmed Yehia
  */
 class CouchDbConfig {
-    private static final Log log = LogFactory.getLog(CouchDbConfig.class);
+    private static final Logger log = Logger.getLogger(CouchDbConfig.class.getName());
```
alfinkel: Same comment as https://github.com/cloudant/java-cloudant/pull/120/files#r43167135

src/main/java/com/cloudant/client/org/lightcouch/Replication.java
```diff
@@ -58,7 +57,7 @@
  */
 public class Replication {
 
-    static final Log log = LogFactory.getLog(Replication.class);
+    static final Logger log = Logger.getLogger(Replication.class.getName());
```
alfinkel: Same comment as https://github.com/cloudant/java-cloudant/pull/120/files#r43167135

emlaver:
@ricellis There is still a TODO comment in `CloudantClientTests` for a client initialized with the cookie. I believe we can remove the `cookieBasedClient` variable and all its references since there's no difference between it and `account`:
https://github.com/cloudant/java-cloudant/blob/feature-http-replacement/src/test/java/com/cloudant/tests/CloudantClientTests.java#L62-L70

src/main/java/com/cloudant/client/api/Database.java
ricellis: I think using `getName()` would be the same in these cases anyway, but I have no objection to using `getCanonicalName()` and making things easier for any future additions. Fixed in 0bbde07.

ricellis:
@emlaver I fixed the TODO in 24f5313.
I also noticed that because of the interceptors and cookie changes there was a bunch of un-needed code left hanging around for `CouchDbClient` configuration, which I have removed in 6caa93e98f6761fdff610f08802e6fa60b6167f3

alfinkel:
+1

emlaver:
Looks good - thanks for catching the other fixes.  +1

PR #128, opened 2015-10-29
Added retries to dbcopy test

### LINKED ISSUE #124: com.cloudant.tests.DesignDocumentsTest#dbcopy fails intermittently
The `dbcopy` test can fail intermittently due to eventual consistency when running against the Cloudant service. It is nice to assert that the correct number of documents were copied, so we can solve this by adding a retry loop.

ricellis:
*What*
Fixes #124 - intermittent failures in dbcopy test

*How*
Added retries to the document verification in the copied database to allow time for cluster replication to complete.

*Testing*
`com.cloudant.tests.DesignDocumentsTest#dbCopy` is updated and passes.

reviewer @emlaver


emlaver:
+1

PR #129, opened 2015-10-30
Logical Cloudant client constructors

emlaver:
*What*
Re-design CloudantClient constructors with new client builder that can handle full account URL or account name.  Builder also includes setting HTTP interceptors and connection options (connect options removed). Example:

```
ClientBuilder.url(new URL(“http://localhost”))
	.setConnectionTimeout(ClientBuilder.TimeoutOption
		.minutes(1))
	.setReadTimeout(ClientBuilder.TimeoutOption
		.minutes(3))
	.build();
```
With Cloudant account:
```
ClientBuilder.account(“emlaver”)
	.username(“emlaver")
	.password(“password”)
	.build();
```

*How*
- ClientBuilder class to replace client constructor
   - Use URL or account name for initial client
   - Remove public CloudantClient constructors
   - Call new protected constructor from ClientBuilder
   - Move setter method for connection options from ConnectOptions
   - Move TimeoutOption inner class to ClientBuilder
- Remove ConnectOptions class
- Move default values for connect options and interceptors into ClientBuilder
- Remove account and getter method from CloudantClient
- Add helper methods for http(s) test address testing and creating URL for simple http server
- Add SERVER_URL in client helper
- Add HttpConnectionInterceptor super class
- Extend request and response interceptor to new super class

*Testing*
Update existing test cases with new ClientBuilder design.

reviewer @ricellis
reviewer @rhyshort


src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import javax.net.ssl.SSLSocketFactory;
+
+public class ClientBuilder {
+
+     /** Default of 5 minutes for connection timeout **/
+    public static final long CONNECTION_TIMEOUT = 5l;
```
ricellis: I don't think these two extra fields are providing much value, can we not just have:
```java
/**Connection timeout defaults to 5 minutes**/
public static final TimeoutOption DEFAULT_CONNECTION_TIMEOUT = TimeoutOption.minutes(5l);
```
and similarly for the read timeout?

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+    /** Default of 5 minutes for read timeout **/
+    public static final long READ_TIMEOUT = 5l;
+
+    /** Default max of 6 connections **/
+    public static final int MAX_CONNECTIONS = 6;
```
ricellis: `DEFAULT_MAX_CONNECTIONS` to be consistent with the naming of the timeouts.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import javax.net.ssl.SSLSocketFactory;
+
+public class ClientBuilder {
```
ricellis: This class is the first place people are going to interact with the client. We need some javadoc explaining how to use the builder.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+    private String proxyPassword;
+    private boolean isSSLAuthenticationDisabled;
+    private SSLSocketFactory authenticatedModeSSLSocketFactory;
+    /** Defaults to {@link #CONNECTION_TIMEOUT} **/
+    private TimeoutOption connectionTimeout = DEFAULT_CONNECTION_TIMEOUT;
+    /** Defaults to {@link #READ_TIMEOUT} **/
+    private TimeoutOption readTimeout = DEFAULT_READ_TIMEOUT;
+
+
+    public static ClientBuilder account (String account) {
```
ricellis: javadoc

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+        return new ClientBuilder(account);
+    }
+
+    public static ClientBuilder url(URL url) {
```
ricellis: javadoc

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+        return new ClientBuilder(url);
+    }
+
+    public ClientBuilder(String account) {
+        //parse account into a url
+        assertNotEmpty(account, "accountName");
+        try {
+            this.url = URIBuilder.buildUri()
```
ricellis: Should we make this URL then call `return url(url)`? It doesn't offer huge advantages, but means we can have the exception handling in just one place, which might be easier if we change it later.
I guess we also wouldn't need to hardcode the port because we would get the default for `https` too.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+                url.getHost(), url.getPort());
+
+        //Create cookie interceptor
+        if (this.username != null && this.password != null) {
+            //make interceptor if both username and password are not null
+
+            //Create cookie interceptor and set in HttpConnection interceptors
+            CookieInterceptor cookieInterceptor = new CookieInterceptor(username, password);
+
+            props.addRequestInterceptors(cookieInterceptor);
```
ricellis: Maybe just call `interceptors(cookieInterceptor)` here then we always get some coverage of the sub-class checking.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+            if (username != null || password != null) {
+                //Username and password both have to contain values
+                throw new CouchDbException("Either a username and password must be provided, or " +
+                        "both values must be null. Please check the credentials and try again.");
+            }
+        }
+
+
+        //If setter methods for read and connection timeout are not called, default values are used.
+        props.addRequestInterceptors(new TimeoutCustomizationInterceptor(connectionTimeout,
```
ricellis: `interceptors(...)`

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+                requestInterceptors.add((HttpConnectionRequestInterceptor) interceptor);
+            }
+            if (interceptor instanceof HttpConnectionResponseInterceptor) {
+                responseInterceptors.add((HttpConnectionResponseInterceptor) interceptor);
+            }
+        }
+        return this;
+    }
+
+    /** Sets the connection timeout for timeout interceptor.
```
ricellis: We don't need to expose the implementation detail (timeout interceptor) in the public javadoc. It would be better to say what the connection timeout is and provide an example. Could probably link to java.net.URLConnection#setConnectTimeout() for a description of what it is.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+     * Defaults to {@link #CONNECTION_TIMEOUT}
```
ricellis: I think this should point to the `DEFAULT_CONNECTION_TIMEOUT` since the method takes a `TimeoutOption`

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+     **/
+    public ClientBuilder setConnectionTimeout(TimeoutOption connectionTimeout) {
+        this.connectionTimeout = connectionTimeout;
+        return this;
+    }
+
+    /** Sets the read timeout for timeout interceptor.
+     * Defaults to {@link #CONNECTION_TIMEOUT}
```
ricellis: The same applies as for `setConnectionTimeout()`, but this has the wrong default.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+     **/
+    public ClientBuilder setReadTimeout(TimeoutOption readTimeout) {
+        this.readTimeout = readTimeout;
+        return this;
+    }
+
+   public final static class TimeoutOption {
```
ricellis: This is going to need javadoc too, now that it will be public.

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -141,85 +132,17 @@
...
-        super();
-        Map<String, String> h = parseAccount(account);
-
-        doInit(h.get("scheme"),
-                h.get("hostname"),
-                new Integer(h.get("port")).intValue(),
-                null, null, connectOptions);
+        // set a custom gsonbuilder that includes additional cloudant deserializers
+        if(gsonBuilder != null) {
+            couchDbClient.setGsonBuilder(gsonBuilder);
```
ricellis: The `setGsonBuilder` method in this class needs to be made private (or the content could potentially be moved into the `ClientBuilder.setGsonBuilder()` method instead).

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+        if(responseInterceptors != null) {
+            for(HttpConnectionResponseInterceptor responseInterceptor : responseInterceptors) {
+                props.addResponseInterceptors(responseInterceptor);
+            }
+        }
+
+        return new CloudantClient(props, gsonBuilder);
+    }
+
+    public ClientBuilder username(String username) {
```
ricellis: Should this be `setUsername` to be consistent with the other fields, or should the other fields lose their `set` prefixes?

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+        this.username = username;
+        return this;
+    }
+
+    public ClientBuilder password(String password) {
```
ricellis: `setPassword`?

src/main/java/com/cloudant/client/api/model/ConnectOptions.java
```diff
@@ -1,150 +0,0 @@
...
-        return this;
-    }
-
-    public ConnectOptions setProxyPassword(String proxyPassword) {
-        this.proxyPassword = proxyPassword;
-        return this;
-    }
-
-    /**
-     * Sets whether hostname verification, certificate chain validation,
```
ricellis: Could rescue some of this javadoc for the `ClientBuilder` methods.

src/main/java/com/cloudant/http/interceptors/TimeoutCustomizationInterceptor.java
```diff
@@ -50,12 +42,11 @@ public TimeoutCustomizationInterceptor() {
...
      */
-    public TimeoutCustomizationInterceptor(TimeoutOption connectTimeout, TimeoutOption
+    public TimeoutCustomizationInterceptor(ClientBuilder.TimeoutOption connectTimeout,
+                                           ClientBuilder.TimeoutOption
             readTimeout) {
-        this.connectTimeout = (connectTimeout != null) ? connectTimeout.asIntMillis() :
-                DEFAULT_TIMEOUT.asIntMillis();
-        this.readTimeout = (readTimeout != null) ? readTimeout.asIntMillis() : DEFAULT_TIMEOUT
-                .asIntMillis();
+        this.connectTimeout = connectTimeout.asIntMillis();
```
ricellis: These could NPE now, but without a local default value to use this gets a bit awkward. Maybe set to `-1` and then don't apply a value later if it is `<0` unless you can think of something better, or we just doc that neither supplied `TimeoutOption` should be `null`.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
```diff
@@ -77,11 +78,10 @@ public void localSslAuthenticationDisabled() throws Exception {
 
         server.await();
 
-        ConnectOptions connectionOptions = new ConnectOptions();
-        connectionOptions.setSSLAuthenticationDisabled(true);
+        dbClient = ClientBuilder.url(new URL(server.getUrl()))
```
ricellis: `SimpleHttpsServer extends SimpleHttpServer` so this (and the other tests in this class) can use `CloudantClientHelper.newSimpleHttpServerClient(server)` as well can't they?

src/test/java/com/cloudant/tests/CloudantClientHelper.java
```diff
@@ -36,16 +36,17 @@
     public static final String COUCH_PASSWORD;
     public static final String SERVER_URI;
     public static final String SERVER_URI_WITH_USER_INFO;
+    public static final URL SERVER_URL;
```
ricellis: Doesn't this make the `SERVER_URI` obsolete? We can delete the field `SERVER_URI` and just use a local to make `SERVER_URL` to help avoid accidental usage of the wrong one?

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Moved into ClientBuilder in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/client/api/model/ConnectOptions.java
emlaver: Fixed in 4b37363

src/main/java/com/cloudant/http/interceptors/TimeoutCustomizationInterceptor.java
emlaver: I've initialized the variables to -1, and added a note about throwing an exception when null in 4b37363.  Let me know what you think.

src/test/java/com/cloudant/tests/SslAuthenticationTest.java
emlaver: Fixed in 4b37363

src/test/java/com/cloudant/tests/CloudantClientHelper.java
emlaver: Fixed in 4b37363

emlaver:
@ricellis Some of the javadocs will need fixing/re-wording.  Also, still need to rebase.

ricellis:
I agree we'll need to do more work on the javadoc, but other than that and [this comment](https://github.com/cloudant/java-cloudant/pull/129#commitcomment-14150201) I think this is good.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,527 @@
...
+     * Constructs a new instance of this class and connects to the cloudant server using the URL.
+     *
+     * @param url       For cloudant.com, the cloudant account to connect to. For Cloudant
+     *                      local, the server URL. i.e. example.cloudant.com
+     */
+    public static ClientBuilder url(URL url) {
+        return new ClientBuilder(url);
+    }
+
+    ClientBuilder(String account) throws MalformedURLException {
```
ricellis: Make `private`.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,527 @@
...
+        //parse account into a url
+        assertNotEmpty(account, "accountName");
+        url(URIBuilder.buildUri()
+                .scheme("https")
+                .host(account + ".cloudant.com")
+                .build().toURL());
+    }
+
+
+    ClientBuilder(URL url) {
```
ricellis: Make `private`.

src/main/java/com/cloudant/client/api/CloudantClient.java
ricellis: It doesn't look like this change happened? Is there a commit missing?

ricellis:
@emlaver I've done a pass on the javadoc in https://github.com/cloudant/java-cloudant/commit/fca7221f7a8fa334ab94d560966dddadef031e6b
If you are happy with it then pick it into this branch.

emlaver:
@ricellis Yes, your changes are great.  I will cherry pick this commit.

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 7a614ca

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in 7a614ca

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+    /**
+     * Constructs a new ClientBuilder for building a CloudantClient instance to connect to the
+     * Cloudant server with the specified account.
+     *
+     * @param account the Cloudant account name to connect to e.g. "example" is the account name
+     *                for the "example.cloudant.com" endpoint
+     * @return a new ClientBuilder for the account
+     * @throws MalformedURLException if the specified account name forms an invalid endpoint URL
+     */
+    public static ClientBuilder account(String account) throws MalformedURLException {
```
rhyshort: Should this really be a MalformedURLException? Constructing a client via an account name should never cause a MalformedURLException, since we are building the URL from the name provided. An IllegalArugmentException would be more appropriate  in the event the Cloudant Account provided cannot form a valid URL.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+     * Cloudant server with the specified URL.
+     *
+     * @param url server URL e.g. "https://yourCloudantLocalAddress.example"
+     * @return a new ClientBuilder for the account
+     */
+    public static ClientBuilder url(URL url) {
+        return new ClientBuilder(url);
+    }
+
+    private ClientBuilder(String account) throws MalformedURLException {
```
rhyshort: I think this constructor should go and its function folded into the  the `ClientBuilder.account` method which calls into the `ClientBuilder.url` method, which then calls down to the private constructor `ClientBuilder(URL url)`

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+
+        try {
+            //Remove user credentials from url
+            this.url = URIBuilder.buildUri()
+                    .scheme(url.getProtocol())
+                    .host(url.getHost())
+                    .port(port)
+                    .build().toURL();
+        } catch (MalformedURLException e) {
+            throw new CouchDbException(e);
```
rhyshort: As part of the bigger picture to eventually have checked exceptions, I think this should be specifically a runtime exception. We shouldn't hit the MalformedURLException, since the url is the same that was provided to us, minus the user info. So I think we should just throw a RuntimeException directly.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    /**
+     * Sets an optional proxy password for the proxy user specified by
+     * {@link #setProxyUser(String)}.
+     *
+     * @param proxyPassword password for the proxy server user
+     * @return this ClientBuilder object for setting additional options
+     */
+    public ClientBuilder setProxyPassword(String proxyPassword) {
```
rhyshort: So a general discussion point. I think we could drop the `set` prefix from all the methods I think it reads better (at least in my head), thoughts @ricellis @emlaver @mikerhodes

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+     * <P>
+     * The default is {@code setSSLAuthenticationDisabled(false)} meaning that hostname
+     * verification and certificate chain validation is done using the JVM default settings.
+     * </P>
+     *
+     * @param disabled true to disable, defaults to false
+     * @return this ClientBuilder object for setting additional options
+     * @see #setAuthenticatedModeSSLSocketFactory(SSLSocketFactory)
+     */
+    public ClientBuilder setSSLAuthenticationDisabled(boolean disabled) {
```
rhyshort: If we go the way for removing the `set` prefix I think we should change this to `disableSSLAuthentication()`, although it would mean this builder couldn't be reused to create a more secure client by turning it back on again, I think its fine, since there isn't a way to clear the interceptors anyway. So I think the user would probably want a whole new builder so they can control which interceptors run where.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,526 @@
...
+        this.isSSLAuthenticationDisabled = disabled;
+        return this;
+    }
+
+    /**
+     * Specifies the custom SSLSocketFactory to use when connecting to Cloudant over a
+     * <code>https</code> URL, when SSL authentication is enabled.
+     * <P>
+     * Note: this will have no effect if {@code setSSLAuthenticationDisabled(true)} is used to
+     * disable SSL authentication.
```
rhyshort: Shouldn't a Custom SSLSocketFactory override the sslAuthenticationOption? Not the other way around? or am I confused about which way round this all works?

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: I've been used to and haven't had issues with the common practice of adding the `set` prefix.  I thought the readability improved with setters like `setUsername` and `setPassword`.

src/main/java/com/cloudant/client/api/ClientBuilder.java
ricellis: The doc is correct, I was looking at the code yesterday, but I also did feel this was a bit counter intuitive. I wonder if we should actually check within the builder to ensure they are mutually exclusive and throw an exception, rather than assigning priority to one or the other.

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Using `new URL(...)` or `toURL()` with no try/catch or exception in the method signature will show an error saying that MalformedURLException is unhandled.

src/main/java/com/cloudant/client/api/ClientBuilder.java
ricellis: I don't mind as long as they are consistent, but it appears that people generally prefer not having the `set` prefix.

emlaver: Yea, that's fine - I'll make the changes.

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: You're right - with the builder design, it doesn't make sense for the user to try and undo disabling the SSL authentication.

src/main/java/com/cloudant/client/api/ClientBuilder.java
rhyshort: No, you catch it in the method so the code looks something like:

```java
try {
    URL url = new URL(String.format("https://%s.cloudant.com",account));
    return ClientBuilder.url(url);
} catch (MalformedURLException e) {
    throw new IllegalArgumentException("Could not generate url from account name",e);
}

File: src/main/java/com/cloudant/client/api/ClientBuilder.java (hunk 27 above)
Comment (emlaver) [inline src/main/java/com/cloudant/client/api/ClientBuilder.java]: Fixed in 6702620

File: src/main/java/com/cloudant/client/api/ClientBuilder.java (hunk 26 above)
Comment (emlaver) [inline src/main/java/com/cloudant/client/api/ClientBuilder.java]: Fixed in 6702620

File: src/main/java/com/cloudant/client/api/ClientBuilder.java (hunk 22 above)
Comment (emlaver) [inline src/main/java/com/cloudant/client/api/ClientBuilder.java]: Ok I see, thanks for the sample code.

Comment (emlaver) [inline src/main/java/com/cloudant/client/api/ClientBuilder.java]: Fixed in b80fd10

File: src/main/java/com/cloudant/client/api/ClientBuilder.java (hunk 25 above)
Comment (emlaver) [inline src/main/java/com/cloudant/client/api/ClientBuilder.java]: Fixed in 6702620

File: src/main/java/com/cloudant/client/api/ClientBuilder.java (hunk 23 above)
Comment (emlaver) [inline src/main/java/com/cloudant/client/api/ClientBuilder.java]: Fixed in b80fd10

File: src/main/java/com/cloudant/client/api/ClientBuilder.java [hunk 28]
```diff
@@ -0,0 +1,531 @@
...
+    /**
+     * Flag to disable hostname verification and certificate chain validation. This is not
+     * recommended, but for example could be useful for testing with a self-signed certificate.
+     * <P>
+     * The SSL authentication is enabled by default meaning that hostname verification
+     * and certificate chain validation is done using the JVM default settings.
+     * </P>
+     *
+     * @return this ClientBuilder object for setting additional options
+     * @throws IllegalStateException when SSLSocketFactory is not null
```
ricellis: I think this should be more about what options have been called and not what the internal check is. Maybe something like:
`@throws IllegalStateException if {@link #authenticatedModeSSLSocketFactory(SSLSocketFactory)} has been called on this ClientBuilder`

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,531 @@
...
+     * @see #authenticatedModeSSLSocketFactory(SSLSocketFactory)
+     */
+    public ClientBuilder disableSSLAuthentication() {
+        if(authenticatedModeSSLSocketFactory == null) {
+            this.isSSLAuthenticationDisabled = true;
+        } else {
+            throw new IllegalStateException("SSLSocketFactory is not null. Please " +
```
ricellis: Again I think we should make this message more helpful to the user, who may not know that the internal implementation is a `null`.
Something like:
`throw new IllegalStateException("Cannot disable SSL authentication when a custom SSLSocketFactory has been set")`

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,531 @@
...
+     * disable SSL authentication.
+     * </P>
+     *
+     * @param factory An SSLSocketFactory, or <code>null</code> for the
+     *                default SSLSocketFactory of the JRE.
+     * @return this ClientBuilder object for setting additional options
+     * @throws IllegalStateException when SSL authentication is disabled
+     * @see #disableSSLAuthentication()
+     */
+    public ClientBuilder authenticatedModeSSLSocketFactory(SSLSocketFactory factory) {
```
ricellis: Should we consider renaming this?
`customSSLSocketFactory`?
I think it was named this previously to indicate that the `SSLSocketFactory` only applied in the "mode" where SSL authentication was not disabled, but since these changes are making them mutually exclusive I don't think the name makes sense any more.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,531 @@
...
+        }
+        return this;
+    }
+    
+
+    /**
+     * Specifies the custom SSLSocketFactory to use when connecting to Cloudant over a
+     * <code>https</code> URL, when SSL authentication is enabled.
+     * <P>
+     * Note: this will have no effect if {@code disableSSLAuthentication()} is used to
```
ricellis: This note needs to be removed, based on the fact it will now throw the `IllegalStateException`.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,531 @@
...
+     *                default SSLSocketFactory of the JRE.
+     * @return this ClientBuilder object for setting additional options
+     * @throws IllegalStateException when SSL authentication is disabled
+     * @see #disableSSLAuthentication()
+     */
+    public ClientBuilder authenticatedModeSSLSocketFactory(SSLSocketFactory factory) {
+        if(!isSSLAuthenticationDisabled) {
+            this.authenticatedModeSSLSocketFactory = factory;
+        } else {
+            throw new IllegalStateException("SSL authentication is disabled. Please enable" +
```
ricellis: `throw new IllegalStateException("Cannot use a custom SSLSocketFactory when SSL authentication is disabled.")`

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Yes, I think renaming to `customSSLSocketFactory` would be better.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,531 @@
...
+                    .host(url.getHost())
+                    .port(port)
+                    .build().toURL();
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Build the {@link CloudantClient} instance based on endpoint used to construct this client
```
ricellis: whoops sorry, this doesn't quite parse 
>based on **the** endpoint

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in eb2ad70

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in eb2ad70

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in eb2ad70

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in eb2ad70

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in eb2ad70

src/main/java/com/cloudant/client/api/ClientBuilder.java
emlaver: Fixed in eb2ad70

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Full fix in d6201f4

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,545 @@
...
+    private TimeoutOption readTimeout = DEFAULT_READ_TIMEOUT;
+
+    /**
+     * Constructs a new ClientBuilder for building a CloudantClient instance to connect to the
+     * Cloudant server with the specified account.
+     *
+     * @param account the Cloudant account name to connect to e.g. "example" is the account name
+     *                for the "example.cloudant.com" endpoint
+     * @return a new ClientBuilder for the account
+     * @throws MalformedURLException if the specified account name forms an invalid endpoint URL
```
ricellis: needs to be updated to `IllegalStateException`

rhyshort: You mean `IllegalArgumentException`

emlaver: Fixed in 28d3f57

ricellis:
The `customGsonDeserializerTest` should probably call `shutdown` in a `finally` block (just in case we add any functionality back to that method later). Otherwise +1.

src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -0,0 +1,545 @@
...
+     *      .interceptors(new BasicAuthInterceptor("yourUsername:yourPassword"))
+     *      .build()
+     * }
+     * </pre>
+     *
+     * @param interceptors one or more HttpConnectionInterceptor objects
+     * @return this ClientBuilder object for setting additional options
+     * @see HttpConnectionInterceptor
+     */
+    public ClientBuilder interceptors(HttpConnectionInterceptor... interceptors) {
```
rhyshort: I think we need to add a version of this that takes a list, so we don't force the user to reduce a possible list of interceptors to an array. Thoughts? @emlaver @ricellis

emlaver: Just to clarify, would you consider we have both an array and list method? I'd be fine with that.

ricellis: I'm not drastically opposed to having two methods, but given we frequently raise concerns about maximum method counts (for Android use cases) is it really necessary? The conversion from `List` to array is not that hard:
`inteceptors(list.toArray(new HttpCollectionInterceptor[list.size()])`

Additionally it is even easier to call:
```
builder.interceptors(a).interceptors(b).build();
```
than it is to add them to a `List`:
```
List<HttpConnectionInterceptor> list = new ArrayList<>();
list.add(a);
list.add(b);
builder.interceptors(list).build();
```

rhyshort: I don't think its necessary, just a nice to have, I'm fine with leaving it with just a var args method, we can always add it later if users ask for it.

rhyshort:
:+1: +1 LGTM

ricellis:
@emlaver assuming you are happy with the commits I added then I think this is ready for squash and merge.
+1

emlaver:
Yes, the changes you've made are great - thank you.

PR #130, opened 2015-11-02
121 connection pools

### LINKED ISSUE #121: Connection pooling support for 2.0
After replacing Apache HttpClient with `HttpUrlConnection` connection pooling is relying on the JVM wide persistent connection settings.
It would be preferable to use the previous style of configuring connection pooling on a per `CloudantClient` instance basis.

ricellis:
*What*
Add support for HTTP connection pooling per `CloudantClient` using `HttpConnection` - fixes #121 

*Why*
The default implementation of `HttpUrlConnection` is pooled at a JVM level and controlled via `http.maxConnections`. If changing the `CloudantClient` connection options altered the JVM wide settings there could be unintended consequences for other applications. Having a per `CloudantClient` instance connection pool is in line with the previous max connections option and allows additional flexibility over a JVM wide setting.

*How*
* Add an optional dependency for OkHttp's URL connection.
* Add a runtime check for the `com.squareup.okhttp.OkUrlFactory` class.
* Provide a `HttpUrlConnectionFactory` interface and default implementation in `HttpConnection`.
* Create an implementation of `HttpUrlConnectionFactory` that utilizes `OkUrlFactory` to use when OkHttp is available.
* If OkHttp is not present then all `CloudantClient` instances running in the JVM will share the JVM  `HttpUrlConnection` connection pool.

*Testing*
Existing tests continue to pass with the exception of `com.cloudant.tests.HttpTest#expectContinue`, which is now `@Ignore` annotated, until an OkHttp fix is available.

reviewer @rhyshort 
reviewer @emlaver

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,11 @@ dependencies {
     compile group: 'com.google.code.gson', name: 'gson', version: '2.2.4'
     compile group: 'commons-io', name: 'commons-io', version: '2.4'
     compile group: 'commons-codec', name: 'commons-codec', version: '1.6'
+    //needed to compile, but optional for consumers of java-cloudant
+    compile(group: 'com.squareup.okhttp', name: 'okhttp-urlconnection', version: '2.5.0') {
+        ext.optional = true
+    }
+    //test dependencies
     testCompile group: 'junit', name: 'junit', version: '4.12'
 }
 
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -20,6 +20,7 @@ repositories {
     mavenCentral()
 }
 dependencies {
+    compile('com.squareup.okhttp:okhttp-urlconnection:2.5.0') { ext.optional = true }
```
rhyshort: Small Stylistic thing, can you make ths so its 

```groovy
compile group: 'com.squareup.okhttp' name:'okhttp-urlconnection' version:'2.5.0' { 
    ext.optional = true 
}
```

rhyshort:
All we need (as discussed) is a mention on the README about okhttp and Java compatibility

build.gradle
ricellis: Fixed in cafda7b.

ricellis:
Added a note about Java 7 and OkHttp in the README in cafda7b.

rhyshort:
+1

emlaver:
+1

PR #131, opened 2015-11-06
122 doc updates

### LINKED ISSUE #122: Readme & doc updates for 2.0
The `README.md`, `CHANGES.md` and javadoc need to be slimmed down and up-to-date for a 2.0 release.

The API reference should be moved entirely into javadoc so the examples don't need to be maintained in two places.
A javadoc `overview.html` and some `package-info.java` classes are needed.
There are also a lot of javadoc warnings that need resolution so we can maintain the javadoc in a clean state.

ricellis:
*What*
Refactor documentation for next milestone.
In addition to reviewing the code changes I think for review it would be valuable to run `./gradlew clean javadoc` and then look at the built javadoc starting from `build/docs/javadoc/index.html`.

*Why*
Currently the `README.md` contains an API reference. This is confusing because it relates to the master branch, not the latest released version. There is also the problem of needing to update code samples in both the `README` and the javadoc, which means that they can easily get out of sync. A number of documentation updates were also required for recent changes.
Fixes #122.

*How*
Updated `CHANGES.md` with information
Add an `overview.html` to provide the "getting started" context in javadoc.
Slim down `README.md` to include links to other content and one simple example.
Move all API reference examples into javadoc for classes or methods.
Fix all javadoc warnings and fail builds if any warnings are introduced.
Removed javadoc that provided no additional value over the method signature.
Fail builds if any compile warnings are introduced.
Add the new http package to javadoc and refactor some HTTP class packaging to not expose them as API in the javadoc.
Removed deprecated methods.
Removed `Database.batch()` method as use of the `batch=ok` parameter is not recommended.

*Testing*
This is mostly documentation changes, but the `DocumentsCRUDTest.batch()` was removed as the code it was testing is no longer present.
Existing tests continue to pass.

reviewer @tomblench 
reviewer @emlaver

### DIFF
--- a/overview.html
+++ b/overview.html
@@ -1,0 +1,473 @@
+<html>
+<!--
+This overview.html provides Javadoc overview for the application.
+The content from the body section is added to the Javadoc overview page.
+There is no need to add a head/title as that is provided by Javadoc.
+-->
+<body>
+
+<P>
+    This is the <a href="https://github.com/cloudant/java-cloudant" target="_blank">official
+    Cloudant library for java.</a>
+</P>
+
+<h1>Contents</h1>
+<OL>
+    <LI><a href="#Compatibility">Compatibility</a></LI>
+    <LI><a href="#Initialization">Initialization</a>
+        <OL>
+            <LI><a href="#Cloudant service example">Cloudant service example</a></LI>
+            <LI><a href="#Cloudant Local example">Cloudant Local example</a></LI>
+        </OL>
+    </LI>
+    <LI><a href="#Authentication">Authentication</a>
+        <OL>
+            <LI><a href="#Cookie authentication">Cookie authentication</a></LI>
+            <LI><a href="#Custom authentication">Custom authentication</a></LI>
+        </OL>
+    </LI>
+    <LI><a href="#Capabilities">Capabilities</a>
+        <OL>
+            <LI><a href="#Server Operations">Server Operations</a></LI>
+            <LI><a href="#Database Operations">Database Operations</a></LI>
+            <LI><a href="#Document Operations">Document Operations</a>
+                <OL>
+                    <LI><a href="#Single document CRUD">Single document CRUD</a></LI>
+                    <LI><a href="#Bulk document CRU">Bulk document CRU</a></LI>
+                </OL>
+            </LI>
+            <LI><a href="#Attachments">Attachments</a>
+                <OL>
+                    <LI><a href="#Standalone Attachments">Standalone Attachments</a></LI>
+                    <LI><a href="#Inline Attachments">Inline Attachments</a></LI>
+                </OL>
+            </LI>
+            <LI><a href="#Design Documents">Design Documents</a></LI>
+            <LI><a href="#Using Views">Using Views</a></LI>
+            <LI><a href="#Cloudant Query">Cloudant Query</a></LI>
+            <LI><a href="#Cloudant Search">Cloudant Search</a></LI>
+        </OL>
+    </LI>
+    <LI><a href="#Advanced Configuration">Advanced Configuration</a>
+        <OL>
+            <LI><a href="#Connection options">Connection options</a></LI>
+            <LI><a href="#Resource sharing">Resource sharing</a>
+                <OL>
+                    <LI><a href="#Thread safety">Thread safety</a></LI>
+                    <LI><a href="#Connection pooling">Connection pooling</a>
+                        <OL>
+                            <LI><a href="#Default pooling">Default pooling</a></LI>
+                            <LI><a href="#With OkHttp">With OkHttp</a></LI>
+                        </OL>
+                    </LI>
+                </OL>
+            </LI>
+            <LI><a href="#J2EE">J2EE</a></LI>
+        </OL>
+    <LI><a href="#Project">Project</a></LI>
+</OL>
+
+<h1 id="Compatibility">Compatibility</h1>
+
+<P>
+    This library can be used with the following databases
+</P>
+<UL>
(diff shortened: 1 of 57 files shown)


overview.html
```diff
@@ -0,0 +1,452 @@
+<html>
+<!--
+This overview.html provides Javadoc overview for the application.
+The content from the body section is added to the Javadoc overview page.
+There is no need to add a head/title as that is provided by Javadoc.
+-->
+<body>
+
+<P>
+    This is the official Cloudant library for java.
```
tomblench: Maybe a link to the homepage on github?

tomblench: Oh and also something about how it can be used with couch, do we have anything which discusses that already - plain couch vs Cloudant features etc?

overview.html
```diff
@@ -0,0 +1,452 @@
...
+</P>
+
+<OL>
+    <LI><a href="#Initialization">Initialization</a>
+        <OL>
+            <LI><a href="#Cloudant service example">Cloudant service example</a></LI>
+            <LI><a href="#Cloudant Local example">Cloudant Local example</a></LI>
```
tomblench: I know we call this Cloudant Local internally, but I think it has some long official name like IBM Cloudant Data Layer, Local Edition - should we use that, at least at first and thereafter abbreviated to Cloudant Local?

overview.html
```diff
@@ -0,0 +1,452 @@
...
+        </OL>
+    </LI>
+    <LI><a href="#Authentication">Authentication</a>
+        <OL>
+            <LI><a href="#Cookie authentication">Cookie authentication</a></LI>
+            <LI><a href="#Custom authentication">Custom authentication</a></LI>
+        </OL>
+    </LI>
+    <LI><a href="#Functions">Functions</a>
```
tomblench: Functions or methods?

overview.html
```diff
@@ -0,0 +1,452 @@
...
+<UL>
+    <LI>Create/Update - {@link com.cloudant.client.api.Database#bulk(List)}</LI>
+    <LI>Read - {@link com.cloudant.client.api.views.AllDocsRequestBuilder} via {@link
+        com.cloudant.client.api.Database#getAllDocsRequestBuilder()}
+    </LI>
+</UL>
+
+<h2 id="Attachment Functions">Attachment Functions</h2>
+
+See the <a href="https://docs.cloudant.com/attachments.html">Cloudant documentation</a> for more
```
tomblench: `href target="_blank"` to open in a new window? In the three pane view this currently loads the content in the main pane which doesn't look right

overview.html
```diff
@@ -0,0 +1,452 @@
...
+        Foo foo = new Foo(); // Foo extends Document
+        foo.addAttachment("attachment.txt", attachment);
+        db.save(foo);
+    }
+</pre>
+<P>Note that attachment data is <strong>not</strong> included by default when reading documents.
+    To
+    retrieve inline attachment data use {@link
+    com.cloudant.client.api.model.Params#attachments()}.
+    Seeh {@link com.cloudant.client.api.Database#find(java.lang.Class, java.lang.String,
```
tomblench: typo on seeh

overview.html
```diff
@@ -0,0 +1,452 @@
...
+    view_ddoc.put("_id", "_design/name");
+    view_ddoc.put("views", views);
+
+    db.save(view_ddoc);
+    }
+</pre>
+
+<h2 id="Using Views">Using Views</h2>
+
+For understanding the concept of views please refer to the
```
tomblench: reword "For understanding the concept of views" to something like "For background information on..." or similar

overview.html
tomblench: Actually I see you link to the project at the bottom but it doesn't hurt to have it in both places.

ricellis: Added a link in 011f84f.

ricellis: Added a "Compatibility" section in 9a7715e.

overview.html
ricellis: Good point, it appears to be called "Cloudant Local" colloquially on the official site, so I have used the full name in the first occurrence in the "Compatibility" list in 9a7715e and left as Cloudant Local elsewhere.

overview.html
ricellis: Updated all external links in 6392530 to use `target="_blank"`

overview.html
ricellis: Fixed in 6392530

overview.html
ricellis: Done in 6392530

overview.html
ricellis: This was copy/paste from the previous API reference. Not all of them are methods, `Cloudant Query`, for example. I agree that perhaps function is a misleading word in this context. What about some mix of capabilities and operations e.g.

* Capabilities
    * Server Operations
    * Database Operations
    * Document Operations
        * Single document CRUD
        * Bulk document CRU
    * Attachments
        * Standalone Attachments
        * Inline attachment
    * Design Documents
    * Using Views
    * Cloudant Query
    * Cloudant Search

emlaver:
+1

overview.html
tomblench: Yes, capabilities sounds better

tomblench:
@ricellis maybe revisit the TimeoutOption class as discussed, see if there is any appropriate class in the java api?

ricellis:
e543d58  - removed `TimeoutOption` class for e.g. `(5, TimeUnit.MINUTES)` style calls as used elsewhere in `java.util.concurrent`.
Also renamed `connectionTimeout` to `connectTimeout` to match `HttpUrlConnection`.

overview.html
ricellis: Fixed in 9f92fa2

tomblench:
+1

PR #143, opened 2015-11-11
142 alldocs parms

### LINKED ISSUE #142: Request parameters are not used on _all_docs request
The new `AllDocsRequest` is not correctly applying parameters from the builder to the request.
The issue is caused by the annotated fields not being visible from the `ViewQueryParameters` superclass.

ricellis:
*What*
Ensure query parameters are honoured for `_all_docs` requests.
Fixes #142.

*How*
The `AllDocsRequestBuilder` uses a subclass of `ViewQueryParameters` so the `getDeclaredFields()` call in `processParameters()` was not able to see the inherited fields and hence the parameters were not appended to the request.
By using `getFields()` there is visibility of the superclass fields (which have public visibility anyway so there is no need to use `getDeclaredFields()`).

*Testing*
Added a new test for an `_all_docs` request using keys to ensure that parameters are used.

reviewer @rhyshort 
reviewer @alfinkel


rhyshort:
+1

alfinkel:
+1

Issue #136, opened 2015-11-11
Add HTTP response codes to Repsonse objects


ricellis:
The `Response` class does not include the HTTP response code. We should add this information to help people align client behaviour to the more general Cloudant documentation, especially useful for cases like `201` and `202`.

We should validate that the status code has been added to `CouchDbException` and its subclasses where possible as well.

PR #148, opened 2015-11-19
Add HTTP response codes to Response objects

emlaver:
*What*
The Response class does not include the HTTP response code. We should add this information to help people align client behavior to the more general Cloudant documentation.
#136 

*How*
- Add setters for http connection's response code and message to Response class
- Set code and message in executeToResponse and in response list of bulk method
- Remove unnecessary class parameter in getResponseList

*Testing*
Add ResponseTest class for asserting response code in returned Response object and in status code for CouchDbException and subclasses.

reviewer @ricellis

### DIFF
--- a/src/main/java/com/cloudant/client/api/model/Response.java
+++ b/src/main/java/com/cloudant/client/api/model/Response.java
@@ -64,6 +64,12 @@ public class Response {
         return response.getReason();
     }
 
+    /**
+     * @return the HTTP status code returned by the server or 0 if no code was available
+     */
+    public int getStatusCode() {
+        return response.getStatusCode();
+    }
 
     /**
      * @return <tt>id</tt> and <tt>rev</tt> concatenated.
--- a/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
+++ b/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
@@ -24,10 +24,13 @@ import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getRespons
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.streamToString;
 import static com.cloudant.client.org.lightcouch.internal.URIBuilder.buildUri;
 
+import com.cloudant.http.Http;
+import com.cloudant.http.HttpConnection;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.List;
@@ -267,18 +270,32 @@ public abstract class CouchDatabaseBase {
      */
     public List<Response> bulk(List<?> objects, boolean allOrNothing) {
         assertNotEmpty(objects, "objects");
-        InputStream response = null;
+        InputStream responseStream = null;
+        HttpConnection connection;
         try {
             final String allOrNothingVal = allOrNothing ? "\"all_or_nothing\": true, " : "";
             final URI uri = buildUri(getDBUri()).path("_bulk_docs").build();
             final String json = String.format("{%s%s%s}", allOrNothingVal, "\"docs\": ", getGson
                     ().toJson(objects));
-            response = couchDbClient.post(uri, json);
-            return getResponseList(response, getGson(), Response.class,
+            connection = Http.POST(uri, "application/json");
+            if (json != null && !json.isEmpty()) {
+                connection.setRequestBody(json);
+            }
+            couchDbClient.execute(connection);
+            responseStream = connection.responseAsInputStream();
+            List<Response> bulkResponses = getResponseList(responseStream, getGson(),
                     new TypeToken<List<Response>>() {
                     }.getType());
+            for(Response response : bulkResponses) {
+                response.setStatusCode(connection.getConnection().getResponseCode());
+                response.setReason(connection.getConnection().getResponseMessage());
+            }
+            return bulkResponses;
+        }
+        catch (IOException e) {
+            throw new CouchDbException("Error retrieving response input stream.", e);
         } finally {
-            close(response);
+            close(responseStream);
         }
     }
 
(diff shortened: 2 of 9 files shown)


src/main/java/com/cloudant/client/api/model/Response.java
```diff
@@ -64,6 +64,12 @@ public String getReason() {
         return response.getReason();
     }
 
+    /**
+     * @return reason code returned by the server or 0
```
ricellis: the comment says

>reason code

but the method is called `getStatusCode()`, I think the comment should say:

> the HTTP status code returned by the server or 0 if no code was available

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -267,18 +269,28 @@ public Response remove(String id, String rev) {
...
-            response = couchDbClient.post(uri, json);
-            return getResponseList(response, getGson(), Response.class,
+            connection = couchDbClient.post(uri, json, HttpConnection.class);
+            responseStream = connection.responseAsInputStream();
+            List<Response> bulkResponses = getResponseList(responseStream, getGson(),
                     new TypeToken<List<Response>>() {
                     }.getType());
+            for(Response response : bulkResponses) {
+                response.setStatusCode(connection.getConnection().getResponseCode());
+                response.setReason(connection.getConnection().getResponseMessage());
```
ricellis: I think this is the best that can be done for now with the current `List<Response>`, but at some point in the future it will probably make more sense to have a different return type that could hold the HTTP status code, and had a getter to unpack the list of individual responses.

src/main/java/com/cloudant/client/org/lightcouch/Response.java
```diff
@@ -54,6 +56,10 @@ public String getReason() {
         return reason;
     }
 
+    public void setReason(String reason) {
```
ricellis: Does this setter need to be `public`?

src/main/java/com/cloudant/client/org/lightcouch/Response.java
```diff
@@ -61,4 +67,12 @@ public String getReason() {
     public String toString() {
         return "Response [id=" + id + ", rev=" + rev + "]";
     }
+
+    public int getStatusCode() {
+        return code;
+    }
+
+    public void setStatusCode(int code) {
```
ricellis: Also does this one need to be `public`?

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -267,18 +269,28 @@ public Response remove(String id, String rev) {
...
+        InputStream responseStream = null;
+        HttpConnection connection;
         try {
             final String allOrNothingVal = allOrNothing ? "\"all_or_nothing\": true, " : "";
             final URI uri = buildUri(getDBUri()).path("_bulk_docs").build();
             final String json = String.format("{%s%s%s}", allOrNothingVal, "\"docs\": ", getGson
                     ().toJson(objects));
-            response = couchDbClient.post(uri, json);
-            return getResponseList(response, getGson(), Response.class,
+            connection = couchDbClient.post(uri, json, HttpConnection.class);
```
ricellis: Whilst I understand the sentiment of trying to have common code for this given that we only have one use of the new `CouchDbClient.post(...)` method you added I think it might be better if that didn't take a type and just returned the connection. Alternatively it might be easier just to make the `HttpConnection` and call execute:
```java
connection = Http.POST(uri, "application/json");
//set content
couchDbClient.execute(connection);
```

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -422,6 +427,20 @@ public InputStream post(URI uri, String json) {
         return executeToInputStream(connection);
     }
 
+    <T> T post(URI uri, String json, Class<T> type) {
```
ricellis: See https://github.com/cloudant/java-cloudant/pull/148/files#r45371283

src/test/java/com/cloudant/tests/ResponseTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Test cases to verify status code from Response object.
+ * Assert status codes in CouchDbException and its subclasses.
+ */
+public class ResponseTest {
+    @ClassRule
+    public static final TestLog TEST_LOG = new TestLog();
```
ricellis: Doesn't look like there is any logging in this test so probably don't need this.

src/test/java/com/cloudant/tests/ResponseTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @ClassRule
+    public static CloudantClientResource clientResource = new CloudantClientResource();
+    @Rule
```
ricellis: Creating a new database for each test method can make the tests quite a lot slower.
Can we just use the `ChainRule` for a single database for the duration of the testing and just using different document IDs?

```java
private Foo foo;
@Before
public void newFoo(){
foo = new Foo(Utils.generateUUID())
}
```

src/main/java/com/cloudant/client/api/model/Response.java
emlaver: Fixed in 4049d30

src/main/java/com/cloudant/client/org/lightcouch/Response.java
emlaver: Fixed in 4049d30

src/main/java/com/cloudant/client/org/lightcouch/Response.java
emlaver: Fixed in 4049d30

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
emlaver: Fixed in 4049d30

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
emlaver: Fixed in 4049d30

src/test/java/com/cloudant/tests/ResponseTest.java
emlaver: Fixed in 4049d30

src/test/java/com/cloudant/tests/ResponseTest.java
emlaver: Fixed in 4049d30

ricellis:
+1

emlaver:
@ricellis New entry in CHANGES.md cdf0a4f

alfinkel:
+1

PR #150, opened 2015-11-24
145 stateless pagination

### LINKED ISSUE #145: Pagination that enables stateless sessions
I am trying to port out java code to the JAVA client 2.0.0 release.

I am trying to find something equivalent to the old style view.queryPage()

This let me get a page and send a param back to the UI for getting the next and previous page. Then when the UI requires the next page it could be retrieved directly using the params originally sent to the UI. This made the server side session stateless.

In 2.0.0 it seems that I need to keep the ViewResponse<K,V> in the session until the UI requests for the next page....

This makes all our sessions state-full, which we want to avoid......

ricellis:
*What*
Enable 1.x style string parameter based paging for 2.x.

*Why*
Fixes #145. There is a use case for passing a String parameter out of the Java application and avoiding keeping any state. The 2.x style paging relied on keeping `ViewResponse` state.

*How*
Added new methods:
* `ViewResponse.getNextPageToken()`
* `ViewResponse.getPreviousPageToken()`
* `ViewRequest.getResponse(String paginationToken)`

Under the covers this requires serialization to JSON of the `PageMetadata` and some `ViewQueryParameters` (`startkey ` `startkey_docid` `endkey ` `endkey_docid` `descending` ` inclusive_end`) followed by Base64 encoding into the token. The new `PaginationToken` class is used for this.

Methods for manipulating the query parameters for paging were moved from `ViewRequestParameters` to `PageMetadata` ensuring isolation of the pagination functions from the query parameters.

*Testing*
Added new `stateless` parameter to `ViewPaginationTests`.
Added new test for `inclusiveEnd=false` to ensure that works with all kinds of paging.

reviewer @brynh 
reviewer @mikerhodes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Unreleased
 - [IMPROVED] Included error and reason information in message from `CouchDbException` classes.
 - [IMPROVED] Added HTTP status code to `Response` objects.
+- [IMPROVED] Added parameter pagination option for views. See `ViewRequest.getResponse(String)`.
 - [FIX] Too many bytes written exception caused by inconsistent encoding between UTF-8 and the
   JVM default. UTF-8 is now correctly used for the request body content length and throughout.
 - [FIX] Fixed deserialization of `ReplicatorDocument` where the source or target url is a JSON
--- a/src/main/java/com/cloudant/client/api/views/ViewRequest.java
+++ b/src/main/java/com/cloudant/client/api/views/ViewRequest.java
@@ -46,6 +46,40 @@ public interface ViewRequest<K, V> {
     ViewResponse<K, V> getResponse() throws IOException;
 
     /**
+     * Performs view request for the page represented by a pagination token obtained from a
+     * previous request.
+     * <P>
+     * Example usage for stateless pagination:
+     * </P>
+     * <pre>
+     * {@code
+     * ViewResponse<String, String> responsePage1 = db.getViewRequestBuilder("designDoc","viewName")
+     *                   .newRequest(Key.Type.STRING, String.class)
+     *                   .build()
+     *                   .getResponse();
+     * String tokenForPage2 = responsePage1.getNextPageToken();
+     *
+     * // To request the next page:
+     * // 1. Recreate the view request by supplying the same parameters to the builder.
+     * // 2. Execute the request using the pagination token retrieved above.
+     * ViewResponse<String, String> responsePage2 = db.getViewRequestBuilder("designDoc","viewName")
+     *                   .newRequest(Key.Type.STRING, String.class)
+     *                   .build()
+     *                   .getResponse(tokenForPage2);
+     * }
+     * </pre>
+     *
+     * @param paginationToken obtained from a previous ViewResponse, {@code null} is equivalent to
+     *                        {@link #getResponse()}
+     * @return the response object
+     * @throws IOException if there is an error communicating with the server
+     * @see ViewResponse#getNextPageToken()
+     * @see ViewResponse#getPreviousPageToken()
+     * @since 2.1.0
+     */
+    ViewResponse<K, V> getResponse(String paginationToken) throws IOException;
+
+    /**
      * Performs the request and returns a single value.
      * <P>
      * This is primarily intended for retrieving a single result (e.g. count) from a reduced view.
--- a/src/main/java/com/cloudant/client/api/views/ViewResponse.java
+++ b/src/main/java/com/cloudant/client/api/views/ViewResponse.java
@@ -140,6 +140,29 @@ public interface ViewResponse<K, V> extends Iterable<ViewResponse<K, V>> {
     ViewResponse<K, V> previousPage() throws IOException;
 
     /**
+     * Returns an opaque pagination token for the next page of results, which can be used with
+     * {@link ViewRequest#getResponse(String)} to retrieve the next page without keeping
+     * ViewResponse state.
+     *
+     * @return opaque pagination token for the next page, or {@code null} if there is no next page
+     * @see ViewRequest#getResponse(String)
+     * @since 2.1.0
+     */
+    String getNextPageToken();
+
+    /**
+     * Returns an opaque pagination token for the previous page of results, which can be used with
+     * {@link ViewRequest#getResponse(String)} to retrieve the previous page without keeping
+     * ViewResponse state.
+     *
+     * @return opaque pagination token for the previous page, or {@code null} if there is no
+     * previous page
+     * @see ViewRequest#getResponse(String)
+     * @since 2.1.0
+     */
+    String getPreviousPageToken();
+
+    /**
      * <P>
      * Get the page number of this response.
      * </P>
--- a/src/main/java/com/cloudant/client/internal/views/PageMetadata.java
+++ b/src/main/java/com/cloudant/client/internal/views/PageMetadata.java
@@ -14,15 +14,17 @@
 
 package com.cloudant.client.internal.views;
 
+import com.google.gson.annotations.SerializedName;
+
 /**
  * <P>
- * Object for serializing metadata about a page.
+ * This class encapsulates metadata about a page: its number, direction and the query parameters
+ * needed to request it. A {@link ViewResponseImpl} generates instances of this class to pass that
+ * data to a request when nextPage() or previousPage() is called.
  * </P>
  * <P>
- * This class is used to persist the parameter values for a page. A {@link ViewResponseImpl}
- * calculates the parameters required to get the pages preceding and following it and generates
- * instances of this class to pass that data to a request when nextPage() or previousPage() is
- * called.
+ * Provides static methods for generating page query parameters for forward or backward pages based
+ * on specified start keys and paging directions.
  * </P>
  */
 final class PageMetadata<K, V> {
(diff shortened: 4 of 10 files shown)


CHANGES.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
-- [IMPROVED] Add HTTP status code to `Response` objects.
+- [IMPROVED] Added parameter pagination option for views. See `ViewRequest.getResponse(String)`.
+- [IMPROVED] Added HTTP status code to `Response` objects.
```
mikerhodes: This last line should be a separate commit, doesn't belong with the view pagination stuff.

mikerhodes:
Commit message for 75f99a4aaf03e1ef8fad961a72973fbd3e59d16f could say what documentation updates you made briefly.

src/main/java/com/cloudant/client/api/views/ViewRequest.java
```diff
@@ -46,6 +46,37 @@
     ViewResponse<K, V> getResponse() throws IOException;
 
     /**
+     * Performs the request for the given parameter ID and returns the response.
```
mikerhodes: I'm not sure parameter is the best name here, as both parameter and ID terms have a bunch of meanings here (e.g., I thought you meant query string parameter for a bit).

I'd go for something pagination specific, like `paginationToken`.

src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -140,6 +140,28 @@
     ViewResponse<K, V> previousPage() throws IOException;
 
     /**
+     * Returns an encoded string representing the query for the next page of results, which can
+     * be used with {@link ViewRequest#getResponse(String)} to retrieve the next page without
+     * keeping ViewResponse state.
+     *
+     * @return encoded string parameter for the next page, or null if there is no next page
```
mikerhodes: If you chose to use `pagination token` this can be made simpler, "opaque pagination token for next page, or null...".

src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -140,6 +140,28 @@
     ViewResponse<K, V> previousPage() throws IOException;
 
     /**
+     * Returns an encoded string representing the query for the next page of results, which can
```
mikerhodes: I'd tend to just go for "opaque" over encoded string.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
```diff
@@ -30,23 +43,79 @@
...
+            Gson g = registerKAdapter(pageRequestParameters, new GsonBuilder())
+                    .excludeFieldsWithoutExposeAnnotation()
+                    .create();
+            return new String(Base64.encodeBase64URLSafe(g.toJson(this).getBytes("UTF-8")),
+                    Charset.forName("UTF-8"));
+        } catch (UnsupportedEncodingException e) {
+            //all JVMs should support UTF-8
+            throw new RuntimeException(e);
+        }
+    }
```
mikerhodes: `toString` should be returning a human-readable string for printing. I'd suggest another name be used.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
```diff
@@ -30,23 +43,79 @@
...
         FORWARD,
+        @SerializedName("B")
         BACKWARD
     }
 
     PageMetadata(PagingDirection direction) {
         this.direction = direction;
     }
+
+    static <K, V> PageMetadata<K, V> fromString(final ViewQueryParameters<K, V> initialParameters,
```
mikerhodes: I don't think this name really conveys what the method is doing, particularly given you're merging with the initialParameters in addition to using the string.

It's somewhat complicated that the pagination parameter doesn't include some of the ViewQueryParamters, which makes calling it "parameter" even more complicated -- I now have many types of parameter in my head.

CHANGES.md
ricellis: I've made the `CHANGES.md` correction in master and reverted it here in 3f529fb. Everything should be fine after rebase/merge.

src/main/java/com/cloudant/client/api/views/ViewRequest.java
ricellis: Fixed in b031a00.

src/main/java/com/cloudant/client/api/views/ViewResponse.java
ricellis: Fixed in b031a00.

src/main/java/com/cloudant/client/api/views/ViewResponse.java
ricellis: Fixed in b031a00.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
ricellis: Fixed in 63c295b `pageMetadataToTokenString`.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
ricellis: Following this comment and our offline discussion about it I have isolated the paging information from the query parameters and created a `PaginationToken` class for serializing the necessary information for the opaque string pagination token. All in 63c295b, other than a quick method rename in 425bf37.

brynh:
Looks good to me: +1

src/main/java/com/cloudant/client/api/views/ViewRequest.java
```diff
@@ -46,6 +46,38 @@
     ViewResponse<K, V> getResponse() throws IOException;
 
     /**
+     * Performs the request for the given pagination token and returns the response.
```
mikerhodes: "Performs view request for the page represented by a pagination token obtained from a previous request"?

src/main/java/com/cloudant/client/api/views/ViewRequest.java
```diff
@@ -46,6 +46,38 @@
...
+     * <pre>
+     * {@code
+     * ViewResponse<String, String> responsePage1 = db.getViewRequestBuilder("designDoc","viewName")
+     *                   .newRequest(Key.Type.STRING, String.class)
+     *                   .build()
+     *                   .getResponse();
+     * String tokenForPage2 = responsePage1.getNextPageToken();
+     *
+     * // New session recreates the ViewRequest (by resupplying the original parameters to the
+     * // builder) and then uses the token to retrieve the desired page.
```
mikerhodes: I'd phrase this more as instructions:

```
To request the next page:
1. Recreate the view request by supplying the same parameters to the builder.
2. Execute the request using the pagination token retrieved above.
```

src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -140,6 +140,29 @@
     ViewResponse<K, V> previousPage() throws IOException;
 
     /**
+     * Returns an opaque pagination token representing the query for the next page of results, which
```
mikerhodes: It doesn't represent the query, only the parts of the query required for pagination. This description doesn't gel with the fact that one must recreate the query before supplying the token in getResponse().

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
```diff
@@ -14,15 +14,17 @@
 
 package com.cloudant.client.internal.views;
 
+import com.google.gson.annotations.SerializedName;
+
 /**
  * <P>
- * Object for serializing metadata about a page.
+ * This class encapsulates metadata about a page, its number, direction and the query parameters
```
mikerhodes: First comma should be a colon?

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
```diff
@@ -30,23 +32,129 @@
...
+        //stay the same for forward
+        pageParameters.inclusive_end = initialQueryParameters.inclusive_end;
+        pageParameters.descending = initialQueryParameters.descending;
+        if (initialQueryParameters.endkey != null) {
+            pageParameters.setEndKey(initialQueryParameters.endkey);
+        }
+        pageParameters.setEndKeyDocId(initialQueryParameters.endkey_docid);
+
+        //all other parameters stay the same
+        copyNonPagingParameters(initialQueryParameters, pageParameters);
```
mikerhodes: Why can't you just set all the query parameters from the initial query parameters object, then override them with the page-specific ones?

src/main/java/com/cloudant/client/internal/views/PaginationToken.java
```diff
@@ -0,0 +1,133 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.reflect.TypeToken;
+
+import org.apache.commons.codec.binary.Base64;
+
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+
+/**
+ * This class encapsulates the fields necessary to generate an opaque pagination token.
```
mikerhodes: You might want to note that it's purpose is purely for serialisation, given it doesn't look like it's used when there's no serialisation involved?

src/main/java/com/cloudant/client/internal/views/PaginationToken.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Generate a PageMetadata object for the page represented by the specified pagination token.
+     *
+     * @param paginationToken   opaque pagination token
+     * @param initialParameters the initial view query parameters (i.e. for the page 1 request).
+     * @param <K>               the view key type
+     * @param <V>               the view value type
+     * @return PageMetadata object for the given page
+     */
+    static <K, V> PageMetadata<K, V> getPageMetadataFromToken(String paginationToken, final
```
mikerhodes: `mergeTokenWithQueryParameters`? `getPageMetadataFromTokenAndQueryParameters`?

src/main/java/com/cloudant/client/internal/views/ViewQueryParameters.java
```diff
@@ -97,7 +97,7 @@
      *
      * @param parameters to copy information from
      */
-    public ViewQueryParameters(ViewQueryParameters<K, V> parameters) {
+    ViewQueryParameters(ViewQueryParameters<K, V> parameters) {
```
mikerhodes: There's a bunch of things you're making non-public in this stateless pagination commit. Is that changing of visibility a necessary step as part of that commit; it feels like it could be a separate commit.

src/main/java/com/cloudant/client/internal/views/ViewResponseImpl.java
```diff
@@ -219,6 +219,22 @@ public boolean hasPreviousPage() {
     }
 
     @Override
+    public String getNextPageToken() {
+        if (hasNext) {
+            return PaginationToken.pageMetadataToTokenString(nextPageMetadata);
```
mikerhodes: `pageMetadataToTokenString` could just be `tokenize`.

src/main/java/com/cloudant/client/internal/views/ViewQueryParameters.java
```diff
@@ -277,68 +277,11 @@ protected URIBuilder getViewURIBuilder() {
                 "/_view/" + viewName);
     }
 
-    public JsonElement asJson() {
+    JsonElement asJson() {
         Map<String, Object> parameters = processParameters(gson);
         return gson.toJsonTree(parameters);
     }
 
-    public ViewQueryParameters<K, V> forwardPaginationQueryParameters(K startkey, String
```
mikerhodes: Just to check my understanding, this code is in PaginationToken now? Or is it just not required?

src/main/java/com/cloudant/client/api/views/ViewRequest.java
ricellis: Fixed in 4df6027.

src/main/java/com/cloudant/client/api/views/ViewRequest.java
ricellis: Fixed in 4df6027.

src/main/java/com/cloudant/client/api/views/ViewResponse.java
ricellis: Removed reference to query in 16e0133.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
ricellis: Do you mean by making ViewQueryParameters `Cloneable`?
Otherwise isn't any method or code to set all the parameters as vulnerable to the addition of new parameters as the non-paging one (although I agree a method that copies everything is probably easier to maintain).

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
ricellis: Fixed in 09225de.

src/main/java/com/cloudant/client/internal/views/PaginationToken.java
ricellis: Fixed in a476c39.

src/main/java/com/cloudant/client/internal/views/PaginationToken.java
ricellis: Renamed to `mergeTokenAndQueryParameters` in d75344e.

src/main/java/com/cloudant/client/internal/views/ViewQueryParameters.java
ricellis: I'll break them out during the squash/rebase.

src/main/java/com/cloudant/client/internal/views/ViewResponseImpl.java
ricellis: Renamed in 096f4db.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
mikerhodes: Implementing `Cloneable` would make things more self-contained. Right now I need to know that if I add a field to `ViewQueryParameters` I also need to make a change in `PageMetadata`, which I suspect even you might forget in a few months :) So yes, I think that'd work well.

src/main/java/com/cloudant/client/api/views/ViewRequest.java
mikerhodes: Great.

src/main/java/com/cloudant/client/api/views/ViewRequest.java
mikerhodes: Great.

src/main/java/com/cloudant/client/api/views/ViewResponse.java
mikerhodes: Great.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
mikerhodes: Great,

src/main/java/com/cloudant/client/internal/views/PaginationToken.java
mikerhodes: Great.

src/main/java/com/cloudant/client/internal/views/PaginationToken.java
mikerhodes: Great.

src/main/java/com/cloudant/client/internal/views/ViewResponseImpl.java
mikerhodes: Great; reads much better to me.

src/main/java/com/cloudant/client/internal/views/ViewQueryParameters.java
mikerhodes: Great.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
ricellis: Done in e51947e.

src/main/java/com/cloudant/client/internal/views/PageMetadata.java
```diff
@@ -30,23 +32,121 @@
...
+        return reversedParameters;
+    }
+
+    /**
+     * Copies the parameters unrelated to paging from one ViewQueryParameters object to another.
+     *
+     * @param sourceParameters      the source of the parameters
+     * @param destinationParameters the destination for the parameters
+     */
+    static <K, V> void copyNonPagingParameters(ViewQueryParameters<K, V> sourceParameters,
```
mikerhodes: I think this is unused.

ricellis: Indeed it is, removed in a4a0236.

mikerhodes:
+1.

bennyhal:
Hi there

I just tried this out it is in line with what we need, thx for that

However there is a bug int he code

If I have this start key 
["TENANT_0", 1, 1, "activity"]

then on page 2 it translates it to this start key 
["TENANT_0", 1.0, 1.0, "activity", "Activity_012_3"]

notice the 1 has been converted to a 1.0

From the order point of view the key 
["TENANT_0", 1.0, 1.0, "activity", "Activity_012_3"]
is before the initial start key 

So page 2 is giving me the wrong results

Pls advise

thx

Benny

PR #154, opened 2015-11-25
149 exception msgs

### LINKED ISSUE #149: Exceptions created from CouchDB error streams need better messages
Exception stack traces like this:
>Exception in thread "main" com.cloudant.client.org.lightcouch.CouchDbException
at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
at ...
at com.cloudant.client.org.lightcouch.CouchDbClient.execute(CouchDbClient.java:497)

are not hugely helpful. More information is available from the and `ex.getError()` and `ex.getReason()` methods, but it would be good if this was in the CouchDbException message to help the user.

For example something like:
> Exception in thread "main" com.cloudant.client.org.lightcouch.CouchDbException: illegal_database_name Name: 'testAnimal'. Only lowercase characters (a-z), digits (0-9), and any of the characters _, $, (, ), +, -, and / are allowed. Must begin with a letter.

One way to achieve this would be using a Gson adapter to use the message constructor.

ricellis:
*What*
Improve the exception messages, fixes #149 (which contains a good example of why we should do this).

*How*
Override `getMessage()` in `CouchDbException` to include the error and reason strings if available.
Ensure those JSON fields are deserialized into the exceptions.

*Testing*
Add some additional asserts to the `ResponseTest`.
Manual inspection of stack trace to verify message output:
```
Exception in thread "main" com.cloudant.client.org.lightcouch.CouchDbException: 400 Bad Request: illegal_database_name: Name: 'testAnimala'. Only lowercase characters (a-z), digits (0-9), and any of the characters _, $, (, ), +, -, and / are allowed. Must begin with a letter.
	at com.cloudant.client.org.lightcouch.CouchDbClient.execute(CouchDbClient.java:491)
	at com.cloudant.client.org.lightcouch.CouchDbClient.executeToInputStream(CouchDbClient.java:534)
	at com.cloudant.client.org.lightcouch.CouchDbClient.get(CouchDbClient.java:317)
	at com.cloudant.client.org.lightcouch.CouchDatabaseBase.find(CouchDatabaseBase.java:150)
	at com.cloudant.client.api.Database.find(Database.java:600)
```

reviewer @alfinkel 
reviewer @tomblench


src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -486,20 +487,32 @@ public HttpConnection execute(HttpConnection connection) {
...
+                        //Uses a new Gson so we don't accidentally recycle an exception.
+                        Gson g = new GsonBuilder().registerTypeAdapter(exceptionClass, new
+                                InstanceCreator<CouchDbException>() {
+                                    @Override
+                                    public CouchDbException createInstance(Type type) {
+                                        return ex;
+                                    }
+                                }).create();
+                        //now populate the exception with the error/reason other info from JSON
+                        g.fromJson(new InputStreamReader(es),
```
tomblench: Should we attempt to return part of the response if we can't deserialise it? There are instances where the error isn't returned as JSON such as HAProxy which can return HTML error messages. This might be one for another PR though.

src/main/java/com/cloudant/client/org/lightcouch/NoResourceException.java
```diff
@@ -1,30 +0,0 @@
-/**
```
tomblench: Why was this removed? Presumably this basically does the same job as `NoResourceException` and is therefore redundant?

alfinkel:
+1

src/main/java/com/cloudant/client/org/lightcouch/NoResourceException.java
ricellis: This was added in 851d7d9d82f46529e6013f407ca77559ecbfcd15 but it is not referenced anywhere. I think it may have been added to improve compatibility with sync-android, but since it is in a different package anyway I don't think it will serve that purpose.

src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
ricellis: Yes, we should. I think we can set anything that couldn't be parsed as JSON as the error on the exception. I'll get that added in.

ricellis: @tomblench added in c67d4d7

tomblench:
+1

PR #156, opened 2015-11-25
155 utf8 encoding

### LINKED ISSUE #155: IOException: too many bytes written when saving documents with special characters
Saving a document with a character like "ç" can fail in java-cloudant 2.0.0 if the default codepage is not `UTF-8`, resulting in an exception like:

>com.cloudant.client.org.lightcouch.CouchDbException: Error retrieving server response
>...
>Caused by: java.io.IOException: too many bytes written

Besides this instance caused by `HttpConnection` there are a few other places in the code base that rely on the default JVM encoding, which is not `UTF-8` on some platforms. All of these need to be fixed to use `UTF-8`.

ricellis:
*What*
Always specify `UTF-8` instead of using the default codepage. Fixes #155.

*How*
Find and fix all instances in the codebase that don't specify the codepage.

*Testing*
Added new test for ç, but it would only currently fail if executed on a non `UTF-8` codepage.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
+++ b/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
@@ -161,17 +161,22 @@ public class CookieInterceptor implements HttpConnectionRequestInterceptor,
     }
 
     private boolean sessionHasStarted(InputStream responseStream) {
-        //check the response body
-        Gson gson = new Gson();
-        JsonObject jsonResponse = gson.fromJson(new InputStreamReader(responseStream), JsonObject
-                .class);
+        try {
+            //check the response body
+            Gson gson = new Gson();
+            JsonObject jsonResponse = gson.fromJson(new InputStreamReader(responseStream,
+                    "UTF-8"), JsonObject.class);
 
-        // only check for ok:true, https://issues.apache.org/jira/browse/COUCHDB-1356
-        // means we cannot check that the name returned is the one we sent.
-        return jsonResponse != null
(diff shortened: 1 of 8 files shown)


src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -161,17 +161,21 @@ private String getCookie(URL url) {
...
+
+            // only check for ok:true, https://issues.apache.org/jira/browse/COUCHDB-1356
+            // means we cannot check that the name returned is the one we sent.
+            return jsonResponse != null
+                    && jsonResponse.has("ok")
+                    && jsonResponse.get("ok").isJsonPrimitive()
+                    && jsonResponse.getAsJsonPrimitive("ok").isBoolean()
+                    && jsonResponse.getAsJsonPrimitive("ok").getAsBoolean();
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(e);
```
tomblench: you should re-use your comment from elsewhere `//UTF-8 should be supported on all JVMs`

tomblench:
+1 - you might want to add the comment I suggested

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: Added in 971f239

rhyshort:
+1

PR #158, opened 2015-11-26
Enable findbugs

ricellis:
*What*
Enable findbugs static analysis as part of the gradle build. This is intended as a start to prevent any new "high" level issues getting added and we can over time reduce the filtering as we clean up issues.

*Why*
Prevent some issues sneaking through, such as the UTF-8 encoding problems identified in #155.

*How*
Added the findbugs plugin to gradle.
Switched on reporting for "high" severity issues only on the main source set at first. Ideally later we can enable other severity issues and include the test code.
Fixed a bad equals method identified by findbugs for `Params`.
Added excludes for a couple of naming issues, which it would be good to resolve, but would require bigger changes.

*Testing*
Runs in the gradle build `check` and therefore is included in Travis CI.

reviewer @rhyshort

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -260,8 +260,14 @@ apply plugin: 'findbugs'
 //findbugs
 findbugs {
     toolVersion = "3.0.1"
+    //report only high severity bugs for now
     reportLevel = "high"
+    //the code base is pretty small so use max effort
     effort = "max"
+    //we don't want to run findbugs on the test code yet
+    sourceSets = [sourceSets.main]
+    //exclude a couple of known bugs until we get the chance to fix them
+    excludeFilter = file("findbugs_excludes.xml")
 }
 
 tasks.withType(FindBugs) {
--- a/build.gradle
+++ b/build.gradle
@@ -271,8 +271,10 @@ findbugs {
 }
 
 tasks.withType(FindBugs) {
+    //currently only one report type can be used toggle which with a property
+    boolean generateXML = Boolean.getBoolean("findbugs.xml.report")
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -255,3 +255,24 @@ def deployer = uploadArchives.repositories.mavenDeployer
...
+    effort = "max"
+    //we don't want to run findbugs on the test code yet
+    sourceSets = [sourceSets.main]
+    //exclude a couple of known bugs until we get the chance to fix them
+    excludeFilter = file("findbugs_excludes.xml")
+}
+
+tasks.withType(FindBugs) {
+    reports {
+        xml.enabled = false
```
rhyshort: shouldn't we have this enabled for when java-cloudant is finally running on jenkins?

ricellis: Does Jenkins automatically read a findbugs generated xml report? If it does then yes, it would be good to switch it on. The HTML is much more human readable which is why I was using that one at the moment.

rhyshort: There a j[enkins plug-in](https://wiki.jenkins-ci.org/display/JENKINS/FindBugs+Plugin) to pick up the findbugs report, its not installed yet, but when we set up the build we can do that. So I think we should turn on the xml output

ricellis: Sadly only generation of one report is supported currently and html is definitely better for a local build - so I've added a toggle property in e2628f6. We can set it in the Jenkins build command `-Dfindbugs.xml.report=true`.

rhyshort:
+1

PR #161, opened 2015-11-30
157 cookie renew

### LINKED ISSUE #157: Cookie interceptor should get new cookie on 403
It looks like Cloudant returns a 403 when the cookie expires, which makes sense.

We should therefore also check for it here: https://github.com/cloudant/java-cloudant/blob/master/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java#L98

ricellis:
*What*
Renew the session cookie if a 403 is received for expiry. Fixes #157.

*Why*
Expired cookies from Cloudant generate a 403 (whereas CouchDB 1.6.1 generates a 401).

*How*
Added 403 status code checks to the `CookieInterceptor` and read the error stream to identify if it was an expiry case.
In non-expiry 403 cases a new `HttpConnectionInterceptorException` is thrown that contains the error information read from the error stream. This is necessary because otherwise the error information would be lost in 403 responses that were not due to cookie expiry. This is not ideal, but is the strategic fix that doesn't involve breaking the interceptor APIs.

*Testing*
Added new tests for a 403 in both expiry and non-expiry cases.

reviewer @mikerhodes 
reviewer @emlaver

### DIFF
--- a/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
+++ b/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
@@ -21,6 +21,9 @@ import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+
+import org.apache.commons.io.IOUtils;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -76,7 +79,7 @@ public class CookieInterceptor implements HttpConnectionRequestInterceptor,
 
     @Override
     public HttpConnectionInterceptorContext interceptRequest(HttpConnectionInterceptorContext
-                                                                         context) {
+                                                                     context) {
 
         HttpURLConnection connection = context.connection.getConnection();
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -92,18 +95,45 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
-                    context.replayRequest = false;
-                }
+            int statusCode = connection.getResponseCode();
+            switch (statusCode) {
+                case HttpURLConnection.HTTP_FORBIDDEN: //403
+                    //check if it was an expiry case
+                    InputStream errorStream = connection.getErrorStream();
+                    String errorString = new String(IOUtils.toString(errorStream, "UTF-8"));
+                    try {
+                        JsonObject errorResponse = new Gson().fromJson(errorString, JsonObject
```
mikerhodes: Why not just read from the stream using a bufferedreader?

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -92,18 +95,45 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
+                                .class);
+                        String error = errorResponse.getAsJsonPrimitive
+                                ("error").getAsString();
+                        String reason = errorResponse.getAsJsonPrimitive
+                                ("reason").getAsString();
+                        if (!"credentials_expired".equals(error)) {
+                            //wasn't a credentials expired, throw exception
+                            throw new HttpConnectionInterceptorException(error, reason);
+                        }
+                        //was expired - fall through to case 401 to get a new cookie
```
mikerhodes: Please don't fall through, this is too easy to bugger up later.

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: We can only read the error stream once and we need the content for either the exception or to know the cookie is expired, but we don't know which until we've read the stream. The error stream might not always be JSON so I thought it made sense just to use a string.

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: Changed for a boolean instead in d611ffb.

emlaver:
+1

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
mikerhodes: Okay.

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
mikerhodes: Looks good.

mikerhodes:
+1

PR #163, opened 2015-12-01
162 cookie ssl

### LINKED ISSUE #162: Custom SSL options not honored when getting `_session` cookie
The options `ClientBuilder.customSSLSocketFactory(SSLSocketFactory)` and `ClientBuilder.disableSSLAuthentication()` are not honored when the `CookieInterceptor` makes requests to the server `_session` endpoint.
This means it is not currently possible to use cookie authentication with, for example, a custom certificate chain.

ricellis:
*What*
Add other interceptors to the `CookieInterceptor`'s `_session` request. Fixes #162.

*Why*
Custom SSL configuration is applied using an interceptor, if this interceptor is not used by the `CookieInterceptor` then the communication can fail and the cookie is not obtained. 

*How*
Add all configured interceptors to the HTTP requests made by the `CookieInterceptor`. Although SSL is the only interceptor required at the moment, all the interceptors should be used to cover other use cases.

*Testing*
Added new `SslAuthenticationTest` cases that check that the SSL settings have been applied to the `_session` request.
To do these tests required a better HTTP mock so added okhttp mockwebserver for increased ability to assert on requests and flexibility of mocking HTTP responses with reduced amount of test code.
Refactored other tests where possible to use okhttp mockwebserver instead of `SimpleHttpServer`.

reviewer @rhyshort 
reviewer @alfinkel


CHANGES.md
```diff
@@ -3,6 +3,7 @@
 - [IMPROVED] Added HTTP status code to `Response` objects.
 - [FIX] Too many bytes written exception caused by inconsistent encoding between UTF-8 and the
   JVM default. UTF-8 is now correctly used for the request body content length and throughout.
+- [FIX] Made cookie authentication work when using custom SSL configurations.
```
rhyshort: I think this should be reworded, to something along the lines of 
-[FIX] `CookieInterceptor` now honours Custom SSL Configurations 

or similar

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -121,6 +121,15 @@ private String getCookie(URL url) {
 
             HttpConnection conn = Http.POST(sessionURL, "application/x-www-form-urlencoded");
             conn.setRequestBody(sessionRequestBody);
+
+            //when we request the session we need all interceptors except this one
+
+            conn.requestInterceptors.addAll(context.connection.requestInterceptors);
+            conn.requestInterceptors.remove(this);
+            conn.responseInterceptors.addAll(context.connection.responseInterceptors);
+            conn.responseInterceptors.remove(this);
```
rhyshort: Should we also guard against the possibility that a user may have manually added a cookie interceptor to the interceptors list in addition to the the automatically generated one?

ricellis: How would we know? We can introspect each interceptor, but they could be of any type. If they use the `CookieInterceptor` type then it will remove itself unless they override the behaviour. I don't think we can (currently) make any assumptions about the other interceptors and each has to take responsibility for how they execute any additional requests they do. I guess the alternative is generalizing this "remove self" behaviour in a utility (or making all interceptors extend an abstract type that provides it).

Did you have something in mind that we could do?

CHANGES.md
ricellis: Changed in fdba535, is that better?

rhyshort: Its great :)

src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
rhyshort: I didn't have anything in mind, just wondered if its something we should do, or just let a user shoot themselves in the foot

alfinkel:
Looks fine to me. +1

rhyshort:
+1

PR #164, opened 2015-12-01
153 rep doc url

### LINKED ISSUE #153: Deserialisation into `ReplicatorDocument` fails.
Deserialisation into `ReplicatorDocument` fails because the `target` field can be a JSON object not just a `String`. An example of this can be found when replicating from a remote DB set up via the Cloudant Dashboard.

ricellis:
*What*
Correct deserialization of `ReplicatorDocument` for object source/target URLs. Fixes #153.

*How*
Changed the source & target fields in `ReplicatorDocument` to`JsonElement` from String so they can support both String or `JsonObject`.
Modified the setters and getters to work with the `JsonElement`.

*Testing*
Resolves the failure of `com.cloudant.tests.ReplicatorTest#replicatorDB` when running against the Cloudant service.

reviewer @rhyshort 
reviewer @alfinkel

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 - [IMPROVED] Added HTTP status code to `Response` objects.
 - [FIX] Too many bytes written exception caused by inconsistent encoding between UTF-8 and the
   JVM default. UTF-8 is now correctly used for the request body content length and throughout.
+- [FIX] Fixed deserialization of `ReplicatorDocument` where the source or target url is a JSON
+  object not a string.
 
 # 2.0.0 (2015-11-12)
 - [NEW] `DesignDocument.MapReduce` now has a setter for the `dbcopy` field.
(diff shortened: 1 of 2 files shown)


rhyshort:
I wonder if we should also add the ability to pass back the headers associated with the source/target if they are included in the son object.

ricellis:
I didn't really want to change the API (which currently only does strings).
The extent of the documentation I can find says:
>Can be a database URL, or an object whose url property contains the full URL of the database.

So that does not exactly define well what other properties one might expect to find in the JSON either. Agreed we know that there can be headers in there, but if someone knows what they are doing enough to tinker with the headers for their replication URL they can probably manage the JSON directly without needing to use the `ReplicatorDocument` class.

Can you think of a common use case for interacting with them? Do any of the other client libraries expose them?

rhyshort:
A possible use case is to add headers to track the requests going through the system. 

I agree that the documentation around this leaves a lot to be desired, it would make our lives easier if we knew what could be in that object. Since the documentation is lacking, I think we should raise it with Cloudant Documentation, and come back to this when the documentation around it has been expanded, so +1

alfinkel:
+1

Issue #135, opened 2015-11-10
Port URI builder from sync-android


emlaver:
The java-cloudant client should use the same URI builder as sync-android. A number of fixes have gone into that URI builder for special cases (e.g. ":") and we should take advantage of them. Additionally it is another step towards making java-cloudant a suitable base for sync-android.

PR #167, opened 2015-12-11
Port URI helper over from sync-android

emlaver:
*What*
The java-cloudant client should use the same URI builder as sync-android. A number of fixes have gone into that URI builder for special cases (e.g. ":") and we should take advantage of them.
Issue #135 

*How*
- URI helper generic design: URIBase, URIBaseChainable, and URIBaseMethods
- Copy over CouchURIHelper and rename to DatabaseURIHelper from sync-android
- New URIBase class for handling Cloudant account URI calls
- Replace instances of buildUri with new database URI helper
- Add required constants from sync-android
- Add com.google.guava package for com.google.common.base.Joiner dependency in CouchURIHelper
- Access client URL from CouchDbProperties and remove protocol, port, and host properties
- Remove unused getProperty method

*Testing*
Add CouchURIHelperTest class from sync-android

reviewer @ricellis 
reviewer @tomblench


ricellis:
I'm concerned about the addition of a dependency on the (relatively) large guava library for the sake of a single `Joiner`. I think we should probably just use a `StringBuilder` and some small wrapping code to get the behaviour we want without the dependency.

ricellis:
There is a lot of (legacy) inconsistency in places around whether `/` is included at the start or end of paths. I think this could cause issues with encoding since it would be difficult to know whether a `/` should be encoded or not even if it is at at the start or end of a path.
I think the `path()` method should encode everything and handle the `/` characters required to concatenate path elements.

ricellis:
I mentioned in https://github.com/cloudant/java-cloudant/pull/167#commitcomment-14930289 that I was a bit concerned about the `build()` method resetting the state of the of the helper. I notice for example in the `Database` class that we keep re-using the same object. If we're going to follow that pattern we need to be really sure that everything is clean before we start building a new URL.

emlaver:
I've replaced the guava dependency with a method to join queries in 4c2bfb5.

ricellis:
What happens if someone calls `databaseUriHelper.query("myParam=myValue").query(someMap)`?
It looks to me like only the map entries would be used and the string would be ignored.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -241,7 +241,9 @@ private JsonObject getPermissionsObject() {
      */
     public Shard getShard(String docId) {
         assertNotEmpty(docId, "docId");
-        return client.couchDbClient.get(buildUri(getDBUri()).path("_shards/").path(docId).build(), Shard.class);
+        return client.couchDbClient.get(new DatabaseURIHelper(db.getDBUri()).path("_shards/")
```
ricellis: Remove trailing "/"?

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -393,7 +395,8 @@ public void createIndex(String indexDefinition) {
     public List<Index> listIndices() {
         InputStream response = null;
         try {
-            response = client.couchDbClient.get(buildUri(getDBUri()).path("_index/").build());
+            URI uri = new DatabaseURIHelper(db.getDBUri()).path("_index/").build();
```
ricellis: Remove trailing "/"?

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -411,8 +414,8 @@ public void createIndex(String indexDefinition) {
     public void deleteIndex(String indexName, String designDocId) {
         assertNotEmpty(indexName, "indexName");
         assertNotEmpty(designDocId, "designDocId");
-        URI uri = buildUri(getDBUri()).path("_index/").path(designDocId).path("/json/").path
-                (indexName).build();
+        URI uri = new DatabaseURIHelper(db.getDBUri()).path("_index/").path(designDocId)
```
ricellis: Remove trailing "/"?

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -411,8 +414,8 @@ public void createIndex(String indexDefinition) {
...
+                .path("/json/").path(indexName).build();
```
ricellis: Remove leading and trailing "/"?

src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -163,7 +163,7 @@
      * @see Database#setPermissions(String, EnumSet)
      */
     public ApiKey generateApiKey() {
-        URI uri = buildUri(getBaseUri()).path("_api/v2/api_keys").build();
+        URI uri = new URIBase(getBaseUri()).path("_api/v2/api_keys").build();
```
ricellis: `path("_api").path("v2").path("api_keys")`?

ricellis:
Have you run the Cloudant service tests?
I feel like some of these recent comments might have seen some `//` or `%2F` in URLs where they probably were not expected.

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
```diff
@@ -242,4 +246,17 @@ public static String ensureDesignPrefix(String id)
...
         return id.startsWith(CouchDbUtil.DESIGN_PREFIX) ? id : CouchDbUtil.DESIGN_PREFIX + id;
     }
+
+    /**
+     * Remove `_design` prefix if it exists.
+     * The _design prefix is handled in {@link ViewQueryParameters#getViewURIBuilder()}
+     * @param id The design document's id
+     * @return String in the form of `id`
+     */
+    public static String removeDesignPrefix(String id)
```
tomblench: This `CouchDbUtil` class seems to be a bit of a mixed bag of utility methods which don't belong elsewhere. Is it more appropriate for `removeDesignPrefix` `ensureDesignPrefix` etc to live in `DatabaseURIHelper` which deals with manipulating URL paths?

src/main/java/com/cloudant/client/org/lightcouch/Param.java
```diff
@@ -50,25 +51,19 @@ public String toURLEncodedString() {
     }
 
     private static String encodeQueryParameter(String in) {
+        // As described above, we need to use the Java URI class to encode
```
tomblench: Comment doesn't make sense to me. Did you copy this method from sync-android? Either way, this method is of little use, it's probably better to change the call sites to call `HierarchicalUriComponents.encodeUriComponent` directly.

src/test/java/com/cloudant/tests/AttachmentsTest.java
```diff
@@ -85,12 +90,17 @@ public void attachmentInline_getWithDocument() {
...
+    public void attachmentStandalone() throws IOException, URISyntaxException {
         byte[] bytesToDB = "binary data".getBytes();
         ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytesToDB);
         Response response = db.saveAttachment(bytesIn, "foo.txt", "text/plain");
 
-        InputStream in = db.find(response.getId() + "/foo.txt");
+        Document doc = db.find(Document.class, response.getId());
+        assertTrue(doc.getAttachments().containsKey("foo.txt"));
+
+        HttpConnection conn = Http.GET(new URI(db.getDBUri() + "/" + response.getId() + "/foo.txt"));
```
tomblench: Is this use of string concatenation intentional? I wonder why you didn't use `DatabaseURIHelper` here, or did you do this so that there is an alternative code path for building up the URL and this serves to check that the URL was encoded correctly by our production code?

src/main/java/com/cloudant/client/internal/DatabaseURIHelper.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Returns URI for {@code _changes} endpoint using passed
+     * {@code query}.
+     */
+    public URI changesUri(String queryKey, Object queryValue) {
+
+        if(queryKey.equals("since")){
+            //Object since = query.get("since");
```
tomblench: Commented code (if this came from sync-android, we should remove it there too)

src/main/java/com/cloudant/client/internal/URIBase.java
```diff
@@ -0,0 +1,19 @@
+package com.cloudant.client.internal;
+
+import java.net.URI;
+
+/**
+ * Contains methods for building a base URI to be used with a Cloudant account.
+ * This class is extended by DatabaseURIHelper to build URI's for Cloudant databases.
+ */
+public class URIBase extends URIBaseMethods<URIBase> {
```
tomblench: Although you have a comment here (which is a good thing!), Rich had to explain to me how these various URI classes fit together, especially the reasoning behind the type parameters and fluent style `returnThis`.

Can we please have a bit more javadoc commenting for each of the classes involved, explaining how they relate to each other.

src/main/java/com/cloudant/client/org/lightcouch/Param.java
emlaver: Yes, this method was copied from sync-android.

src/main/java/com/cloudant/client/internal/URIBase.java
emlaver: Added class comments in 537bf9a.

src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
emlaver: Moved design methods into database URI helper class in 44d32f6.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in 537bf9a.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in 537bf9a.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in 537bf9a.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in 537bf9a.

src/main/java/com/cloudant/client/api/CloudantClient.java
emlaver: Fixed in 537bf9a.

src/main/java/com/cloudant/client/org/lightcouch/Param.java
emlaver: Fixed in 4fab31a.

src/main/java/com/cloudant/client/internal/DatabaseURIHelper.java
emlaver: Remove in 37335da.

src/test/java/com/cloudant/tests/AttachmentsTest.java
emlaver: Fixed in 37335da.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1024,6 +1028,13 @@ public void ensureFullCommit() {
         db.ensureFullCommit();
     }
 
+    /**
+     * Database URI helper.
+     */
```
ricellis: I noticed this failed in my local build because there is no `@return` in the javadoc. I have updated Travis to make sure these errors get picked up too.
```
src/main/java/com/cloudant/client/api/Database.java:1034: warning: no @return
    public DatabaseURIHelper getDbUriHelper() {
```

src/main/java/com/cloudant/client/internal/URIBaseMethods.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Sets a path(s) for the URI.
+     * Note: File separator is handled in this method
+     * and should not be passed as parameter.
+     *
+     * @return The updated {@link T} object.
+     */
+    public T path(String path) {
+        this.path += "/" + encodePath(path);
```
ricellis: I think `baseUri.toAsciiString()` always has a trailing `/` in our usage so with the current `build()` method I think you will end up with (for example):
`http://some.example:80//path/path/path?query=true`
changing to `this.path += encodePath(path) + "/";` will result in:
`http://some.example:80/path/path/path/?query=true`

I think both of these are valid URLs, but I think we need to be sure there are no server side issues if we use either `//` or `/?`.
@tomblench any thoughts on this?

ricellis:
Can the old URIBuilder be deleted now?

src/main/java/com/cloudant/client/internal/URIBaseMethods.java
tomblench: I would honestly prefer to avoid either situation occurring.

How about
```
if (path.length() == 0) {
  this.path += encodePath(path);
} else {
  this.path += "/" + encodePath(path);
}
```
?

emlaver: @tomblench I agree with this logic and it will be in the next commit.

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in f5e8ab8.

src/main/java/com/cloudant/client/internal/URIBaseMethods.java
emlaver: Fixed in f58d16c.

emlaver:
@ricellis `URIBuilder` deleted in d63488c.

ricellis:
+1

tomblench:
+1

Issue #138, opened 2015-11-11
Refactor IndexTests


ricellis:
Currently the `IndexTests` perform a lot of operations within a single method which makes it awkward to debug. It would be better to break it down into multiple tests so we can easily identify which part is not working in the event of a test failure.

PR #168, opened 2015-12-15
Refactor IndexTests

emlaver:
*What*
Currently the IndexTests perform a lot of operations within a single method which makes it awkward to debug. It would be better to break it down into multiple tests so we can easily identify which part is not working in the event of a test failure.
Issue #138.

*How*
- Split indexTestAll case into separate test cases
- Create indexes and json map in setUp

*Testing*
Existing test cases split up.

reviewer @ricellis
reviewer @alfinkel

### DIFF
--- a/src/test/java/com/cloudant/tests/IndexTests.java
+++ b/src/test/java/com/cloudant/tests/IndexTests.java
@@ -38,6 +38,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+@Category(RequiresCloudant.class)
 public class IndexTests {
 
     private static CloudantClientResource clientResource = new CloudantClientResource();
@@ -47,6 +48,7 @@ public class IndexTests {
 
     private static Database db;
     private static CloudantClient account;
+    private static Map<String, Object> selector;
 
     @BeforeClass
     public static void setUp() throws Exception {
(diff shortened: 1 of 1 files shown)


ricellis:
I think we can have just have one `@Category(RequiresCloudant.class)` annotation at the class level instead of one on each test method.

ricellis:
+1

src/test/java/com/cloudant/tests/IndexTests.java
```diff
@@ -72,6 +70,16 @@ public void indexTestAll() {
         db.createIndex("Movie_year", "Movie_year", null,
                 new IndexField[]{new IndexField("Movie_year", SortOrder.asc)});
 
+        // selectorJson as a proper json object
```
alfinkel: What is selectorJson?

emlaver: This was copied over with the succeeding lines.  It's referencing the `selector` variable, but this may at one point have been called `selectorJson`.  I can either update this comment or remove it.

alfinkel: Your call on updating or removing (either one works for me) but since selectorJson no longer exists, the comment, in its current state, is confusing.

alfinkel:
+1 once you address the selectorJson comment.

PR #170, opened 2015-12-15
Fix testNonJsonErrorStream test

### LINKED ISSUE #169: testNonJsonErrorStream fails to logout session
The bad headers added to test the non-JSON error stream also get added to other requests in that test, which ultimately causes the `client.shutdown()` finally block to fail ever since #152 was resolved.

This can be resolved by adding a flag to only enable the headers for the specific portion of the test.

ricellis:
*What*
Fixes #169 for the Cloudant service test `testNonJsonErrorStream` failing.

*How*
Added a flag to only turn on the bad headers for part of the test, so that they don't impact the setup/teardown of the session.

*Testing*
The `com.cloudant.tests.ResponseTest.testNonJsonErrorStream` now passes again.

reviewer @rhyshort

### DIFF
--- a/src/test/java/com/cloudant/tests/ResponseTest.java
+++ b/src/test/java/com/cloudant/tests/ResponseTest.java
@@ -27,6 +27,7 @@ import org.junit.rules.RuleChain;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


rhyshort:
+1

Issue #101, opened 2015-10-02
Don't build JSON using string operations


s100:
Building JSON literals using string operations is dangerous. [Database.getIndexDefinition()](https://github.com/cloudant/java-cloudant/blob/a56aa8d7bfd9d8f41e0c2f90446ddb44f94a41f0/src/main/java/com/cloudant/client/api/Database.java#L824) is the first place I spotted this but there may be other examples in this codebase. GSON is a better choice.

ricellis:
Thanks for pointing this instance out. There are some inconsistencies in this library in this respect, but we do use GSON in most places. We will try and get this and any other remaining string uses fixed up.

s100:
Attempting to build valid JSON using string operations instead of a library is dangerous in principle but it is actually the cause of several defects. This line of code:

    database.findByIndex("\"selector\"asdfasdf", Subscription.class);

or this block of code:

    FindByIndexOptions findByIndexOptions = new FindByIndexOptions();
    findByIndexOptions.fields("\"");
    database.findByIndex("{\"type\":\"subscription\"}", Subscription.class, findByIndexOptions);

both bypass the input sanitisation in [Database.getFindByIndexBody()](https://github.com/cloudant/java-cloudant/blob/master/src/main/java/com/cloudant/client/api/Database.java#L1070) and cause it to construct invalid JSON, which is then passed to the Cloudant backend, resulting in:

    {"error":"bad_request","reason":"invalid UTF-8 JSON"}

I see at [line 1108](https://github.com/cloudant/java-cloudant/blob/master/src/main/java/com/cloudant/client/api/Database.java#L1108) you're actually using GSON to parse the selector JSON, but then just discarding the result. It would be better to just use GSON to assemble the object you need directly.

ricellis:
Sorry if I wasn't clear - I completely agree that this is a bad way of doing things! I've added it to a milestone to make sure we get it fixed soon.

ricellis:
@emlaver I noticed there are some instances of building arrays using strings in `Search.java` that should also be corrected.

PR #173, opened 2015-12-16
Use GSON objects to build JSON and not string operations

emlaver:
*What*
Building JSON using string operations is causing several defects. Use the GSON library to assemble the object instead.
Issue #101 

*How*
- Replace string building implementation with GSON objects in Database, Search, and CouchDatabaseBase class

*Testing*
New test cases to assert that invalid JSON will throw a bad request.

reviewer @ricellis
reviewer @rhyshort

### DIFF
--- a/src/main/java/com/cloudant/client/api/Database.java
+++ b/src/main/java/com/cloudant/client/api/Database.java
@@ -275,8 +275,8 @@ public class Database {
      */
     public void createIndex(String indexName, String designDocName, String indexType,
                             IndexField[] fields) {
-        String indexDefn = getIndexDefinition(indexName, designDocName, indexType, fields);
-        createIndex(indexDefn);
+        JsonObject indexDefn = getIndexDefinition(indexName, designDocName, indexType, fields);
+        createIndex(indexDefn.toString());
     }
 
     /**
@@ -357,10 +357,10 @@ public class Database {
         assertNotEmpty(options, "options");
 
         URI uri = new DatabaseURIHelper(db.getDBUri()).path("_find").build();
-        String body = getFindByIndexBody(selectorJson, options);
+        JsonObject body = getFindByIndexBody(selectorJson, options);
         InputStream stream = null;
         try {
-            stream = client.couchDbClient.executeToInputStream(createPost(uri, body,
+            stream = client.couchDbClient.executeToInputStream(createPost(uri, body.toString(),
                     "application/json"));
             Reader reader = new InputStreamReader(stream, "UTF-8");
             JsonArray jsonArray = new JsonParser().parse(reader)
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1029,135 +1029,67 @@ public void ensureFullCommit() {
...
-        if (isObject) {
-            finalbody.append("{\"selector\": ")
-                    .append(selectorJson);
-        } else {
-            //old support
-            finalbody.append("{" + selectorJson);
-        }
+        JsonObject indexObject = new JsonObject();
+        //Add selector json object
+        indexObject.addProperty("selector", selectorJson);
```
ricellis: Does this work? I though the supplied selector JSON was of the form `"selector" : {}`

src/main/java/com/cloudant/client/api/Search.java
```diff
@@ -369,7 +369,12 @@ public Search counts(String[] countsfields) {
     public Search drillDown(String fieldName, String fieldValue) {
         assertNotEmpty(fieldName, "fieldName");
         assertNotEmpty(fieldValue, "fieldValue");
-        uriBuilder.query("drilldown", "[\"" + fieldName + "\",\"" + fieldValue + "\"]");
+        JsonArray drillDownArray = new JsonArray();
+        JsonPrimitive fieldNamePrimitive = new JsonPrimitive(fieldName);
+        drillDownArray.add(fieldNamePrimitive);
+        JsonPrimitive fieldValuePrimitive = new JsonPrimitive(fieldValue);
+        drillDownArray.add(fieldValuePrimitive);
+        uriBuilder.query("drilldown", drillDownArray);
```
ricellis: `drilldown` can be specified multiple times, looks like the previous code was incorrect in this regard.
It should be `uriBuilder.query("drilldown", drillDownArray, false)` to not replace the parameter. I'm guessing this will be impacted by #167 though. Maybe it is worth moving this out into another issue as it is a bit unrelated.

src/main/java/com/cloudant/client/api/Database.java
emlaver: This is the toString value for `indexObject` after adding the selector object:
`{"selector":"{\"Person_name\":\"Alec Guinness\",\"Movie_year\":{\"$gt\":1960}}"}`

src/main/java/com/cloudant/client/api/Search.java
emlaver: Added as issue #174

src/main/java/com/cloudant/client/api/Database.java
ricellis: Isn't the index object going to end up with: `"selector" : "\"selector\" : {...}` because the string`selectorJson` at this point includes `"selector" : ` at the beginning.
If you run the cloudantServiceTest you'll see what I mean.

emlaver: I see your point but will add that the `"selector" :` at the beginning does not always exist in the `selectorJson` string.  This is a problem either way and will need to be handled.

emlaver:
@ricellis I've added logic to try and deserialize the JSON in `selectorJson`.   If it's valid, the selector object will be added to the final JSON object.  If it's not valid, check if the string starts with `"selector"`.  If it does start with selector, then create a substring and deserialize.  If not, still try to deserialize and throw an exception if it's invalid json syntax.

Also, after changing the original string implementation, I didn't add the index options field and sort into the final JSON object.  Those are also added in 66d039b.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1029,135 +1029,116 @@ public void ensureFullCommit() {
...
+            for(IndexField sort : options.getSort()) {
+                JsonObject sortObject = new JsonObject();
+                //JsonObject jsonSort = new JsonObject();
+                //jsonSort.addProperty(sort.getName(), sort.getOrder().toString());
+                //sortArray.add(jsonSort);
+                /*JsonPrimitive json = client.getGson().fromJson(sort.getName(),
+                        JsonPrimitive.class);
+                sortArray.add(json);
+                json = client.getGson().fromJson(sort.getOrder().toString(),
+                        JsonPrimitive.class);*/
```
ricellis: Remove commented out code

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1029,135 +1029,116 @@ public void ensureFullCommit() {
...
+                sortObject.addProperty(sort.getName(), sort.getOrder().toString());
+                sortArray.add(sortObject);
             }
-            so.append("]");
+            //sortObject.add("sort", sortArray);
```
ricellis: More commented out code

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1029,135 +1029,116 @@ public void ensureFullCommit() {
...
-            finalbody.append(",")
-                    .append("\"r\": ")
-                    .append(options.getReadQuorum());
+            indexObject.addProperty("r", options.getReadQuorum());
         }
         if (options.getUseIndex() != null) {
-            finalbody.append(",")
-                    .append("\"use_index\": ")
-                    .append(options.getUseIndex());
+            JsonArray useIndexArray = new JsonArray();
```
ricellis: This `useIndexArray` is never used.

src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1029,135 +1029,116 @@ public void ensureFullCommit() {
...
+            JsonPrimitive element = new JsonPrimitive(options.getUseIndex());
+            useIndexArray.add(element);
+            indexObject.add("use_index", getGson().fromJson(options.getUseIndex(), JsonElement.class));
```
ricellis: I think this is the only line needed, but it might be worth changing `JsonElement.class` to `JsonArray.class` since we should probably break if the `getUseIndex()` String doesn't turn into an array.

I think this also highlights a place in `FindByIndexOptions` that is still doing String JSON operations. That is creating the `useIndex. Probably worth changing the internal type for `useIndex` in `FindByIndexOptions` into a `JsonArray` and `return useIndex.toString();` from the `getUseIndex()` method. Although the `JsonArray` -> `String` -> `JsonArray` is slightly cumbersome it avoids exposing the internal JSON choice on the public `getUseIndex()` method. An alternative might be to use a reduced visibility method that returns the `JsonArray` directly to avoid the journey via String.

emlaver: If I change `JsonElement.class` to `JsonArray.class`, I get an error on test cases when the useIndex is not in array form.
I think it's a better idea to see how I can change the internal type for `useIndex` into a `JsonArray`.

emlaver: Looks like I would need to use a `public` access modifier to call an internal `getGsonUseIndex()` method in `Search`.   How about just changing the return type for `FindByIndexOptions.getUseIndex()` to `JsonArray`? I know it's very possible that it would break code, but we could just add a comment that `toString` can be used?

src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
```diff
@@ -142,7 +148,13 @@ public Integer getReadQuorum() {
     }
 
     public String getUseIndex() {
-        return useIndex;
+        return useIndex.toString();
     }
 
+    /**
+     * For use in {@link Database#getFindByIndexBody(String, FindByIndexOptions)}
+     */
```
rhyshort: Missing `@return` from java doc

src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
```diff
@@ -142,7 +148,13 @@ public Integer getReadQuorum() {
...
 
     public String getUseIndex() {
-        return useIndex;
+        return useIndex.toString();
     }
 
+    /**
+     * For use in {@link Database#getFindByIndexBody(String, FindByIndexOptions)}
+     */
+    public JsonArray getUseIndexAsJsonArray() {
```
ricellis: I'm reluctant to add a public method to this API class.
Maybe just make the field public?

ricellis:
Can you also add an entry to `CHANGES.md` for this?

src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
emlaver: I've removed this method in aba42c9.

src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
emlaver: After chatting over Slack, I decided on calling GSON's fromJson method to parse the JSON.  Fixed in aba42c9

ricellis:
+1

rhyshort:
+1

PR #176, opened 2016-01-04
172 complex int

### LINKED ISSUE #172: Pagination tokens turn integer parts of complex keys into floats
From: https://github.com/cloudant/java-cloudant/issues/145#issuecomment-165070891
 >If I have this start key
>`["TENANT_0", 1, 1, "activity"]`
>
>then on page 2 it translates it to this start key
>`["TENANT_0", 1.0, 1.0, "activity", "Activity_012_3"]`
>
>notice the 1 has been converted to a 1.0 
>
>From the order point of view the key
>`["TENANT_0", 1.0, 1.0, "activity", "Activity_012_3"]`
>is before the initial start key
>
>So page 2 is giving me the wrong results

ricellis:
*What*
Fixes #172 ensuring `int` in a complex keys is not turned into `float` by token pagination.

*How*
Internally use `JsonElement` for keys instead of the type erased parameter `K`. This prevents the `ComplexKey` being treated as an `Object` and preserves the JSON array representing the complex key.

*Testing*
Added a new test to check that `int` in a complex key start key remains an `int` for subsequent pages requested by token.
Enabled extra views for complex key pagination tests that appeared to have been accidentally disabled.

reviewer @alfinkel 
reviewer @emlaver

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Unreleased
 - [IMPROVED] Request a session delete on client shutdown.
 - [IMPROVED] Consistently encode all parts of request URLs and handle additional special characters.
+- [FIX] Stopped integers in complex key arrays turning into floats when using view pagination with tokens.
 
 # 2.1.0 (2015-12-04)
 - [IMPROVED] Included error and reason information in message from `CouchDbException` classes.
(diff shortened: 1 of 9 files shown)


alfinkel:
Should we be concerned about the Travis CI push failure?  Looks like a test is failing?

alfinkel:
All looks good to me.  If there is no real issue with the travis-ci/push failure then I am +1.

emlaver:
LGTM +1

ricellis:
Re: the travis push failure:
```
com.cloudant.tests.ReplicationTest > replication_filteredWithQueryParams FAILED
    com.cloudant.client.org.lightcouch.CouchDbException: 500 Internal Server Error: changes_reader_died
```
From what I can tell `changes_reader_died` can happen if the query parameters aren't valid for a filtered replication, though why they weren't valid for this one instance of the test I'm not sure. The same test passed in the other two builds associated with this PR and as far as I know that test is independent of the changes being made here so I'm going to press on as I think this is unrelated.

PR #177, opened 2016-01-04
Support POST requests in invokeUpdateHandler

### LINKED ISSUE #37: POST with invokeUpdateHandler?
Is there any way in this API to POST to an update handler? Passing in no docId just results in an error `docId is required` where I would expect it to `POST` instead of `PUT`.
```
java.lang.IllegalArgumentException: docId may not be null.
	at org.lightcouch.internal.CouchDbUtil.assertNotEmpty(Unknown Source)
	at org.lightcouch.CouchDatabaseBase.invokeUpdateHandler(Unknown Source)
	at com.cloudant.client.api.Database.invokeUpdateHandler(Unknown Source)
```

emlaver:
*What*
`Database.invokeUpdateHandler` only supports PUT requests.  Passing in an empty or null document Id just returns the error `docId is required`.
This fixes #37 by creating a POST request in the update handler when a null document id exists.

*How*
- If no document ID exists, use POST request in `CouchDatabaseBase.invokeUpdateHandler`
- Add new entry for update handler in example's design document. This is used in new test case.

*Testing*
New POST request test case using uuid.

reviewer @ricellis
reviewer @alfinkel

### DIFF
--- a/src/main/java/com/cloudant/client/api/Database.java
+++ b/src/main/java/com/cloudant/client/api/Database.java
@@ -985,6 +985,7 @@ public class Database {
      *
      * @param updateHandlerUri The Update Handler URI, in the format: <code>designDoc/update1</code>
      * @param docId            The document id to update.
+     *                         If no id is provided, then a document will be created.
      * @param params           The query parameters as {@link Params}.
      * @return The output of the request.
      * @see <a target="_blank"
@@ -994,7 +995,7 @@ public class Database {
     public String invokeUpdateHandler(String updateHandlerUri, String docId,
                                       Params params) {
         assertNotEmpty(params, "params");
-        return db.invokeUpdateHandler(updateHandlerUri, docId, params.getInternalParams());
+        return db.invokeUpdateHandler(updateHandlerUri, docId, params.getInternalParams(), null);
     }
 
     /**
--- a/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
+++ b/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
@@ -355,15 +355,26 @@ public abstract class CouchDatabaseBase {
      * @param updateHandlerUri The Update Handler URI, in the format: <code>designDoc/update1</code>
      * @param docId            The document id to update.
      * @param params           The query parameters as {@link Params}.
+     * @param postBodyRequest  The POST body to create a document if needed.
      * @return The output of the request.
      */
-    public String invokeUpdateHandler(String updateHandlerUri, String docId, Params params) {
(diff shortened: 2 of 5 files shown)


src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -985,6 +985,7 @@ public boolean contains(String id) {
      *
      * @param updateHandlerUri The Update Handler URI, in the format: <code>designDoc/update1</code>
      * @param docId            The document id to update.
+     *                         If null, a POST request is invoked.
```
ricellis: Saying "a POST request is invoked" relies on someone knowing that a POST will do a create. I think it is better to say that if no ID is provided a document will be be created.

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -353,15 +353,26 @@ public Response saveAttachment(InputStream in, String name, String contentType,
...
         assertNotEmpty(updateHandlerUri, "uri");
-        assertNotEmpty(docId, "docId");
         final String[] v = updateHandlerUri.split("/");
-        final URI uri = new DatabaseURIHelper(dbUri).path("_design").path(v[0]).path("_update")
-            .path(v[1]).path(docId).query(params).build();
-        final InputStream response = couchDbClient.put(uri);
+        final InputStream response;
+        final URI uri;
+        if(docId != null && !docId.isEmpty()) {
+            uri = new DatabaseURIHelper(dbUri).path("_design").path(v[0]).path("_update")
```
ricellis: A lot of this URI is common between both requests, maybe better to share between the branches? Maybe something like this, but with the comments
```java
DatabaseURIHelper uriHelper = new DatabaseURIHelper(dbUri).path("_design").path(v[0]).path("_update").path(v[1]).query(params);
if (docId != null && docId.isEmpty()) {
  uri = uriHelper.path(docId).build();
  response = couchDbClient.put(uri);
} else {
  uri = uriHelper.build();
  response = couchDbClient.post(uri, postBodyRequest);
}
```

src/main/java/com/cloudant/client/api/Database.java
emlaver: Fixed in 9a810ce.

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
emlaver: Fixed in 20d68f0.

ricellis:
+1 assuming the failures are unrelated.

alfinkel:
+1

Issue #174, opened 2015-12-17
Drilldown query parameter can be specified multiple times


emlaver:
`Search drillDown(String fieldName, String fieldValue)` method currently only supports one instance of the drilldown query.  The `query` call within the method needs to be changed to use a `false` boolean for not replacing existing query parameters.

PR #178, opened 2016-01-06
Drilldown query parameter can be specified multiple times

emlaver:
*What*
Support multiple drilldown queries by not replacing existing query.
Issue #174.

*How*
- Add `query(String name, Object value, boolean replace)` to be called by Search's `drilldown` method.

*Tests*
New test case for multiple drilldown queries.

reviewer @ricellis
reviewer @alfinkel

### DIFF
--- a/src/test/java/com/cloudant/tests/SearchTests.java
+++ b/src/test/java/com/cloudant/tests/SearchTests.java
@@ -182,7 +182,6 @@ public class SearchTests {
         assertEquals("herbivore", it.next());
     }
 
-
     @Test
     public void drillDownTest() {
         // do a faceted search for drilldown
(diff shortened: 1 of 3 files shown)


src/test/java/com/cloudant/tests/SearchTests.java
```diff
@@ -202,6 +201,25 @@ public void drillDownTest() {
...
+        SearchResult<Animal> rslt = srch.includeDocs(true)
+                .counts(new String[]{"class", "diet"})
+                .ranges("{ \"min_length\": {\"small\": \"[0 TO 1.0]\","
+                        + "\"medium\": \"[1.1 TO 3.0]\", \"large\": \"[3.1 TO 9999999]\"} }")
+                .drillDown("class", "bird")
+                .drillDown("diet", "omnivore")
+                .querySearchResult("class:bird", Animal.class);
+        assertNotNull(rslt);
+        assertNotNull(rslt.getRanges());
+        assertEquals(rslt.getRanges().entrySet().size(), 1);
```
ricellis: I think JUnit assertions are `assert*(expected, actual)` so I think these args (and for the following asserts) might need to be reversed to avoid confusing failure messages.

I think it is probably also worth having a comment on this test explaining how it works. I'm assuming that having two drilldowns limits the returned result to a single small animal?

emlaver: If the JUnit assertions args need to be reversed in this test case, then we may need a work item to reverse all the other test cases.

And yes, the two drilldowns limits the returned result for the views101 design document to one single animal.

emlaver: Fixed in 98ba5df.

ricellis: Thanks, I've got PR #179 for other incorrect `assertEquals` ordering.

ricellis:
+1

alfinkel:
+1

PR #179, opened 2016-01-07
Corrected assertEquals order in tests

ricellis:
*What*
Correct `assertEquals` argument order in tests.

*Why*
JUnit assertion failure messages use a string like "expected `<a>` but was `<b>`" so the failure messages are confusing if the objects are provided to `assertEquals` in the incorrect order.

*How*
Reversed some JUnit assertions to be in the correct expected then actual order.

*Testing*
No new tests required, fixed in SearchTests, DatabaseTest and HttpTest.

reviewer @emlaver


emlaver:
+1

PR #180, opened 2016-01-08
Added db name length configuration

ricellis:
*What*
Made db name length for tests configurable.

*Why*
Some databases use a different max length, so the normal max length of 128 cannot be used.

*How*
Used the system property `test.db.name.length` with a default of 128.

*Testing*
Ran Jenkins build using the new configuration option.

reviewer @rhyshort


rhyshort:
+1

PR #181, opened 2016-01-08
119 db interface

ricellis:
*What*
Extract a `Database` interface from the current `Database` class.

*Why*
Having an interface allows other implementations of the `Database`. The motivation is to make it easier to provide a delegating `Database` that can interact with a cache.

*How*
Extract the inner classes used for deserialization of some types from `Database` into their own classes.
Made the `Database` class an interface and moved the implementation
Removed un-needed references from `Search` to a non-interface `DatabaseImpl` method.

*Testing*
No new tests, exercised by existing tests using new `Database` interface.

reviewer @rhyshort
reviewer @emlaver


src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -833,12 +549,7 @@ public boolean contains(String id) {
      * @see <a target="_blank" href="https://docs.cloudant.com/document.html#delete">Documents -
      * delete</a>
      */
-    public com.cloudant.client.api.model.Response remove(Object object) {
-        Response couchDbResponse = db.remove(object);
-        com.cloudant.client.api.model.Response response = new com.cloudant.client.api.model
-                .Response(couchDbResponse);
-        return response;
-    }
+    com.cloudant.client.api.model.Response remove(Object object);
```
rhyshort: If possible, we should remove the namespacing and just import response directly.

ricellis: Yes, we should, done in 0f63e19.

rhyshort:
+1

emlaver:
LGTM +1

Issue #186, opened 2016-01-21
Defect: Cannot Save Attachment Without Revision ID


dlandis:
Using the latest version, it is not possible to save a new attachment with an id.

Here is the stack trace:

```
2016-01-20T16:47:34.92-0800 [App/0]      OUT java.lang.NullPointerException
2016-01-20T16:47:34.92-0800 [App/0]      OUT 	at com.cloudant.client.internal.URIBaseMethods.path(URIBaseMethods.java:37)
2016-01-20T16:47:34.92-0800 [App/0]      OUT 	at com.cloudant.client.internal.DatabaseURIHelper.revId(DatabaseURIHelper.java:179)
2016-01-20T16:47:34.92-0800 [App/0]      OUT 	at com.cloudant.client.internal.DatabaseURIHelper.attachmentUri(DatabaseURIHelper.java:160)
2016-01-20T16:47:34.92-0800 [App/0]      OUT 	at com.cloudant.client.org.lightcouch.CouchDatabaseBase.saveAttachment(CouchDatabaseBase.java:342)
2016-01-20T16:47:34.92-0800 [App/0]      OUT 	at com.cloudant.client.api.Database.saveAttachment(Database.java:960)
```

The `saveAttachment` method was called with a `null` value for the `docRev` argument, which is correct according to the JavaDoc:

    @param docRev The document revision to save the attachment under, or {@code null} when saving to a new document.

dlandis:
ping @emlaver 

Also noticed some other issues in the APIs for example:

     * @param docId       The document id to save the attachment under, or {@code null} to save
     *                    under a new document.

And then the first line of the method is:

    assertNotEmpty(docId, "docId");

dlandis:
see pull request #187

dlandis:
Note, a temporary workaround for this issue is to use an empty string instead of null for the `docRev` argument.

PR #187, opened 2016-01-21
Enables saving a new attachment by providing an id and a null revision.

dlandis:
This issue is a regression in the version 2 java-cloudant library. It
looks like it is due to some refactoring in the URIHelper related 
classes.

This change restores the ability described above, which is still 
documented as supported behavior. This change also fixes the JavaDoc
for the same method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ src/test/resources/cloudant-2.properties
 build/
 .gradle/
 local.properties
+bin/
--- a/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
+++ b/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
@@ -23,18 +23,18 @@ import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getRespons
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getResponseList;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.streamToString;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.List;
+import java.util.logging.Logger;
+
 import com.cloudant.client.internal.DatabaseURIHelper;
 import com.cloudant.http.Http;
 import com.cloudant.http.HttpConnection;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
-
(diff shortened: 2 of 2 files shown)


rhyshort:
Thanks for catching this. Can you add a test case that fails, and passes as a result of this change?

Also before we can accept this contribution we'll need a signed CLA:

You can download the CLAs here:

 - [Individual](http://cloudant.github.io/cloudant-sync-eap/cla/cla-individual.pdf)
 - [Corporate](http://cloudant.github.io/cloudant-sync-eap/cla/cla-corporate.pdf)

If you are an IBMer please reach out directly to @mikerhodes via email: [mike.rhodes@uk.ibm.com](mailto:mike.rhodes@uk.ibm.com), as the process is slightly different.

rhyshort:
Thanks for getting the admin bits sorted :)

dlandis:
Hi @rhyshort , 
In light of issue #190 which also has to do with the Database URL Helper class, maybe you want to fix this issue by changing that class as well instead of changing the class I changed. Let me know, thanks

rhyshort:
I'm not sure if that would be the best place to fix the issue, @ricellis what are your thoughts?

ricellis:
Based on the original documentation - I would expect these outcomes for `Database.saveAttachment(InputStream in, String name, String contentType, String docId, String
            docRev)`:

| case | docId | docRev | outcome |
| ---- | ---- | ---- | ---- |
| A | null | null | attach to a new document with a generated ID
| B | "id" | null | attach to a new document with ID: "id"
| C | "id" | "1-rev" | attach to an existing document with ID: "id" at rev: "1-rev"

That said the assertion `assertNotEmpty(docId, "docId");` that @dlandis identified in #186 means that case A has probably never worked as documented. I think cases B and C are both working correctly since the fix for #190 went in.

I'd like to keep the code in `DatabaseURIHelper` only manipulating the URI so we have two choices here:

1. Something similar to this change updating the javadoc (for `Database`) and code in `CouchDatabaseBase`
2. Leave the javadoc as is and delete the assertion line and either i) call the 3 arg form of `saveAttachment` or ii) generate a UUID (in which case the 3 arg form of `saveAttachment` could call the 5 arg form with `null` id and rev).

I'm leaning towards option 2 as the documentation changes from option 1 probably count as breaking the API change (even if the actual implementation never met the API contract).

ricellis:
@dlandis 
I'd like to make sure this is fixed in the next release so based on the discussions above I'm going to close this and supersede it with PR #200.
Hope that's ok.

PR #192, opened 2016-01-25
Added a test for saving a new attachment to an existing doc

ricellis:
*What*
Added a test for saving a new attachment to an existing doc

*How*
Create a document and save it.
Then create an attachment and save it to the document.
Assert that the attachment is correctly saved.

*Testing*
Ran new test against CouchDb and Cloudant service.

reviewer @alfinkel


alfinkel:
+1

PR #198, opened 2016-02-03
Make Database a concrete class again

ricellis:
*What*
Make `Database` a concrete class again

*Why*
The changes from #181 that made `Database` an interface are good for future flexibility, but ultimately produces a binary incompatible change, which would result in a major version change. We don't want that, so for now we will have to extend `Database` instead for other use cases until the next major version bump when there will be an opportunity to change to an interface.

*How*
Undo the `Database` interface extraction from #181, but leave the other internal changes from that PR. Those internal changes were extracting inner deserialization classes and removing unnecessary references to non-public `Database` methods in `Search`.
Return the method implementations to the `Database` class and remove `DatabaseImpl`.
Added a protected constructor to allow `Database` subclasses to instantiate the super from a delegate `Database` instance.

*Testing*
Structural change no new tests needed.

reviewer @emlaver
reviewer @rhyshort


emlaver:
LGTM +1

rhyshort:
+1

PR #200, opened 2016-02-05
186 null id or rev attachments

### LINKED ISSUE #186: Defect: Cannot Save Attachment Without Revision ID

ricellis:
*What*
Fixes #186 regression when saving an attachment to a new document with a given ID (i.e. null revision) and makes `saveAttachment` match the API documentation.

*How*

* Clarified javadoc for the `Database.saveAttachment` with id & rev method.
* Fixed the implementation to generate a UUID instead of throwing an assertion
failure if `null` is passed as a document ID in line with the documentation.
* Added assertions to ensure that we have one of these three combinations

| docId | docRev |
|----|----|
|`null`|`null`|
|not-empty|null|
|not-empty|not-empty|

*Testing*

* New test for a generating a new document with a generated ID when using `saveAttachment`: `attachmentattachmentStandaloneNullIdNullRev`.
* New test for generating a new document with a supplied ID and an attachment: `attachmentStandaloneGivenId`.
* Three new tests for `IllegalArgumentException` cases.
    * `attachmentStandaloneNullDocNonNulLRev`
    * `attachmentStandaloneEmptyDocId`
    * `attachmentStandaloneDocIdEmptyRev`

reviewer @rhyshort 
reviewer @alfinkel

### DIFF
--- a/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
+++ b/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
@@ -16,6 +16,7 @@
 package com.cloudant.client.org.lightcouch;
 
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.assertNotEmpty;
+import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.assertNull;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.close;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.generateUUID;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getAsString;
@@ -311,11 +312,7 @@ public abstract class CouchDatabaseBase {
      * @return {@link Response}
      */
     public Response saveAttachment(InputStream in, String name, String contentType) {
-        assertNotEmpty(in, "in");
-        assertNotEmpty(name, "name");
-        assertNotEmpty(contentType, "ContentType");
-        final URI uri = new DatabaseURIHelper(dbUri).attachmentUri(generateUUID(), name);
-        return couchDbClient.put(uri, in, contentType);
+        return saveAttachment(in, name, contentType, null, null);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -338,7 +334,9 @@ public Response saveAttachment(InputStream in, String name, String contentType,
         assertNotEmpty(in, "in");
         assertNotEmpty(name, "name");
         assertNotEmpty(contentType, "ContentType");
-        assertNotEmpty(docId, "docId");
+        if (docId == null) {
+            docId = generateUUID();
+        }
```
alfinkel: What is the desired behavior if the docId is not null but is of length 0?

rhyshort:
After change Al noted for behaviour +1

src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
ricellis: Thanks Al, that is a good spot.
I've re-added the previous behaviour for a non-null docId, i.e. verifying it is not an empty string.
Additionally I added checks to ensure that if docId has been provided that the revision is either `null` or a non-empty string (for the create and update cases respectively).
All in 3b75b3d including 3 new tests for these `IllegalArgumentException` cases.

alfinkel:
+1

PR #203, opened 2016-02-09
199 doc updates

### LINKED ISSUE #199: Clarify documentation for methods that write objects to the database
Revision identifiers assigned when items are stored in the database are not automatically assigned from the server response to the Java object that was stored.
This is the correct behaviour as we shouldn't write information into the user's existing object instances.

However, this behaviour might be unexpected and should be documented more clearly in the `Database` methods that store objects.

ricellis:
*What*
Update documentation with respect to document revisions on the server and in objects, fixes #199.
It may be worth reviewing the `Database` changes in the commit b036725 rather than at the PR diff level to avoid them being clouded by the whitespace changes which needed to be made because all the javadocs were incorrectly indented.

*How*
Update javadoc on `Database` `save` `post` and `update` methods.
Added section "Document revisions" to javadoc overview.

*Testing*
None, documentation changes.

reviewer @mikerhodes 
reviewer @rhyshort


rhyshort:
+1

overview.html
```diff
@@ -219,6 +220,44 @@ <h3 id="Bulk document CRU">Bulk document CRU</h3>
...
+    document revision. For any of these methods the new document revision can be obtained from
+    the server's {@link com.cloudant.client.api.model.Response} as shown here.
+</P>
+<pre>
+    {@code
+    Foo foo = new Foo();
+    Response response = db.save(foo);
+    foo.setRevision(response.getRevision());
+    }
+</pre>
```
mikerhodes: Do you really need both here?

overview.html
```diff
@@ -219,6 +220,44 @@ <h3 id="Bulk document CRU">Bulk document CRU</h3>
...
+</P>
+<P>
+    Documents retrieved from the server, for example using the
+    {@link com.cloudant.client.api.Database#find(Class, String)} method, are deserialized into
+    objects that will contain the revision as long as the type being deserialized into has a
+    {@code _rev} field or an appropriate custom serializer to ensure that the {@code _rev} is mapped
+    to another field.
+</P>
+<P>
+    Using the correct revision is essential for the
```
mikerhodes: s/essential/required

overview.html
```diff
@@ -219,6 +220,44 @@ <h3 id="Bulk document CRU">Bulk document CRU</h3>
...
+    {@link com.cloudant.client.api.Database#update(Object)} and
+    {@link com.cloudant.client.api.Database#remove(Object)} methods, as well as when performing bulk
+    updates or deletions via the {@link com.cloudant.client.api.Database#bulk(List)} method. If the
+    latest revision is not used then a conflict will result.
```
mikerhodes: This is ambiguous -- a conflict _response_ will be sent but a conflict isn't created.

overview.html
```diff
@@ -219,6 +220,44 @@ <h3 id="Bulk document CRU">Bulk document CRU</h3>
...
+    Foo foo = new Foo();
+    Response response = db.save(foo);
+    foo.setRevision(response.getRevision());
+    }
+</pre>
+<P>
+    Note that storing the revision from a server response after saving a document may be
+    insufficient for the Java object to have the latest revision if the server document is
+    subsequently updated.
+</P>
```
mikerhodes: This sentence isn't clear. I think it needs redrafting -- I had to read it a few times before I understood what you meant. I admit it's a slightly tough one, I couldn't think of anything better immediately.

overview.html
mikerhodes: Does this work for status codes other than 2XX -- I guess they throw an exception so you don't get to line 3?

overview.html
```diff
@@ -219,6 +220,44 @@ <h3 id="Bulk document CRU">Bulk document CRU</h3>
...
+<h3 id="Document revisions">Document revisions</h3>
+<P>
+    This library does not modify the Java objects passed to the
+    {@link com.cloudant.client.api.Database#save(Object)},
+    {@link com.cloudant.client.api.Database#post(Object)},
+    {@link com.cloudant.client.api.Database#update(Object)},
+    or {@link com.cloudant.client.api.Database#bulk(List)} methods. As such once the server write
+    completes the Java object(s) will be out of sync with the database server regarding the
+    document revision. For any of these methods the new document revision can be obtained from
+    the server's {@link com.cloudant.client.api.model.Response} as shown here.
```
mikerhodes: I would say "should be obtained"

overview.html
ricellis: Having `<pre>{@code}</pre>` is the recipe that works best for all javadoc generated pages.
`<pre>` preserves the linebreaks, whilst `{@code}` ensures things like `Map<String, Object>` don't break the html with `<`.

ricellis: Correct, the non 2XX codes throw exceptions.

overview.html
ricellis: Fixed in 3868402.

overview.html
ricellis: Updated in 36f6705 to:
> If the latest revision is not used then the server will reject the change and send a 409 conflict response that results in a {@link DocumentConflictException} being thrown.

wdyt?

overview.html
ricellis: How about 5e3d62d:

> Note that storing the revision from a server response means the object has the latest server revision only at the point in time the write was completed. The server document may be subsequently updated from another source at which point the revision in the object will no longer be the most recent.

?

overview.html
ricellis: Fixed in acc43c9.

ricellis:
I fixed some formatting and missing FQNs in links in 0f7c9c6.

overview.html
mikerhodes: Accepted.

overview.html
mikerhodes: Okay

overview.html
mikerhodes: Definitely better. Do you need to add a note that the behaviour is different for `_bulk_docs`?  Something like `update` and `remove` will get the exception, `_bulk_docs` you need to look at the response (see blah).

overview.html
mikerhodes: Is there a "best practice" we can recommend for people ending up in this state, probably just something like "In this situation, the newest revision should be retrieved via `find` and updated accordingly before retrying `update`"

overview.html
mikerhodes: Okay

overview.html
ricellis: Yes, I think that is valuable, updated in 39b590b.

overview.html
ricellis: In 5e29e45 have split this update between two places:

1. I've added the explanation of how to get the latest revision here.
2. In the paragraph about conflict responses because the latest revision wasn't used I've added the advice to get the latest revision and retry.

mikerhodes: Okay

overview.html
mikerhodes: Okay

mikerhodes:
+1

PR #204, opened 2016-02-09
202 query param semicolon

### LINKED ISSUE #202: ViewRequestBuilder.keys doesn't URL encode semicolon ";"
Cloudant expect view key to be url encoded. When a key contains semicolon (i.e a;b), Database.getViewRequestBuilder().keys() doesn't encode this semicolon and server returns "{"error":"bad_request","reason":"invalid UTF-8 JSON"}. When I encode the key (i.e a%3B) and pass this value to getViewRequestBuilder().keys(..) it double encodes (a%253B). No error is returned by the server and no record is returned.

ricellis:
*What*
Fixes #202 encoding `;` in query parameters.

*How*
Added `;` to the list of characters to encode in the `HierarchicalUriComponents.Type#QUERY_PARAM`.
Clarified method documentation on why the characters were chosen.

*Testing*
Added new `ViewTests.testUriReservedCharsInStartKey` that makes requests using all URI reserved characters in a view start key to ensure no `400 bad request` is encountered.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 # Unreleased
 - [NEW] Constructor for `Database` subclasses.
 - [IMPROVED] Documentation for `Database.findByIndex` to show complete selector.
+- [FIX] `CouchDbException: 400 Bad Request: bad_request: invalid_json` when a
+  query parameter contains a semicolon.
 - [FIX] `NullPointerException` when using `Database.saveAttachment` with a `null`
   revision to attach to a new document with the specified ID.
 - [FIX] `CouchDbException` when using `Database.saveAttachment` to update
--- a/src/main/java/com/cloudant/client/internal/HierarchicalUriComponents.java
+++ b/src/main/java/com/cloudant/client/internal/HierarchicalUriComponents.java
@@ -136,10 +136,23 @@ public class HierarchicalUriComponents {
         QUERY_PARAM {
             @Override
             public boolean isAllowed(int c) {
-                if ('=' == c || '+' == c || '&' == c) {
+                // The query component defined in RFC 3986 does not specify the structure of data
+                // within a query or query parameter, which is deferred to the URI's scheme.
+                // The HTTP scheme in turn does not specify any restrictions on the structure of
+                // data within the query component, only that it must conform to RFC 3986.
+                // As a result the valid characters are determined by the server processing the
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 - [NEW] Constructor for `Database` subclasses.
 - [IMPROVED] Documentation for `Database.findByIndex` to show complete selector.
+- [FIX] `CouchDbException: 400 Bad Request: bad_request: invalid_json` when a
+  query parameter contains a semicolon.
```
tomblench: Technically the fix is for any of `=+&;` but I don't know if you want to go to that level of detail

src/main/java/com/cloudant/client/internal/HierarchicalUriComponents.java
```diff
@@ -136,10 +136,21 @@ public boolean isAllowed(int c) {
...
             @Override
             public boolean isAllowed(int c) {
-                if ('=' == c || '+' == c || '&' == c) {
+                // The query component defined in RFC 3986 does not specify the structure of data
+                // within a query or query parameter, which is deferred to the URI's scheme.
+                // The HTTP scheme in turn does not specify any restrictions on the structure of
+                // data within the query component, only that it must conform to RFC 3986.
+                // As a result the valid characters are determined by the server processing the
+                // request, so in this case we defer to the conventions of
+                // application/x-www-form-urlencoded types forcing encoding of the following:
```
tomblench: This comment is nice and detailed, but is there a reference for `x-www-form-urlencoded` anywhere, is it a standard or convention?

rhyshort:
+1

src/main/java/com/cloudant/client/internal/HierarchicalUriComponents.java
ricellis: It is part of the HTML spec, I have updated with links in 45ba59b.

CHANGES.md
ricellis: It was just `;` the `=`,`+`,`&` were already there, but the diff made it show up badly and it was undocumented as to why they were special.

tomblench:
+1

obidul:
When the binary for this fix will be available for download?

ricellis:
@obidul 2.3.0 should be on maven central very soon.

PR #205, opened 2016-02-15
Added logging information to overview.html

ricellis:
*What*
Added logging info

*How*
Added logging section to the project's root javadoc `overview.html` which appears on the javadoc index page.

*Testing*
N/A documentation.

reviewer: @alfinkel
reviewer: @tomblench

### DIFF
--- a/overview.html
+++ b/overview.html
@@ -64,6 +64,7 @@ There is no need to add a head/title as that is provided by Javadoc.
                 </OL>
             </LI>
             <LI><a href="#J2EE">J2EE</a></LI>
+            <LI><a href="#Logging">Logging</a></LI>
         </OL>
     <LI><a href="#Project">Project</a></LI>
 </OL>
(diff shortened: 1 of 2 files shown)


overview.html
```diff
@@ -523,6 +524,89 @@ <h2 id="J2EE">J2EE</h2>
...
+    This library uses the {@link java.util.logging} package. Classes that log will register a
+    {@link java.util.logging.Logger} under their fully qualified class name.
+    By default java-cloudant does not apply any logging configuration so the system defaults will be
+    used.
+</P>
+
+<P>
+    There are two ways to configure {@link java.util.logging}. It can be configured via a properties
+    file or programmatically. These examples show configuration for writing to the console for FINE
+    level or more severe for all {@code com.cloudant*} classes. The JVM documentation contains more
```
tomblench: "FINE level and above" would be my preferred wording, and it matches the wording in the properties file.

overview.html
```diff
@@ -523,6 +524,89 @@ <h2 id="J2EE">J2EE</h2>
...
+                command line argument like
+                {@code -Djava.util.logging.config.file=/path/to/logging/config} when running the
+                application).
+
+            </LI>
+            <LI>programmatically via the {@link java.util.logging.LogManager}.
+                    <pre>
+                        {@code
+                        LogManager manager = LogManager.getLogManager();
+                        manager.readConfiguration(new FileInputStream("path/to/logging/config");
```
tomblench: Missing closing parenthesis at the end of this line.

overview.html
```diff
@@ -523,6 +524,89 @@ <h2 id="J2EE">J2EE</h2>
...
+            # Limit the messages that are printed on the console to FINE and above.
+            # Full list of levels http://docs.oracle.com/javase/7/docs/api/java/util/logging/Level.html
+            java.util.logging.ConsoleHandler.level=FINE
+
+            # Set the formatter for this handler, there are several default formatter classes
+            # available, or you can create your own custom formatter.
+            java.util.logging.ConsoleHandler.formatter=java.util.logging.SimpleFormatter
+
+            # Set cloudant sync classes to log at level FINE
+            com.cloudant.level=FINE
```
tomblench: How does this FINE setting relate to the above FINE setting? Is the example supposed to show two different ways of setting this (by handler and by class)?

alfinkel:
+1 after Tom's comments are addressed.

overview.html
ricellis: Fixed in 9777c71.

overview.html
ricellis: Fixed in b374b28.

overview.html
ricellis: No, you need both.
Basically the handler determines what the handler will print, whilst the one on the logger determines which levels the logger will send to its handlers.
In this way it is possible to have a console handler with level `SEVERE`, but a file handler with level `INFO` and if both those handlers were registered to the com.cloudant logger then with `com.cloudant.level=INFO` you would get INFO messages in the file, but only SEVERE messages on the console. If you set `com.cloudant.level=SEVERE` then you would only get severe in both the console and the file.

Does that make sense?

tomblench: Understood. Without being overly verbose I think you need to explain to the user why we need both because otherwise it might look like accidental copy/paste or an exposition of two different possibilities: "you can do x or you can do y"

ricellis: Added a note in 1caf53e.

tomblench:
+1

PR #214, opened 2016-02-24
Update DesignDocument#equals to include indexes

### LINKED ISSUE #212: DesignDocument.equals(..) true when design documents differs by indexes
CloudantClient version: 2.2
Java version: 7

DesignDocument doesn't get updated when called DocumentManager.put(..). 
After inspecting the code, it seems that DesignDocument.equals(..) doesn't consider indexes in detecting whether two documents are same. Note that design document are different in-terms of indexes.

Response response = db.getDesignDocumentManager().put(newDesignDoc);

rhyshort:
## What

Correct Issue where ddocs would not be updated if only the `indexes` field had changed.

## How

Regenerate DesignDocument#equals and DesignDocument#hashCode to take
account of the indexes field.

## Testing

New tests added for the `DesignDocument.equals` method in `DesignDocumentTest`, and a test which updates the `indexes` field of a ddoc and saves to the server in `DesignDocumentsTest`.

## Reviewers
reviewer @ricellis 

## Issues

Fixes #212

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 - [NEW] Documentation for logging in project javadoc `overview.html`.
+- [FIX] Fix issue where design documents would not be updated if only the
+  `indexes` field was updated.
 
 # 2.3.0 (2016-02-12)
 - [NEW] Constructor for `Database` subclasses.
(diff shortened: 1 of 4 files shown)


ricellis:
`hashCode()` and `equals()` are no longer consistent, you need to update `hashCode()` with `indexes` as well.

rhyshort:
@ricellis done in 666d0ec

ricellis:
There was a lot of duplication in the new equality tests, so I parameterized them in 5424985.

ricellis:
+1

PR #217, opened 2016-03-01
Http InputStreamGenerator

ricellis:
## What

Added `InputStreamGenerator` for HTTP request replays.

## How

Added `HttpConnection.InputStreamGenerator` interface.
Deprecated `HttpConnection.setRequestBody(InputStream)` method variants.

## Testing
Added new tests to `HttpTest`:
`inputStreamRetryString()`
`inputStreamRetryBytes()`
`inputStreamRetry()`
`inputStreamRetryWithLength()`
`inputStreamRetryGenerator()` 
`inputStreamRetryGeneratorWithLength()`

## Reviewers

reviewer @rhyshort
reviewer @tomblench

### DIFF
--- a/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/src/main/java/com/cloudant/http/HttpConnection.java
@@ -26,6 +26,7 @@ import java.net.URL;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 
@@ -64,7 +65,7 @@ import java.util.logging.Logger;
  *
  * @see java.net.HttpURLConnection
  */
-public class HttpConnection  {
+public class HttpConnection {
 
     private static final Logger logger = Logger.getLogger(HttpConnection.class.getCanonicalName());
     private final String requestMethod;
@@ -75,7 +76,7 @@ public class HttpConnection  {
     private HttpURLConnection connection;
 
     // set by the various setRequestBody() methods
-    private InputStream input;
+    private InputStreamGenerator input;
     private long inputLength;
 
     public final HashMap<String, String> requestProperties;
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -357,4 +398,61 @@ public void disconnect() {
...
     }
 
+    /**
+     * An InputStreamGenerator has a single method getInputStream. Implementors return an
+     * InputStream ready for consuming by the HttpConnection. The purpose of this is to
+     * facilitate regeneration of an InputStream for cases where a payload is going to be resent
+     * for a retry. Implementors must not return the same InputStream instance from multiple calls
+     * to {@link #getInputStream()} unless the stream has been reset.
+     */
+    public interface InputStreamGenerator {
```
rhyshort: This should have the `@since` tag, and I think perhaps breaking it out into its own file will give us more control on HTTP internals in the future, eg renaming `HTTPConnection` class won't result in a breaking API change,

rhyshort: actually scrap that example, I just realised we expose the class anyway.

src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -357,4 +398,61 @@ public void disconnect() {
...
+        /**
+         * @return an InputStream to use to read the body content for a HTTP request
```
rhyshort: no `@throws` tag present, should probably have a summary too even though its only a getter

ricellis: Added in de1968e.

src/main/java/com/cloudant/http/HttpConnection.java
ricellis: `@since` added in de1968e along with on the generator `setRequestBody` methods.

rhyshort:
+1

ricellis:
@tomblench replaced stream copy in 43da0b1

tomblench:
+1

PR #219, opened 2016-03-03
218 remove expect

### LINKED ISSUE #218: `Expect: 100-continue` prevents cookie renewal
2.x
Any
No

The session cookie renewal fails if the request that needs to get a new cookie is a `POST`. The problem is that `HttpURLConnection` throws a `java.net.ProtocolException` for the `403` response (it is expecting a `100`) and that unchecked exception breaks out of the response handling preventing the cookie renewal request taking place.

ricellis:
## What
Fixes #218 and other cases where `java.net.ProtocolException` is thrown by `Expect:100-continue` receiving a non-`100` return code.

## How
Removed `Expect:100-continue` header from requests with bodies.

## Why
The `100-continue` check delays the sending of a body until we are sure a server will accept the request. However, the value of this optimization is in the cases where we receive a non-`100` code and therefore avoid sending a (potentially large) body. In cases where the request gets a `100` we are going to send the body anyway so we have expended an extra round trip.
The `HttpURLConnection` does not have a well defined API for handling the `Expect:100-continue` cycle and in cases where non-`100` codes are returned (e.g. `403` for an expired cookie) this results in a `ProtocolException` being thrown. However, this exception is not very specific to the expectation failure and there are other potential `ProtocolException` scenarios that prevent us from simply catching the exception. As a result it makes sense to remove the `100-continue` cycle until there is a better API to handle it. The extra round trip reduces the value of the optimization except in cases with relatively large bodies anyway.

## Testing
Removed `100-continue` tests and unused test classes.
Added a new test for cookie renew scenario with a `POST`.
Added a new test for BasicAuth (as it is a workaround for cookie issues).

## Reviewers

reviewer @rhyshort
reviewer @tomblench 

## Issues
#218


rhyshort:
+1

tomblench:
Does the removal of `SimpleHttpServer.java` imply that the OKHttp mock server is preferred for any similar tests in the future?

ricellis:
The `MockHttpServer` is a lot easier to use than `SimpleHttpServer`, the only reason it hadn't been removed before was because we couldn't mock a `100 continue` response so that was the only remaining user.

tomblench:
+1

PR #220, opened 2016-03-03
215 cxn spec

### LINKED ISSUE #215: Issue with IBM JDK 7 or 8 resulting SSL handshake failure
Trying to upgrade to java-cloudant library 2.3.0 with the okhttp dependency, and encountered a weird SSL handshake failure. For testing purpose, I created a simple test Java program that just connects to my Cloudant account and list all of my databases. Please see the attached simple code below. It works perfectly fine with Oracle Java 7 or 8, but when running it in IBM JDK 7 or 8, it fails and produces an SSL handshake failures. Attached the simple java class file here: 
[ListDBTest.zip](https://github.com/cloudant/java-cloudant/files/147039/ListDBTest.zip)
You need to enter your Cloudant account, username, and password in the file to run it, and also use -Djavax.net.debug=all to see all the SSL handshake trace if needed.

Some excerpt of error log:
......
main, handling exception: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
main, called close()
main, called closeInternal(true)
Feb 25, 2016 9:38:01 AM com.cloudant.http.interceptors.CookieInterceptor interceptResponse
SEVERE: Failed to get response code from request
javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
	at com.ibm.jsse2.j.a(j.java:33)
	at com.ibm.jsse2.j.a(j.java:31)
	at com.ibm.jsse2.qc.b(qc.java:190)
	at com.ibm.jsse2.qc.a(qc.java:456)
	at com.ibm.jsse2.qc.h(qc.java:352)
	at com.ibm.jsse2.qc.a(qc.java:523)
	at com.ibm.jsse2.qc.startHandshake(qc.java:730)
	at com.squareup.okhttp.Connection.connectTls(Connection.java:235)
	at com.squareup.okhttp.Connection.connectSocket(Connection.java:199)
	at com.squareup.okhttp.Connection.connect(Connection.java:172)
......

Now thinking of SSL handshake failures, one might think of invalid certificate and trust CA issues or cipher incompatibility, so I spent quite some time to trace the SSL hand shaking, did not find any such issue. Actually the first round SSL handshake was successful (from ClientHello, ServerHello, ServerHelloDone, to Finished), and the client and Cloudant was communicating with app data on user authentication and setting cookie. But in the IBM JDK 7 or 8 case, the SSL connection then broken down after that at the CookieInterceptor class (the client would re-initiate an SSL hand-shake, but the server will no longer respond as it thinks there is already an established communication session), while running under Oracle Java 7 or 8 it continues and works fine.

So my guess is that somehow the interceptor mechanism (possibly the CookieInterceptor) somehow disrupts the normal SSL flow in the IBM JDK 7 or 8 case, resulting the problem, on the other hand it does not cause any issue in Oracle Java 7 or 8.

Please let me know if there are some configuration changes or other changes I can apply to prevent this issue. Your help is greatly appreciated.

Thanks much.
Bing.

ricellis:
## What

Fixes #215 to prevent `javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure` with okhttp and IBM Java.

## How

Changed the `ConnectionSpec` for okhttp to use the enabled protocols and cipher suites from the JVM rather than the default hardcoded list in okhttp's `MODERN_TLS` spec. This resolves the issue with the cipher suite prefixes not matching and also means that the enabled protocols and cipher suites match for both the okhttp and `HttpURLConnection` cases.
Upgraded okhttp to 2.7.5 to allow use of `allEnabled*` methods on `ConnectionSpec`.

## Testing
No additional tests, tested as working on IBM Java.

## Reviewers
reviewer @rhyshort
reviewer @tomblench 

## Issues
#215

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Unreleased
 - [NEW] Documentation for logging in project javadoc `overview.html`.
+- [IMPROVED] Upgraded optional okhttp to 2.7.5.
+- [FIX] Issue of `javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure` on
+  IBM Java with okhttp. SSL connections using okhttp are now configured to use the JVM enabled
+   protocols and cipher suites in the same way as the `HttpURLConnection`.
 - [FIX] Issue where a `java.net.ProtocolException` was thrown if the cookie had expired when a
   request that included a body was sent. Note that the client no longer uses the
   `Expect:100-continue` header on requests.
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Gradle with [optional `okhttp-urlconnection` dependency](#optional-okhttp-depend
 ```groovy
 dependencies {
     compile group: 'com.cloudant', name: 'cloudant-client', version: '2.3.0'
-    compile group: 'com.squareup.okhttp', name: 'okhttp-urlconnection', version: '2.5.0'
+    compile group: 'com.squareup.okhttp', name: 'okhttp-urlconnection', version: '2.7.5'
 }
 ```
 
@@ -54,7 +54,7 @@ Maven with [optional `okhttp-urlconnection` dependency](#optional-okhttp-depende
 <dependency>
   <groupId>com.squareup.okhttp</groupId>
   <artifactId>okhttp-urlconnection</artifactId>
-  <version>2.5.0</version>
+  <version>2.7.5</version>
 </dependency>
 ~~~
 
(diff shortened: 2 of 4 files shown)


rhyshort:
Is there a potential problem if we do this for all JVM's and not just IBM's?

ricellis:
I don't foresee a problem, it will just be using the same protocols and cipher suites as the `HttpsURLConnection` on any JVM. If particular versions of TLS or particular ciphers are broken then users need to either configure their JVM or get a JVM update for those to be disabled/removed as would be the case normally. Additionally the handshake is a negotiation so protocols or ciphers may well be ruled out by the server side.

rhyshort:
+1

tomblench:
+1

PR #221, opened 2016-03-03
Set the User-Agent via a Version object loaded at runtime

rhyshort:
## What

Change how the user-agent is generated for HTTP requests in preparation for the splitting out the HTTP classes as a separate jar.

## How

- Created a new Interface `Version` which contains a single method `getVersionString()` which returns the string to be used as the user-agent. 
- AgentHelper is renamed to Version in the new `com.cloudant.library` package.
- The HTTPConnection will load the class at Runtime and default to the JVM default user-agent if it fails.

## Testing

No new tests, only modified a test case to use the renamed class and interface method.

## Reviewers

reviewer @tomblench 
reviewer @ricellis

### DIFF
--- a/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/src/main/java/com/cloudant/http/HttpConnection.java
@@ -11,7 +11,6 @@
 package com.cloudant.http;
 
 import com.cloudant.http.interceptors.BasicAuthInterceptor;
-import com.cloudant.http.internal.AgentHelper;
 import com.cloudant.http.internal.DefaultHttpUrlConnectionFactory;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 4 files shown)


tomblench:
Agree with @ricellis - I also get a huge diff on `src/test/java/com/cloudant/tests/CloudantClientTests.java` - line endings changed?

rhyshort:
@tomblench I think so, I must I have clicked the wrong button in Idea.

src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -68,6 +67,23 @@
 public class HttpConnection {
 
     private static final Logger logger = Logger.getLogger(HttpConnection.class.getCanonicalName());
+
+    private static final String USER_AGENT;
+
+    static {
+        String ua = "java-cloudant-http/unknown";
```
tomblench: If/when this component is shared between sync-android and java-cloudant, what should the default string be?

tomblench: Also we have a version here and a user-agent, which is which? Surely a version is eg "1.2.3" and user-agent has all the gory details about platform, OS, etc. Is that the responsibility of the `Version` class or this static block?

rhyshort: Its the responsibility of the Version class to provide anything it wants to use as the UA string, which for our libs is a version string with some extra bits. When it's shared between sync-android, it will still be this as the default since its the default for the package, eg this is the java-cloudant-http so that's what it should be reported as. Perhaps a rename to getUserAgent? instead of getVersion?

tomblench: OK, perhaps getUserAgent is better. This means we might have some duplicated code for getting platform versions etc.

rhyshort: It's something we could perhaps sort out later, basically, in sync-android when sync is built on java-cloudant, we'll be able to do alot more, so the version stuff can be expanded to a few more methods etc to make it easier, but thats a long way down the line yet.

rhyshort: done in 2d48cc3

ricellis:
+1

tomblench:
+1

PR #222, opened 2016-03-03
Update stored cookie when a Set-Cookie header is received

### LINKED ISSUE #195: SessionCookie Interceptor should check headers for a refreshed cookie
SessionCookieInterceptor should check headers for a a refreshed cookie. CouchDB will refresh the cookie if more than 10% of the lifetime of the cookie presented has expired.

See https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd_auth.erl#L235.

We should look for the `Set-Cookie` header on responses and use the cookie it presents, this will reduce the amount of network requests we send when in a long running application.

rhyshort:
## What

Update stored cookie when a Set-Cookie header is received.

## How
Check for the `Set-Cookie` HTTP header, and if its present extract the cookie form its value and replace
the one in memory with the newly issued cookie. Although we already do this for the expiry case it is possible for the server to send a renewed cookie before expiry, so by handling those cases we can avoid the extra round trips of the "expired cookie failed request"/"get new cookie"/"repeat failed request" cycle.

## Testing

Added an extra test to simulate a `Set-Cookie` header being sent to the client.

## Reviewers
@ricellis

## Issues

- #195

### DIFF
--- a/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
+++ b/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
@@ -97,6 +97,13 @@ public class CookieInterceptor implements HttpConnectionRequestInterceptor,
     public HttpConnectionInterceptorContext interceptResponse(HttpConnectionInterceptorContext
                                                                       context) {
         HttpURLConnection connection = context.connection.getConnection();
+
+        String cookieHeader = connection.getHeaderField("Set-Cookie");
+        if(cookieHeader != null){
+            cookie = this.extractCookieFromHeaderValue(cookieHeader);
+            return context;
+        }
+
         try {
             boolean renewCookie = false;
             int statusCode = connection.getResponseCode();
@@ -174,7 +181,7 @@ public class CookieInterceptor implements HttpConnectionRequestInterceptor,
             if (responseCode / 100 == 2) {
 
                 if (sessionHasStarted(connection.getInputStream())) {
-                    return cookieHeader.substring(0, cookieHeader.indexOf(";"));
+                    return this.extractCookieFromHeaderValue(cookieHeader);
                 } else {
                     return null;
                 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -131,6 +131,10 @@ public HttpConnectionInterceptorContext interceptResponse(HttpConnectionIntercep
                     renewCookie = true;
                     break;
                 default:
+                    String cookieHeader = connection.getHeaderField("Set-Cookie");
```
ricellis: There may be some possible responses such as a `403` that is not the expiry case. If that included a `Set-Cookie` this would miss it. I wonder if we should check for the header earlier and then not worry about response codes?

rhyshort: done in c8db300

src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -406,6 +407,49 @@ public void cookie403Renewal() throws Exception {
...
+        c.executeRequest(Http.GET(c.getBaseUri()));
+        mockWebServer.takeRequest(); // cookie
+        mockWebServer.takeRequest(); // actual get
+
+        //make another get
+        c.executeRequest(Http.GET(c.getBaseUri()));
+
+
+        // also assert that there were 4 calls
+        assertEquals("The server should have received 4 requests", 3, mockWebServer
```
ricellis: The message of this assert and the comment above don't agree with what is being asserted.

ricellis: Also since `takeRequest` is blocking (potentially until the tests time out if there were not the correct number of requests or an exception thrown) you may want to re-order this so that the assert comes before the `takeRequest` calls.

rhyshort: done in 2f5e342

ricellis:
+1

PR #224, opened 2016-03-07
Http cxn close

ricellis:
## What
Some small improvements identified by leaking connections investigation and findbugs.

## How
Ensured stream closed after typed get calls in `CouchDbClient`.
Closed both input and output streams after writing request bodies in `HttpConnection`.
Used more efficient `entrySet` loop when setting request properties in `HttpConnection`.
Renamed `com.cloudant.library.Version` to `com.cloudant.library.LibraryVersion` to avoid potential name clash.

## Testing
No new tests added, existing tests already exercise these paths.

## Reviewers
@rhyshort 
@tomblench

### DIFF
--- a/cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
@@ -274,18 +274,15 @@ public class HttpConnection {
                     connection.setChunkedStreamingMode(1024);
                 }
 
-                int bufSize = 1024;
-                int nRead = 0;
-                byte[] buf = new byte[bufSize];
                 InputStream is = input.getInputStream();
                 OutputStream os = connection.getOutputStream();
-
-                while ((nRead = is.read(buf)) >= 0) {
-                    os.write(buf, 0, nRead);
+                try {
+                    IOUtils.copy(is, os);
+                    os.flush();
+                } finally {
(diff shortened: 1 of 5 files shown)


rhyshort:
+1

ricellis:
@tomblench replaced the "manual" stream copy with `IOUtils.copy()` in a6070b1.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -274,18 +275,19 @@ public HttpConnection execute() throws IOException {
...
                 InputStream is = input.getInputStream();
                 OutputStream os = connection.getOutputStream();
 
-                while ((nRead = is.read(buf)) >= 0) {
-                    os.write(buf, 0, nRead);
+                try {
+                    try {
+                        IOUtils.copy(is, os);
+                        os.flush();
+                    } finally {
```
tomblench: I think we can avoid these nested `try` blocks by using `IOUtils.closeQuietly()`

ricellis: I've removed one of them in 8891940.
As discussed I think one needs to remain in case the `flush` throws an exception.

tomblench:
+1

Issue #175, opened 2016-01-04
Test Failures against Cloudant Local


rhyshort:
Internal jenkins server reports failures when running tests against Cloudant Local, see test report http://cloudantsync001.bristol.uk.ibm.com:8080/job/Java-Cloudant%20Nightly/CouchFlavour=cloudantlocal/lastCompletedBuild/testReport/

ricellis:
After merging #180 there appear to be two remaining groups of failures:
i) Some tests fail because the client cannot deserialize an array sequence id (it expects a String)
ii) Tests that use the `_replicate` endpoint are failing with 404 database not found.

ricellis:
As expected the group ii failures have been resolved by creating the `_replicator` database.

ricellis:
This was fixed by #227

PR #227, opened 2016-03-09
Updated seq ids to use JSON

ricellis:
## What
Updated seq ids to use JSON

## How
Updated the classes:
`ChangesResult`
`ReplicationResult`
`DbInfo`
`CouchDbInfo`
`Task`
to use `JsonElement` internally for storing the sequence ids for compatibility with non-String seq id
 (e.g. those used by some versions of Cloudant Local). Return the String representation of JSON for API compatibility. Note that some seq ids are still treated as numbers, but may be updated later if the API changes.

Updated CHANGES.md.

## Testing
Ran tests successfully with Cloudant Data Layer Local Edition.

## Reviewers
reviewer @rhyshort 

## Issues
#175

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 # Unreleased
 - [NEW] Documentation for logging in project javadoc `overview.html`.
 - [IMPROVED] Upgraded optional okhttp to 2.7.5.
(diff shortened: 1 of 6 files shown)


rhyshort:
looks good +1

PR #228, opened 2016-03-10
Fixed gradle for split artifacts

ricellis:
## What
Fixed gradle for releases with subprojects.

## How
Fixed gradle comments.
Moved optional dependency processing into cloudant-http project.
Updated snapshot version.

## Testing
Checked optional tag in local and snapshot maven repostories.

## Reviewers
reviewer @tomblench


tomblench:
+1

PR #233, opened 2016-03-21
Trimmed selector string before parsing

ricellis:
## What
Fixed regression where a selector with leading whitespace caused a `JsonParseException`.

## How
Trimmed selector string before parsing.

## Testing
Added a new test `testNotNullIndexMovieNameAndYearSelectorStringWithSpace` with whitespace at the start (and within) the selector string.


mikerhodes:
+1, probably put 3d5c335 first and squash 3073ecf and 908e646 together into a single commit (or perhaps just squash them all together).

PR #234, opened 2016-03-23
Changed to use JVM default chunk size

ricellis:
## What
Changed to use JVM default chunk size.

## How
Changed call to `setChunkedStreamingMode(0)` to use JVM default.
Moved call to before request interceptors to allow the size to be changed in an interceptor if necessary.
Increased size of buffer used for HTTP output stream to avoid artificially limiting the chunk size. It is worth noting that current `HttpURLConnection` implementations appear to treat the chunk size as a preference only and in testing the output buffer size was more likely to be used as the chunk size than the size suggested by `setChunkedStreamingMode()`.

## Testing
Added new test with chunking `HttpTest#testChunking`.

## Reviewers
reviewer @rhyshort 
reviewer @brynh

### DIFF
--- a/cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
+++ b/cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
@@ -251,6 +251,24 @@ public class HttpConnection {
                 connection.setRequestProperty("Content-type", contentType);
             }
 
+            // We set up the output config before the interceptors to allow the configuration to be
+            // modified. For example an interceptor might change the chunk size by calling
+            // context.connection.getConnection().setChunkedStreamingMode(16384);
+            if (input != null) {
+                connection.setDoOutput(true);
+                if (inputLength != -1) {
+                    // TODO Remove this cast to int when the minimum supported level is 1.7.
+                    // On 1.7 upwards this method takes a long, otherwise int.
+                    connection.setFixedLengthStreamingMode((int) this.inputLength);
+                } else {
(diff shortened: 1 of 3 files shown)


cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -251,6 +251,21 @@ public HttpConnection execute() throws IOException {
...
+            // modified. For example an interceptor might change the chunk size by calling
+            // context.connection.getConnection().setChunkedStreamingMode(16384);
+            if (input != null) {
+                connection.setDoOutput(true);
+                if (inputLength != -1) {
+                    // TODO on 1.7 upwards this method takes a long, otherwise int
+                    connection.setFixedLengthStreamingMode((int) this.inputLength);
+                } else {
+                    // TODO some situations where we can't do chunking, like multipart/related
+                    /// https://issues.apache.org/jira/browse/COUCHDB-1403
```
rhyshort: Should we fix this todo while we are here?

ricellis: The comments just got copied when I moved that block. I've discussed with @tomblench and we can't do the first TODO until 1.7 is the minimum supported level. The second TODO isn't really a TODO it is more an FYI, I will update accordingly.

ricellis: Updated the TODOs in db9bbc7.

rhyshort:
+1

brynh:
+1

Issue #239, opened 2016-04-11
Failure when removing design documents (for indexes)


stephen-kruger:
Cloudant 2.4.1
Using a previously suggested method to list design documents : 
```java
List<String> designDocIds = database.getAllDocsRequestBuilder()
                                    .startKey("_design/")
                                    .endKey("_design0")
                                    .build()
                                    .getResponse()
                                    .getDocIds();
```

I then remove each one using : 
`DesignDocumentManager ddm = database.getDesignDocumentManager();
ddm.remove(designDocId);`

This works fine for views etc, but when it hits a DD for Indexes, I get this error : 
```
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 133
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:176)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:40)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:187)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:145)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:93)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:172)
	at com.google.gson.Gson.fromJson(Gson.java:803)
	at com.google.gson.Gson.fromJson(Gson.java:741)
	at com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getResponse(CouchDbUtil.java:211)
	at com.cloudant.client.org.lightcouch.CouchDbClient.get(CouchDbClient.java:329)
	at com.cloudant.client.org.lightcouch.CouchDatabaseBase.find(CouchDatabaseBase.java:88)
	at com.cloudant.client.api.Database.find(Database.java:536)
	at com.cloudant.client.api.DesignDocumentManager.remove(DesignDocumentManager.java:154)```

ricellis:
[...]

PR #242, opened 2016-04-12
239 dd remove rev HEAD

ricellis:
## What

Avoid deserializing design documents on remove calls.

## How

Used a `HEAD` request and `ETag` header to get revision information instead of doing a `find()` which did a `GET` of the entire document content plus deserialization.

## Testing

Code path is covered by the `DesignDocumentsTest#designDocRemoveNoPrefix`.
Added two new tests for exception cases:
* couchDbExceptionIfIOExceptionDuringDDocRemove
* couchDbExceptionIfNoETagOnDDocRemove

## Reviewers
reviewer @emlaver
reviewer @alfinkel

## Issues
This addresses the second part of #239.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/DesignDocumentManager.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/DesignDocumentManager.java
@@ -18,8 +18,11 @@ import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.assertNotE
 
 import com.cloudant.client.api.model.DesignDocument;
 import com.cloudant.client.api.model.Response;
+import com.cloudant.client.internal.DatabaseURIHelper;
+import com.cloudant.client.org.lightcouch.CouchDbException;
 import com.cloudant.client.org.lightcouch.NoDocumentException;
 import com.cloudant.client.org.lightcouch.internal.CouchDbUtil;
+import com.cloudant.http.Http;
 import com.google.gson.Gson;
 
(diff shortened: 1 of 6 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/DesignDocumentManager.java
```diff
@@ -151,8 +156,21 @@ public DesignDocument get(String id, String rev) {
...
+        } catch (IOException ioe) {
+            throw new RuntimeException("There was an error getting revision information so the " +
+                    "design document cannot be removed.", ioe);
+        }
+        if (revision != null) {
+            revision = revision.substring(1, revision.length() - 1);
+            return db.remove(id, revision);
+        } else {
+            throw new RuntimeException("No ETag header found for design document with id " + id);
+        }
```
alfinkel: Would it be helpful if a couple of tests were added for this new logic?  I can see that the end game of the logic is the same as it was before but it looks like two new conditions exist where exceptions are being thrown now.

ricellis: Added in cd6adb5. It helpfully demonstrated that I need not catch an `IOException` because I didn't need to call `execute` as `executeRequest` was already doing that.
Also changed the no ETag case to use a `CouchDbException` rather than a generic `RuntimeException` because if there is no ETag header that really is a server problem.

alfinkel:
+1

emlaver:
+1

PR #244, opened 2016-04-13
Fixed regressions in use_index option

### LINKED ISSUE #243: Error 'Invalid key use_index' returned when querying against Cloudant-0.1.0-local
The Cloudant query API call, implemented using java-cloudant's "cloudant-client" classes, runs successfully against the Cloudant version 1.0.2 (if db exists on cloudant.com), but same API call fails against the local installation of Cloudant (db version 0.1.0-local).

**Steps followed**:
1) Created an index using POST http call for ` /_index` endpoint
2) Invoke a cloudant query using the Database.findByIndex(String selectorJson, Class<T> classOfT) method:
`db.findByIndex("\"selector\": {\"docType\": {\"$eq\": \"person\"}}",Person.class)`
3) Error returned: "com.cloudant.client.org.lightcouch.CouchDbException: 400 Bad Request: invalid_key: Invalid key use_index for this request."

**Root cause identified**:
FindByIndexOptions class initializes the useIndex attribute by default ( `private JsonArray useIndex = new JsonArray()` ) therefore the generated query body contains "use_index" parameter even when no explicit index is provided.

ricellis:
## What

Fixed regressions introduced in 2.2.0 where:

1. `use_index=[]` being added to requests using `FindByIndexOptions` when `useIndex()` had not been called.
1. `use_index` being specified as an array when it should have been a string for the design document only case.

## How

Changed to use `JsonElement` instead of `JsonArray` to support both
string and array use cases.

Stopped adding to the array if `useIndex` is called more than once,
replace the property instead.

Added the `use_index` option to the request only in cases where
`useIndex` was called instead of adding an empty array.

## Testing

Added 2 new tests to check the type of the use_index property:
-useIndexDesignDocJsonTypeIsString
-useIndexDesignDocAndIndexNameJsonTypeIsArray

Added a new test to check that the use_index property is not
specified if the option was not used:
-useIndexNotSpecified

Added a test to validate that the use_index property is replaced
if the method is called multiple times on the same FindByIndexOptions:
-useIndexReplaced

## Reviewers

reviewer @mikerhodes
reviewer @rhyshort
## Issues
Fixes #243

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
@@ -14,8 +14,11 @@
 
 package com.cloudant.client.api.model;
 
+import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.assertNotNull;
+
 import com.cloudant.client.api.Database;
 import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonPrimitive;
 
 import java.util.ArrayList;
@@ -48,7 +51,8 @@ public class FindByIndexOptions {
     private List<IndexField> sort = new ArrayList<IndexField>();
     private List<String> fields = new ArrayList<String>();
     private Integer readQuorum;
-    private JsonArray useIndex = new JsonArray();
+    // Either a string for design document only or an array of ["designDoc", "indexName"]
+    private JsonElement useIndex = null;
 
     /**
      * @param limit limit the number of results return
@@ -84,6 +88,7 @@ public class FindByIndexOptions {
      * @return this to set additional options
      */
     public FindByIndexOptions sort(IndexField sort) {
+        assertNotNull(sort, "sort");
         this.sort.add(sort);
         return this;
     }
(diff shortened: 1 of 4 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/model/FindByIndexOptions.java
```diff
@@ -120,10 +121,10 @@ public FindByIndexOptions useIndex(String designDocument) {
...
      */
     public FindByIndexOptions useIndex(String designDocument, String indexName) {
-        JsonPrimitive jsonDesign = new JsonPrimitive(designDocument);
-        JsonPrimitive jsonIndex = new JsonPrimitive(indexName);
-        this.useIndex.add(jsonDesign);
-        this.useIndex.add(jsonIndex);
+        JsonArray index = new JsonArray();
+        index.add(new JsonPrimitive(designDocument));
+        index.add(new JsonPrimitive(indexName));
+        this.useIndex = index;
```
mikerhodes: Do either of the `useIndex` methods need to check the args to the method are not null?

ricellis: They (and the other `FindByIndexOptions`) should. Fixed in e580e13.

mikerhodes:
+1

rhyshort:
👍

PR #246, opened 2016-04-21
Reduced the length of the User-Agent string

ricellis:
## What

Removed very specific platform version information and reduced length of User-Agent string.

## How

Removed java.runtime.version & os.version property values.
Removed unnecessary “Java” word - we know where the Java info is in the string without it.
Re-ordered the properties so that java.version, is followed by vendor and os name is followed by arch.
Used single `/` separators instead of more complex groupings of `;[]()`.

## Testing

Updated test checks to match new form.

## Reviewers
reviewer @rhyshort
reviewer @alfinkel

## Issues

See case 64095.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Unreleased
+- [IMPROVED] Reduced the length of the User-Agent header string.
 - [IMPROVED] Use a more efficient HEAD request for getting revision information when using
   `DesignDocumentManager.remove(String id)`.
 - [FIX] Regression where `_design/` was not optional in ID when using `DesignDocumentManager` methods.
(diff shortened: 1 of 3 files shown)


rhyshort:
+1

alfinkel:
I think when composing the User-Agent string for python-cloudant we used java-cloudant as the template.  Is there any benefit to standardizing the User-Agent content across all of the client libraries as much as possible?  If so, I can create a separate issue in python-cloudant and make this change there as well.  Let me know your thoughts @ricellis and @rhyshort.

rhyshort:
Yeah, being standard would be helpful.

alfinkel:
+1

PR #249, opened 2016-05-05
60008 http 429 backoff

ricellis:
## What

Added handling for `429 Too many requests` server responses.

## How

* cloudant-http
    * Expose the current number of retries from `HttpConnection`.
    * Added cause constructor to `HttpConnectionInterceptorException`.
    * Added unwrapping of `IOException` from interceptor thrown `HttpConnectionInterceptorException`s.
    * Create a `RequestLimitInterceptor` that checks for `429` status codes and sleeps before replaying the request.
    * The sleep duration is a doubling backoff starting at 0.25 s and doubling with each retry (up to 10 tries by default) unless a `Retry-After` header was specified by the server in which case that is parsed (as seconds) and used as the duration.
    * Whilst it would be nice to reschedule the request rather than sleeping the current `HttpConnection` approach does not have an executor to support that.
    * Added an instance of the `RequestLimitInterceptor` to all `HttpConnection`s.

* cloudant-client
    * Added a `TooManyRequestsException` to be thrown if a `429` is received and there are no retries remaining

## Testing

* Removed `MockWebServerResource` wrapping of `MockWebServer` since `MockWebServer` is usable as a JUnit rule anyway. Renamed to `MockWebServerResources` for the remaining utilities.
* Configured a reduced starting backoff time (1 ms) to speed up testing.
* Added new `HttpTest` class tests:
    * test429Backoff - test that a request is replayed in response to a 429.
    * test429BackoffMaxDefault - test the default number of retries and backoff duration.
    * test429BackoffMaxConfigured - test a configured number of retries.
    * test429BackoffRetryAfter - test backoff duration specified in `Retry-After` response header is honoured.

## Reviewers

reviewer @brynh 
reviewer @rhyshort

## Issues

Internal case 60008.

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
@@ -9,6 +9,7 @@ import static org.junit.Assert.fail;
 
 import com.cloudant.client.api.CloudantClient;
 import com.cloudant.client.org.lightcouch.CouchDbException;
+import com.cloudant.client.org.lightcouch.TooManyRequestsException;
 import com.cloudant.http.Http;
 import com.cloudant.http.HttpConnection;
 import com.cloudant.http.HttpConnectionInterceptorContext;
@@ -16,6 +17,7 @@ import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
 import com.cloudant.http.interceptors.BasicAuthInterceptor;
 import com.cloudant.http.interceptors.CookieInterceptor;
+import com.cloudant.http.interceptors.RequestLimitInterceptor;
 import com.cloudant.test.main.RequiresCloudant;
 import com.cloudant.tests.util.CloudantClientResource;
 import com.cloudant.tests.util.DatabaseResource;
@@ -23,10 +25,12 @@ import com.cloudant.tests.util.MockWebServerResources;
 import com.cloudant.tests.util.Utils;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import com.squareup.okhttp.mockwebserver.Dispatcher;
 import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
 import com.squareup.okhttp.mockwebserver.RecordedRequest;
 
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
@@ -38,6 +42,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.reflect.Field;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.nio.charset.Charset;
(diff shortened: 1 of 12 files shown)


cloudant-http/src/main/java/com/cloudant/http/interceptors/RequestLimitInterceptor.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+public class RequestLimitInterceptor implements HttpConnectionResponseInterceptor {
+
+    private static final Logger logger = Logger.getLogger(RequestLimitInterceptor.class.getName());
+    private static int _initialSleep = 250; // Default to 250 ms initial backoff time
```
rhyshort: isn't using an `_` here against java naming convention?

rhyshort:
I don't like the way the interceptor is configured, with it being a static we do get it for free rather than adding it via the client builder, however that being said, I think it should be possible to configure on a per client basis, to give the user more granular control over the max retries.

cloudant-http/src/main/java/com/cloudant/http/interceptors/RequestLimitInterceptor.java
ricellis: Yep, not sure what was going on there, must have a been one of _those_ days!
Fixed in 191373b.

cloudant-http/src/main/java/com/cloudant/http/interceptors/RequestLimitInterceptor.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+public class RequestLimitInterceptor implements HttpConnectionResponseInterceptor {
+
+    private static final Logger logger = Logger.getLogger(RequestLimitInterceptor.class.getName());
+    private static int initialSleep = 250; // Default to 250 ms initial backoff time
```
rhyshort: Why is this field not final? Is it supposed to be configurable? If so it should be configurable per client/interceptor not per JVM.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -564,4 +568,125 @@ public InputStream getInputStream() throws IOException {
...
+                return response429;
+            }
+        });
+
+
+        long startTime = System.currentTimeMillis();
+        try {
+            CloudantClient c = CloudantClientHelper.newMockWebServerClientBuilder(mockWebServer)
+                    .build();
+            c.executeRequest(Http.GET(c.getBaseUri()).setNumberOfRetries(3));
```
rhyshort: Now that we are exposing the number of retries per connection, shouldn't we expose a way of doing it per client via the builder?

cloudant-http/src/main/java/com/cloudant/http/interceptors/RequestLimitInterceptor.java
ricellis: It is not intended to be configurable. It wasn't made final so it could be reduced (via reflection) during testing to avoid having the tests run long.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
ricellis: The setter for retries has always exposed on `HttpConnection`, I added a getter so that the interceptor could use the value.
I think 10 is a sensible default and if someone really wants to configure it they can use a custom `HttpConnection` and `executeRequest`.

cloudant-http/src/main/java/com/cloudant/http/interceptors/RequestLimitInterceptor.java
brynh: I wonder if a comment to that effect should be added to avoid confusion in future?

brynh:
+1

rhyshort:
+1

cloudant-http/src/main/java/com/cloudant/http/interceptors/RequestLimitInterceptor.java
ricellis: Fixed comment in 3506e2e.

PR #250, opened 2016-05-09
Fixed deserialization of query design documents

### LINKED ISSUE #239: Failure when removing design documents (for indexes)

ricellis:
## What
Fixed deserialization of query design documents

## How

Replaced `String` map/reduce fields with `JsonElement` in `DesignDocument.MapReduce` to allow for either a JSON string or JSON object to be stored as a map function.
Default for `JsonElement.toString()` is to enclose JSON string with `"` (i.e. escaped quotes `\"`) in the Java string. Added logic to remove these leading and trailing quotes from JSON string map/reduce functions (i.e. javascript language case) to maintain backwards compatibility with existing behaviour. 
Updated CHANGES.md.

## Testing

Added tests for serializing and deserializing map functions of string and object types for the javascript and query language design documents.

## Reviewers
reviewer @rhyshort
reviewer @emlaver
## Issues
Fixes #239 (part 1), part 2 was done in #242

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/model/DesignDocument.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/model/DesignDocument.java
@@ -15,7 +15,9 @@
 
 package com.cloudant.client.api.model;
 
+import com.google.gson.Gson;
 import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.annotations.SerializedName;
 
@@ -55,6 +57,7 @@ import java.util.Map;
  * Map<String, String> updates = new HashMap<String, String>();
  * updates.put("newUpdateHandler", "function (doc, req) { ... }");
  * ddoc.setUpdates(updates);
+ * // Update the design document
  * database.update(ddoc);
  *
  * }
@@ -65,6 +68,10 @@ import java.util.Map;
  * @since 0.0.2
  */
 public class DesignDocument extends com.cloudant.client.org.lightcouch.Document {
+
+    private static final String LANG_QUERY = "query";
+    // Default GSON instance for serializing/deserializing JsonElements of views
+    private static final Gson GSON = new Gson();
 
     private String language;
     private Map<String, MapReduce> views;
@@ -383,9 +390,13 @@ public class DesignDocument extends com.cloudant.client.org.lightcouch.Document 
      * @author Ahmed Yehia
      */
     public static class MapReduce {
-        private String map;
-        private String reduce;
+        private JsonElement map;
+        private JsonElement reduce;
         private String dbcopy;
+
+        public MapReduce() {
+            super();
+        }
 
         /**
          * Get the defined map function.
@@ -393,7 +404,7 @@ public class DesignDocument extends com.cloudant.client.org.lightcouch.Document 
          * @return string of the javascript map function
          */
         public String getMap() {
-            return map;
+            return stringifyMRElement(map);
         }
 
         /**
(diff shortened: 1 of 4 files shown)


rhyshort:
+1

cloudant-client/src/main/java/com/cloudant/client/api/model/DesignDocument.java
```diff
@@ -446,6 +457,20 @@ public String getDbCopy() {
             return dbcopy;
         }
 
+        /**
+         * @param element JSON object or string
+         * @return a string form of the object or the JSON string with leading and trailing "
+         * stripped
+         */
+        private String stringifyMRElement(JsonElement element) {
+            String s = element.toString();
```
emlaver: Wouldn't this throw an exception if the element is `null`? Also, what about renaming the variable `s` to show that the variable holds the string representation of the JSON element?

ricellis: Yeah there should be a `null` check, thanks. I've added it in 198ed8b.
I also reversed the checking to use `isJsonString` instead of `isJsonObject` so that we only attempt to strip quotes if we are sure it is a string, as other JSON elements would not be quoted. Even though it (at least currently) should only be an Object or a String it should be less fragile this way.
@rhyshort FYI in case you wanted to look at this change and change your +1

rhyshort:
Still +1, although I probably should have noticed that

cloudant-client/src/main/java/com/cloudant/client/api/model/DesignDocument.java
ricellis: Renamed in 16aa3f8.

emlaver:
+1

Issue #183, opened 2016-01-15
Add descending option for changes feed


emlaver:
Descending option is missing from the method chaining in `com.cloudant.client.api.Changes` and `com.cloudant.client.org.lightcouch.Changes'.

It would need a method similar to below and additional testing:
```
    public Changes descending(boolean descending) {
        this.databaseHelper.query("descending", descending);
        return this;
    }
```

PR #253, opened 2016-05-10
183 changes desc

ricellis:
## What

Added `descending` option to changes feed.

## How
Removed lightcouch `Changes` and `ChangesResult`.
Moved method implementations into cloudant package for `Changes` and `ChangesResult` classes.
Refactored `Changes` constructor and called it from `Database.changes()`
Fixed doc on `continuousChanges` method.
Added descending method to `Changes` class.
Updated CHANGES.md.

## Testing

Added test for changes `descending=true`.

## Reviewers
reviewer @rhyshort
reviewer @emlaver

## Issues
#183


rhyshort:
+1

cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
```diff
@@ -115,54 +148,104 @@ public void stop() {
...
     public ChangesResult getChanges() {
-        com.cloudant.client.org.lightcouch.ChangesResult couchDbChangesResult = changes
-                .getChanges();
-        ChangesResult changeResult = new ChangesResult(couchDbChangesResult);
-        return changeResult;
+        final URI uri = this.databaseHelper.changesUri("feed", "normal");
+        return client.get(uri, ChangesResult.class);
     }
 
     // Query Params
```
emlaver: Wouldn't having java docs (even if it's one sentence) be helpful for the query parameters below?

ricellis: That was just a direct code move of the duplicate lightcouch methods into a single class, neither of which was documented. I agree they should be documented though so I have added docs in 088dbcc.

emlaver:
+1

PR #254, opened 2016-05-10
235 changes params

### LINKED ISSUE #235: Provide a way to add parameters to _changes requests
version: all
java version: n/a

The `changes` API on the `Database` object does not allow for setting of parameters to use with a filter function, we need to add this capability.

ricellis:
## What

Added custom query parameter for changes

## How

Added `Changes.parameter(String, String)` method for adding a custom query parameter.
Apply the arguments of that method to the changes request.
Updated `CHANGES.md`.

## Testing

Added test for custom parameter `ChangeNotificationsTest#changesCustomParameter` that validates the custom parameter is added to a URL.

## Reviewers

reviewer @rhyshort 

## Issues
Fixes #235

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
@@ -241,6 +241,19 @@ public class Changes {
         return this;
     }
 
+    /**
+     * Add a custom query parameter to the _changes request. Useful for specifying extra parameters
+     * to a filter function for example.
+     *
+     * @param name the name of the query parameter
+     * @param value the value of the query parameter
+     * @return this Changes instance
+     */
+    public Changes parameter(String name, String value) {
+        this.databaseHelper.query(name, value);
+        return this;
+    }
+
     // Helper
 
     /**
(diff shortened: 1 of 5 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
```diff
@@ -241,6 +241,19 @@ public Changes descending(boolean descending) {
         return this;
     }
 
+    /**
+     * Add a custom query parameter to the _changes request. Useful for specifying extra parameters
+     * to a filter function for example.
+     *
+     * @param name the name of the query parameter
+     * @param value the value of the query parameter
+     * @return this Changes instance
```
rhyshort: Add an `@since` tag.

rhyshort:
Other than that small issue +1

cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
ricellis: Added tags for a bunch of things in d0b205b which I'll leave as a separate commit.

Issue #226, opened 2016-03-08
Provide more error information if there is a setPermissions error


ricellis:
- CloudantClient (java-cloudant) version(s) that are affected by this issue.

All versions.

- Java version (including vendor and platform).

N/A

- If you're using the optional okhttp dependency.

N/A

- A small code sample that demonstrates the issue.

`db.setPermissions("new_user", EnumSet.of(Permissions._admin));`

when you don't have enough permission to set this.

The idea is to add any JSON `error` or `response` to the exception thrown.

PR #255, opened 2016-05-11
226 error resp

ricellis:
## What

Ensured error information is included in setPermissions exceptions.

## How

The issue #226 was opened as a result of a PR comment around the throwing of a `CouchDbException` from the `setPermissions` method without including `error` or `reason` information. Investigation of the code paths involved revealed that the branch that threw that exception should have been unreachable anyway because if `ok` was not `true` in the response JSON then a `CouchDbException` would have been thrown already by the `executeToInputStream` method from a non-2XX response code. So the changes here are primarily removing that unnecessary path and adding some tests.

* Simplified `Database.setPermissions()` error handling by removing unreachable exception path.

## Testing

* Added new test `com.cloudant.tests.DatabaseTest#permissionsParsing` to check ok/error states of setPermissions.

## Reviewers

reviewer @emlaver
## Issues
#226

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
@@ -171,19 +171,10 @@ public class Database {
         }
         elem.getAsJsonObject().add(userNameorApikey, jsonPermissions);
 
-        InputStream response = null;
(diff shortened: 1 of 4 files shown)


rhyshort:
+1

emlaver:
+1

PR #260, opened 2016-05-16
258 db create

### LINKED ISSUE #258: NoDocumentException is not raised with non existing and not to be created databases
`com.cloudant.client.api.CloudantClient.database(dbName, create)` method doesn't raise anymore a `com.cloudant.client.org.lightcouch.NoDocumentException` exception. In the chain of calls, it seems to be related to the constructor `com.cloudant.client.org.lightcouch.CouchDatabaseBase.CouchDatabaseBase(CouchDbClient, String, boolean)`, which doesn't raise that exception anymore (compared to java-cloudant version 1.0.0).

In version 1.0.0, class `org.lightcouch.CouchDatabaseBase.CouchDatabaseBase(CouchDbClientBase, String, boolean)` had a `connect()` method which was  raising `NoDocumentException`, now it has been substituted with a `create()` which has a different logic.

- CloudantClient (java-cloudant) version(s) that are affected by this issue: for sure 2.4.2 and 2.4.3 (version 1.0.0 is not affected)
- Java version (including vendor and platform): Oracle JDK 1.7
- If you're using the optional okhttp dependency: no
- A small code sample that demonstrates the issue:

```java
public Database getDB(String dbName, boolean create) {
		try {
			return client.database(dbName, create);
		} catch(NoDocumentException e) {
			// NEVER CALLED!
			log.debug("Database " + dbName + " doesn't exist and won't be created: " + e.getMessage());
			return null;
		}
}
```

A workaround (at the cost of one more GET request) is:
```
public Database getDB(String dbName, boolean create) {
		try {
			Database result client.database(dbName, create);
			result.info();
			return result;
		} catch(NoDocumentException e) {
			// NEVER CALLED!
			log.debug("Database " + dbName + " doesn't exist and won't be created: " + e.getMessage());
			return null;
		}
}
```

ricellis:
## What

Fixed database create behaviour and documentation.

## How

Fixed documentation for `CloudantClient.database` database creation.
Consolidated database create behaviour into a single method.
Removed unnecessary `GET` from `createDB` method that was preventing the documented `PreconditionFailedException` from propagating to`CloudantClient.createDB`.
Added new test for validating the `PreconditionFailedException` is thrown.

## Testing
Added new test for validating the `PreconditionFailedException` is thrown when using `createDB` and the database does not exist.

## Reviewers

reviewer @emlaver
reviewer @rhyshort adding you as you already looked at this anyway

## Issues
Fixes #258

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,11 @@
   the request for example to be used by a filter function.
 - [FIX] `JsonSyntaxException` when deserializing Cloudant query language generated design
   documents into the `DesignDocument` class.
+- [FIX] `PreconditionFailedException` was never thrown when calling `createDB("dbname")` when the
+  database already existed.
+- [FIX] Documentation that suggested calling `database("dbname", false)` would immediately throw a
+  `NoDocumentException` if the database did not exist. The exception is not thrown until the first
+  operation on the `Database` instance.
 
 # 2.4.3 (2016-05-05)
 - [IMPROVED] Reduced the length of the User-Agent header string.
--- a/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -200,12 +200,16 @@ public class CloudantClient {
 
     /**
      * Get a database reference for the database with the specified name.
+     * <P>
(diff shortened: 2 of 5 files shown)


rhyshort:
👍

emlaver:
+1

PR #262, opened 2016-05-18
140 logging

### LINKED ISSUE #140: Logging improvements
There is not much information about logging for this library. We should at the very least document how to enable what logging is available and probably add some additional logging where we think it could be useful.

An example `HttpConnectionInterceptor` for logging HTTP requests and responses would be useful.

ricellis:
## What

Added additional logging

## How

Added a documented example logging.properties file and included it in javadoc.
Added more logging to `HttpConnection` including request/response URL and method at level `FINE`
and request/response headers at level `FINER`.
Added filters `com.cloudant.http.filter.method` and `com.cloudant.http.filter.url` to enable selective logging from `HttpConnection`.
Added a log message prefix for `HttpConnection` so that multiple interleaved requests/responses can be associated.
Added `ClientBuilder` configuration logging.
Added client init config logging.
Updated documentation for logging.

## Testing

Added new `LoggingTests` class to check the new logging works.
Also did some manual testing.

## Issues
Fixes #140

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
@@ -99,7 +99,7 @@ public class CouchDbClient {
             if (maxConns > 0) {
                 log.config("Setting max connections to " + maxConns);
                 //keep connections open for as long as possible, anything over 2.5 minutes will be
-                //longer than the server
+                //longer than the server so we'll use a 3 minute timeout
                 ConnectionPool pool = new ConnectionPool(maxConns, TimeUnit.MINUTES.toMillis(3));
                 factory.getOkHttpClient().setConnectionPool(pool);
             }
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
@@ -76,10 +76,7 @@ public class CouchDbClient {
     private List<HttpConnectionRequestInterceptor> requestInterceptors;
     private List<HttpConnectionResponseInterceptor> responseInterceptors;
 
-    private HttpConnection.HttpUrlConnectionFactory factory =
-            (OkHttpClientHttpUrlConnectionFactory.isOkUsable())
-                    ? new OkHttpClientHttpUrlConnectionFactory()
-                    : new DefaultHttpUrlConnectionFactory();
+    private final HttpConnection.HttpUrlConnectionFactory factory;
 
     CouchDbClient(CouchDbConfig config) {
         final CouchDbProperties props = config.getProperties();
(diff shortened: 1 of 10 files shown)


cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -92,19 +89,24 @@
...
-            //longer than the server
-            ConnectionPool pool = new ConnectionPool(props.getMaxConnections(), TimeUnit.MINUTES
-                    .toMillis(3));
-            factory.getOkHttpClient().setConnectionPool(pool);
+            final int maxConns = props.getMaxConnections();
+            if (maxConns > 0) {
+                log.config("Setting max connections to " + maxConns);
+                //keep connections open for as long as possible, anything over 2.5 minutes will be
+                //longer than the server
+                ConnectionPool pool = new ConnectionPool(maxConns, TimeUnit.MINUTES.toMillis(3));
```
tomblench: Does this match the comment? looks like 3 minutes

cloudant-client/src/test/java/com/cloudant/tests/LoggingTest.java
```diff
@@ -0,0 +1,249 @@
...
+        // Check the messages were the ones we expected
+        assertMessage("PUT .* request", 2);
+        assertMessage("PUT .* response 200 OK", 3);
+    }
+
+    @Test
+    public void clientBuilderLogging() throws Exception {
+        logger = setupLogger(ClientBuilder.class, Level.CONFIG);
+        CloudantClientHelper.newMockWebServerClientBuilder(mockWebServer).build();
+        assertEquals("There should be 5 log entries", 5, handler.logEntries.size());
```
tomblench: Comment should explain what the 5 messages are so we can deal with regressions easily

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -109,6 +111,8 @@
     public HttpUrlConnectionFactory connectionFactory = new DefaultHttpUrlConnectionFactory();
 
     private int numberOfRetries = 10;
+    private int attemptNumber = 1;
```
tomblench: Do we need both attemptNumber and numberOfRetries?

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
ricellis: The point was that 3 > 2.5, but I've made the comment a bit more clear in bd1c0be.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
ricellis: Sort of. `numberOfRetries` is configurable and counts down as attempts are made. If we don't use `attemptNumber` as well then we'd have to store whatever the initial number was to do a subtraction to work out what number attempt we were on. We could iterate based on `attemptNumber`, but we'd still need a field for the max number to try.

As the primary purpose was to allow correlation of requests and responses I could get rid of it by making the log prefix something like hashcode-numberOfRetries so requests would get logged as 1234abcd-10, then 1234abcd-9 etc.

cloudant-client/src/test/java/com/cloudant/tests/LoggingTest.java
ricellis: Added assertions for the messages in de3869b.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: @ricellis yes I think you should go with that proposal of `1234abcd-10, then 1234abcd-9 etc`

tomblench:
+1 once https://github.com/cloudant/java-cloudant/pull/262#discussion_r64049071 done

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
ricellis: Done in 23a19e5.

PR #267, opened 2016-07-04
60008 optional 429

ricellis:
## What

Made 429 backoff optional and configurable.

## Why

Some users would prefer to receive an error earlier than the current 429 replay behaviour.

## How

Added state storage and retrieval to the `HttpConnectionInterceptorContext`.

Refactored `RequestLimitingInterceptor` as `Replay429Interceptor` with constructor options for the number of replays, initial sleep duration and whether to prefer a `Retry-After` header from the server.

Version 2.5.0 behaviour can be enabled by using: `new Replay429Interceptor(10, 250l);`
The new defaults of 3 retries starting at 250 ms can be obtained from a constant `Replay429Interceptor.WITH_DEFAULTS`.
The 429 replays are optionally enabled by the user adding the interceptor to the client e.g. `ClientBuilder.account("example").interceptors(Replay429Interceptor.WITH_DEFAULTS).build();`

Updated documentation.

## Testing

Added `MockWebServer` utils for a constant response dispatcher and a method for getting a 429 `MockResponse`.
Refactored 429 tests to use new utilities.
Added new tests for configuring 429.
Add a test for the global number of retries.
Added an improved `TestTimer` class.


cloudant-http/src/main/java/com/cloudant/http/interceptors/Replay429Interceptor.java
```diff
@@ -0,0 +1,160 @@
...
+
+            // We only want to take action on a 429 response
+            if (code != 429) {
+                return context;
+            }
+
+            // We received a 429
+
+            // Get the counter from the request context state
+            AtomicInteger attemptCounter = context.getState(this, "attempt", AtomicInteger.class);
```
tomblench: Keys like `"attempt"` should be defined as constants somewhere

cloudant-http/src/main/java/com/cloudant/http/interceptors/Replay429Interceptor.java
```diff
@@ -0,0 +1,160 @@
...
+                String retryAfter = (preferRetryAfter) ? urlConnection.getHeaderField
+                        ("Retry-After") : null;
+                if (retryAfter != null) {
+                    // See https://tools.ietf.org/html/rfc6585#section-4
+                    // Whilst not specified for 429 for 3xx or 503 responses the Retry-After header
+                    // is expressed as an integer number of seconds or a date in one of the 3 HTTP
+                    // date formats https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3
+                    // Cloudant servers should give us an integer number of seconds, so don't worry
+                    // about parsing dates for now.
+                    sleepTime = Long.parseLong(retryAfter) * 1000;
```
tomblench: Possibly need to sanity check `Retry-After` - what if it isn't a valid number or it's absurdly large? A naughty server could cause us to hang by making us sleep for a really long time.

cloudant-http/src/main/java/com/cloudant/http/interceptors/Replay429Interceptor.java
```diff
@@ -0,0 +1,160 @@
...
+            int attempt = attemptCounter.getAndIncrement();
+
+            // Check if we have remaining replays
+            if (attempt < numberOfReplays && context.connection.getNumberOfRetriesRemaining() > 0) {
+
+                // If the response includes a Retry-After then that is when we will retry, otherwise
+                // we use a doubling sleep
+                long sleepTime = 0l;
+
+                String retryAfter = (preferRetryAfter) ? urlConnection.getHeaderField
```
tomblench: Style nitpick, but `preferRetryAfter` doesn't need to be in parens

cloudant-http/src/main/java/com/cloudant/http/interceptors/Replay429Interceptor.java
ricellis: Done in 2fc1184.

cloudant-http/src/main/java/com/cloudant/http/interceptors/Replay429Interceptor.java
ricellis: Done in 349cbbd.

cloudant-http/src/main/java/com/cloudant/http/interceptors/Replay429Interceptor.java
ricellis: Done in c5eff39.

tomblench:
+1

emlaver:
LGTM +1

PR #269, opened 2016-07-13
Closed streams and documented the need to close

### LINKED ISSUE #268: Failure to close streams causing okhttp NPE
@stephen-kruger reported intermittently seeing this stack trace:

>```
Exception in thread "OkHttp ConnectionPool" java.lang.NullPointerException
	at com.squareup.okhttp.Address.hashCode(Address.java:195)
	at java.lang.Object.toString(Object.java:144)
	at java.lang.String.valueOf(String.java:1820)
	at com.squareup.okhttp.ConnectionPool.pruneAndGetAllocationCount(ConnectionPool.java:305)
	at com.squareup.okhttp.ConnectionPool.cleanup(ConnectionPool.java:247)
	at com.squareup.okhttp.ConnectionPool$1.run(ConnectionPool.java:94)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1143)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:618)
	at java.lang.Thread.run(Thread.java:785) 
```

    - Reported for 2.5.0
    - N/A
    - Yes
    - Intermittent issue

Clearly this is coming from okhttp, but following the stack we can see that okhttp is trying to [log a warning](https://github.com/square/okhttp/blob/parent-2.7.5/okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java#L305) when it has the NPE. The warning is for a connection being leaked probably because a response body wasn't closed.

It looks like there are a few places where we need to do better at closing response streams.

### PR DESCRIPTION
**Pull request by ricellis**
## What

Added calls to close response streams in places they were missing.

## How

There were a couple of places where calls to `CouchDbUtil` were not
closing streams. `CouchDbUtil.getAsString` was already closing the
streams, so it makes sense if the rest of the util functions that use
streams do the same.

Added comments to remind users to close InputStreams in various places,
where the streams are returned outside of our control.

Moved stream closes into `finally` blocks to make sure they are called.
* Closed stream in a `finally` for `streamToString`
* Routed response calls in `HttpConnection` through fewer lines
  and then moved the remaining `close` into a `finally` block to
  make sure it is called.

## Testing

No new tests added. Existing tests continue to pass.

## Issues

Fixes #268

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
@@ -302,7 +302,8 @@ public class CouchDbClient {
     }
 
     /**
-     * Performs a HTTP GET request.
+     * <p>Performs a HTTP GET request.</p>
+     * <p>The stream <b>must</b> be closed after use.</p>
      *
      * @return Input stream with response
      */
@@ -327,7 +328,8 @@ public class CouchDbClient {
     }
 
     /**
-     * Performs a HTTP HEAD request.
+     * <p>Performs a HTTP HEAD request.</p>
+     * <p>The stream <b>must</b> be closed after use.</p>
      *
      * @return {@link Response}
      */
@@ -337,8 +339,9 @@ public class CouchDbClient {
     }
 
     /**
-     * Performs a HTTP PUT request, saves or updates a document.
-     * This defaults to "application/json" content type.
+     * <p>Performs a HTTP PUT request, saves or updates a document.
+     * This defaults to "application/json" content type.</p>
+     * <p>The stream <b>must</b> be closed after use.</p>
      *
      * @return Input stream with response
      */
@@ -347,7 +350,8 @@ public class CouchDbClient {
     }
 
     /**
-     * Performs a HTTP PUT request with content type, saves or updates a document.
+     * <p>Performs a HTTP PUT request with content type, saves or updates a document.</p>
+     * <p>The stream <b>must</b> be closed after use.</p>
      *
      * @return Input stream with response
      */
@@ -412,7 +416,8 @@ public class CouchDbClient {
     }
 
     /**
-     * Performs a HTTP POST request with JSON request body.
+     * <p>Performs a HTTP POST request with JSON request body.</p>
+     * <p>The stream <b>must</b> be closed after use.</p>
      *
      * @return Input stream with response
      */
@@ -549,7 +554,8 @@ public class CouchDbClient {
     }
 
     /**
-     * Execute the HttpConnection request and return the InputStream if there were no errors.
+     * <p>Execute the HttpConnection request and return the InputStream if there were no errors.</p>
+     * <p>The stream <b>must</b> be closed after use.</p>
      *
      * @param connection the request HttpConnection
      * @return InputStream from the HttpConnection response
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
@@ -30,6 +30,7 @@ import com.google.gson.JsonParser;
 import org.apache.commons.io.IOUtils;
 
 import java.io.Closeable;
+import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by rhyshort** [comment]
+1

File: cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java [hunk 1]
```diff
@@ -303,6 +303,7 @@ Response delete(URI uri) {
 
     /**
      * Performs a HTTP GET request.
+     * <p>The stream <b>must</b> be closed after use.</p>
```
tomblench: The first paragraph for all of these should be enclosed in `<p>`s

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
```diff
@@ -127,12 +127,15 @@ public static String removeExtension(String fileName) {
...
-        s.useDelimiter("\\A");
-        String str = s.hasNext() ? s.next() : null;
-        close(in);
-        s.close();// mdb
-        return str;
+        try {
+            Scanner s = new Scanner(in, "UTF-8");
+            s.useDelimiter("\\A");
+            String str = s.hasNext() ? s.next() : null;
+            s.close();// mdb
```
tomblench: I see this `// mdb` comment was carried over from a previous change, what does it mean? It's rather cryptic

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
```diff
@@ -127,12 +127,15 @@ public static String removeExtension(String fileName) {
...
+            return str;
+        } finally {
+            close(in);
```
tomblench: This applies to the other new `close()`s as well - in some places we do `IOUtils.closeQuietly()` which will swallow the exception. This was probably a convention which built up over time on the `sync-android` library. There are pros and cons about either approach: in the one hand it's generally bad practice to silently swallow exceptions, on the other hand it's not very helpful to throw exceptions out to API users especially when there's little they can do about them.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
tomblench: Actually this might be outside the scope of this PR but I'm confused as to what this method is supposed to do. It appears to use a regex to match all of the text (`\A` matches "The beginning of the input") as a rather obscure way of reading a stream as a string. As long as we are happy to have `IOUtils` as a dependency, I would suggest we use the same one-liner as `HttpConnection.responseAsString`

ricellis: Yes, no idea what the context of mdb was. I agree it is probably slightly out of scope, but I will replace this `Scanner` code.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
ricellis: Fixed in 2399b87

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
ricellis: Fixed in 50d4461.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/CouchDbUtil.java
ricellis: This `close()` method here is a static in `CouchDbUtil` that does call `IOUtils.closeQuietly()`. I agree in some places it would be better to do the raw close (e.g. in `HttpConnection` where we are already can throw a checked `IOException`). That said until we move up the Java level to be able to use try-with-resources I'm reluctant to add all the levels of nesting required to perform the necessary `null` checks and multiple try/finally blocks especially where we are trying to close multiple `Closeable`s.

tomblench:
+1

PR #275, opened 2016-07-25
272 reduceable multi

### LINKED ISSUE #272: Cannot add reduce options when using MultipleRequestBuilder
    * 2.x
    * Any
    * N/A

 ```java
ViewRequestBuilder viewRequestBuilder = db.getViewRequestBuilder("design", "view"); 
ViewMultipleRequest<String, String> viewRequest = viewRequestBuilder
        .newMultipleRequest(Key.Type.STRING, String.class)
        .includeDocs(true)
        .reduce(false)
        .build();   
```

will cause a compilation error 
>error: cannot find symbol            
    .reduce(false)

### PR DESCRIPTION
**Pull request by ricellis**
## What

Added builder methods for `reduce` related parameters when using view multiple requests.

## How

Added `SettableViewParameters.Reduceable` interface to `MultipleRequestBuilder` interface.

## Testing

Added a new test that uses reduce with `MultipleRequestBuilder`.

## Issues

Fixes #272

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# Unreleased
+- [NEW] Enabled `reduce` and other reduce related parameters to be set when using
+  `MultipleRequestBuilder`.
+
 # 2.5.1 (2016-07-19)
 - [IMPROVED] Made the 429 response code backoff optional and configurable. To enable the backoff add
   an instance of a `Replay429Interceptor` with the desired number of retries and initial backoff:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: CHANGES.md [hunk 1]
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+- [FIX] Enabled `reduce` and other reduce related parameters to be set when using
+  `MultipleRequestBuilder`.
```
rhyshort: After reading back over our API and the semver spec knowing that we shipped 2.0 without reduce on the `MultipleRequestBuilder` interface, I think we need to mark it as `IMPROVED` or `NEW` and release it as a minor.  Because it effectively adds new function. I'm not sure how it would apply to someone who is implementing the interface, but I've convinced myself that it isn't an issue, after reading the spec.

ricellis: An implementer would have to add the new methods.

Updated to `[NEW]` in e6f6be9.

rhyshort:
+1

PR #276, opened 2016-07-29
Updated HTML error response test

ricellis:
## What

Updated HTML response test

## How

New versions of Cloudant return JSON instead of HTML for these error cases the test is updated for this by asserting response codes and error message from JSON instead of looking for `<html>` tag.

## Testing

Renamed test from `testNonJsonErrorStream` to `testJsonErrorStreamFromLB`.


rhyshort:
+1

PR #278, opened 2016-08-02
277 leaked conns

### LINKED ISSUE #277: Unconsumed response streams leaking connections
Following on from #268 there are some other places where streams are potentially not consumed which lead to a connection leaks. This could prevent connection re-use creating additional connection creation overhead.

The problem is caused by some test behaviour and some code behaviour..

    * 2.x
    * 1.8
    * Yes

Run the service tests with okhttp and the logs have instances of the message:
>Jul 29, 2016 4:59:43 PM com.squareup.okhttp.ConnectionPool pruneAndGetAllocationCount
WARNING: A connection to <test host> was leaked. Did you forget to close a response body?

ricellis:
## What

Consumed and closed response streams to avoid leaking connections.

## How

Made `CookieInterceptor` consume error streams in 4xx response cases.

Made `DELETE _session` request in `client.shutdown()` consume the response stream.

## Testing

Fixed tests that were not closing connection streams:
* `AttachmentsTest`
* `CloudantClientTests`
* `HttpProxyTest`
* `HttpTest`
* `LoggingTest`
* `ResponseTest`

Fixed `HttpTest.cookieInterceptorURLEncoding` test to use `mockWebServer` field instead of starting and stopping an additional instance.

No more instances of leaked connection message seen in logs.

## Issues

Fixes #277

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2011 lightcouch.org
- * Copyright (c) 2015 IBM Corp. All rights reserved.
+ * Copyright © 2015, 2016 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -34,6 +34,7 @@ import com.cloudant.tests.util.DatabaseResource;
 import com.cloudant.tests.util.Utils;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.IOUtils;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
@@ -105,17 +106,15 @@ public class AttachmentsTest {
                 .attachmentUri(response.getId(), "foo.txt"));
         InputStream in = clientResource.get().executeRequest(conn).responseAsInputStream();
 
-        ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
-        int n;
-        while ((n = in.read()) != -1) {
-            bytesOut.write(n);
+        try {
+
+            ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
+            IOUtils.copy(in, bytesOut);
+            byte[] bytesFromDB = bytesOut.toByteArray();
+            assertArrayEquals(bytesToDB, bytesFromDB);
+        } finally {
+            in.close();
         }
-        bytesOut.flush();
-        in.close();
-
-        byte[] bytesFromDB = bytesOut.toByteArray();
-
-        assertArrayEquals(bytesToDB, bytesFromDB);
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
```diff
@@ -105,17 +105,20 @@ public void attachmentStandalone() throws IOException, URISyntaxException {
...
 
-        ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
-        int n;
-        while ((n = in.read()) != -1) {
-            bytesOut.write(n);
-        }
-        bytesOut.flush();
-        in.close();
+        try {
+            ByteArrayOutputStream bytesOut = new ByteArrayOutputStream();
```
tomblench: I know these are only test code, and this isn't strictly the domain of this PR, but these tests which read and write 1 byte at a time will be _really_ slow.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -136,7 +136,8 @@ public void testUserAgentHeaderIsAddedToRequest() throws Exception {
             //instantiating the client performs a single post request
             CloudantClient client = CloudantClientHelper.newMockWebServerClientBuilder(server)
                     .build();
-            client.executeRequest(createPost(client.getBaseUri(), null, "application/json"));
+            client.executeRequest(createPost(client.getBaseUri(), null, "application/json"))
```
tomblench: These sorts of invocations could trigger findbugs `RV_RETURN_VALUE_IGNORED_INFERRED` warnings.

cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
ricellis: Agreed it is out of scope, but I did it anyway af283b4.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: True, but we don't run findbugs on the tests.
I've added assertions etc for the return values in 32d6fec.

tomblench:
+1

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -74,10 +88,17 @@ public void testWriteToServerOk() throws Exception {
         assertEquals(data.getBytes().length, bis.available());
 
         conn.setRequestBody(bis);
-        conn.execute();
+        HttpConnection response = conn.execute();
+        // Consume response stream
+        String responseStr = response.responseAsString();
+        assertTrue("There should be an ok response: " + responseStr, Pattern.compile("" +
```
brynh: Unnecessary `"" +`

brynh:
+1 once minor comment addressed.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
ricellis: The auto-format kept adding that back in, I'll try again to get rid of it!

PR #279, opened 2016-08-02
Added parameterised testing for both HttpURLConnection types

ricellis:
## What

Always test both HttpURLConnection implementations with a parameterized test.

## How

Parameterised the `HttpTest` class to test using both `HttpURLConnection`
implementations.
Added JMockit to allow a `MockUp` of `OkHttpClientHttpUrlConnectionFactory` so it is possible to force the default `HttpURLConnection` even when okhttp is on the classpath.

## Testing

These are test changes, the tests pass for both implementations.


emlaver:
LGTM +1

PR #288, opened 2016-08-31
Changed CookieInterceptor to use CookieManager

### LINKED ISSUE #273: CookieInterceptor should not try to renew with invalid credentials
    * 2.x
    * Any
    * N/A

```java
CloudantClient client = ClientBuilder.account("example")
    .username("user")
    .password("wrong")
    .build();
client.getAllDbs(); // make any request
```

Using the wrong credentials results in this log output:

>Jul 21, 2016 11:14:31 AM com.cloudant.http.interceptors.CookieInterceptor getCookie
SEVERE: Credentials are incorrect, cookie authentication will not be attempted again by this interceptor object
Jul 21, 2016 11:14:32 AM com.cloudant.http.interceptors.CookieInterceptor getCookie
SEVERE: Credentials are incorrect, cookie authentication will not be attempted again by this interceptor object

Currently the code attempts to `getCookie` as the request is made and when the response is received (for renewal scenarios). In this case since we know the credentials are wrong there is no value trying to get the cookie again in the response.

There may be some value in throwing the exception from the `CookieInterceptor` to indicate the credentials failure earlier, but since the `CookieInterceptor` is always present in the chain allowing it to cause a request to fail earlier could interfere in scenarios where it is merely a bystander to some other authentication interceptor further down the chain so this would need some careful consideration.

### LINKED ISSUE #287: CookieInterceptor -  java.lang.StringIndexOutOfBoundsException
We have our application running on IBM Bluemix Containers and after several hours (4-8hrs) we see a lot of StringIndexOutOfBoundsException.

CloudantClient 2.5.1

java 1.7

Not sure.

[8/19/16 1:11:58:961 UTC] 000000a0 .cloud.liberty.archaius_cloudant.CloudantConfigurationSource E An exception occurred while reading from [CloudantDB:cosmos-svt-cluster_ecmconfig,cosmosci]:
java.lang.StringIndexOutOfBoundsException
    at java.lang.String.substring(String.java:1360)
    at com.cloudant.http.interceptors.CookieInterceptor.extractCookieFromHeaderValue(CookieInterceptor.java:232)
    at com.cloudant.http.interceptors.CookieInterceptor.interceptResponse(CookieInterceptor.java:99)
    at com.cloudant.http.HttpConnection.execute(HttpConnection.java:357)
    at com.cloudant.client.org.lightcouch.CouchDbClient.execute(CouchDbClient.java:475)
    at com.cloudant.client.api.CloudantClient.executeRequest(CloudantClient.java:335)
    at com.cloudant.client.internal.views.ViewRequester.executeRequestWithResponseAsJson(ViewRequester.java:34)
    at com.cloudant.client.internal.views.ViewRequester.getResponseAsJson(ViewRequester.java:28)
    at com.cloudant.client.internal.views.ViewRequestImpl.getResponse(ViewRequestImpl.java:34)
    at com.cloudant.client.internal.views.AllDocsRequestResponse.getResponse(AllDocsRequestResponse.java:44)
    at com.ibm.ecm.cloud.liberty.archaius_cloudant.CloudantConfigurationSource.loadAllProperties(CloudantConfigurationSource.java:82)
    at com.ibm.ecm.cloud.liberty.archaius_cloudant.CloudantConfigurationSource.poll(CloudantConfigurationSource.java:120)
    at com.netflix.config.AbstractPollingScheduler$1.run(AbstractPollingScheduler.java:163)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:309)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1143)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:618)
    at java.lang.Thread.run(Thread.java:785)

ricellis:
## What

Changed `CookieInterceptor` to use `java.net.CookieManager`

## How

The previous `CookieInterceptor` code failed to account for the possibility of multiple `Set-Cookie` headers being present in a response. It also failed to account for the possibility of multiple cookie values within a single header. Finally and most obviously from #287 it failed to check that the string parsing it was doing was safe. To avoid having to write code for all these scenarios it seemed preferable to use the `java.net.CookieManager` instead, which should be spec compliant.

* Removed our existing cookie/header parsing and used an instance of `java.net.CookieManager` instead.
* Refactored methods and error handling to work with `CookieManager` changes.
* Made the `shouldAttemptCookieRequest` atomic as it is potentially accessed by multiple `HttpConnection` instances.

## Testing

Added tests:
* HttpTest#badCredsDisablesCookie - to check that a 401 when requesting a
cookie stops any further attempts.
* HttpTest#cookieAppliedToDifferentURL - to validate that the cookie is
successfully applied to a different request URL.

Updated the `OK_COOKIE` to better match CouchDB.
Fixed assertions impacted by cookie change (optional " not present in a `CookieManager` handled cookie).

## Issues

Fixes #287 
Also fixes #273

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
@@ -13,10 +13,13 @@
  */
 package com.cloudant.tests;
 
+import static org.hamcrest.CoreMatchers.anyOf;
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -43,6 +46,7 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import com.squareup.okhttp.mockwebserver.Dispatcher;
 import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
 import com.squareup.okhttp.mockwebserver.RecordedRequest;
@@ -64,6 +68,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.URI;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.nio.charset.Charset;
@@ -275,7 +280,7 @@ public class HttpTest {
         String response = c.executeRequest(Http.GET(c.getBaseUri())).responseAsString();
         assertTrue("There should be no response body on the mock response", response.isEmpty());
 
-        RecordedRequest r = mockWebServer.takeRequest(10, TimeUnit.SECONDS);
+        RecordedRequest r = MockWebServerResources.takeRequestWithTimeout(mockWebServer);
         String sessionRequestContent = r.getBody().readString(Charset.forName("UTF-8"));
         assertNotNull("The _session request should have non-null content",
                 sessionRequestContent);
@@ -298,16 +303,19 @@ public class HttpTest {
     @Test
     public void cookieRenewal() throws Exception {
         final String hello = "{\"hello\":\"world\"}\r\n";
-        final String renewalCookieValue =
-                "AuthSession=\"RenewCookie_a2ltc3RlYmVsOjUxMzRBQTUzOtiY2_IDUIdsTJEVNEjObAbyhrgz\";";
+        final String authSession = "AuthSession=";
+        final String renewalCookieToken =
+                "RenewCookie_a2ltc3RlYmVsOjUxMzRBQTUzOtiY2_IDUIdsTJEVNEjObAbyhrgz";
+        final String renewalCookieValue = authSession + renewalCookieToken;
+
         // Request sequence
         // _session request to get Cookie
         // GET request -> 200 with a Set-Cookie
-        // _session for new cookie
         // GET replay -> 200
         mockWebServer.enqueue(MockWebServerResources.OK_COOKIE);
         mockWebServer.enqueue(new MockResponse().setResponseCode(200).addHeader("Set-Cookie",
-                String.format(Locale.ENGLISH, "%s;", renewalCookieValue))
+                String.format(Locale.ENGLISH, "%s;", renewalCookieValue
+                        + MockWebServerResources.COOKIE_PROPS))
                 .setBody(hello));
         mockWebServer.enqueue(new MockResponse());
 
@@ -316,27 +324,34 @@ public class HttpTest {
                 .password("b")
                 .build();
 
-
         String response = c.executeRequest(Http.GET(c.getBaseUri())).responseAsString();
         assertEquals("The expected response should be received", hello, response);
+
+        // assert that there were 2 calls
+        assertEquals("The server should have received 2 requests", 2, mockWebServer
+                .getRequestCount());
+
+        assertEquals("The request should have been for /_session", "/_session",
+                MockWebServerResources.takeRequestWithTimeout(mockWebServer).getPath());
+        assertEquals("The request should have been for /", "/",
+                MockWebServerResources.takeRequestWithTimeout(mockWebServer).getPath());
+
         String secondResponse = c.executeRequest(Http.GET(c.getBaseUri())).responseAsString();
         assertTrue("There should be no response body on the mock response" + secondResponse,
                 secondResponse.isEmpty());
-
 
         // also assert that there were 3 calls
         assertEquals("The server should have received 3 requests", 3, mockWebServer
                 .getRequestCount());
 
-        mockWebServer.takeRequest(); // cookie
-        mockWebServer.takeRequest(); // actual get
-
         // this is the request that should have the new cookie.
-        RecordedRequest request = mockWebServer.takeRequest();
+        RecordedRequest request = MockWebServerResources.takeRequestWithTimeout(mockWebServer);
+        assertEquals("The request should have been for path /", "/", request.getPath());
         String headerValue = request.getHeader("Cookie");
-        assertEquals
-                ("AuthSession=\"RenewCookie_a2ltc3RlYmVsOjUxMzRBQTUzOtiY2_IDUIdsTJEVNEjObAbyhrgz" +
-                        "\"", headerValue);
+        // The cookie may or may not have the session id quoted, so check both
+        assertThat("The Cookie header should contain the expected session value", headerValue,
+                anyOf(containsString(renewalCookieValue), containsString(authSession + "\"" +
+                        renewalCookieToken + "\"")));
     }
 
     /**
@@ -574,8 +589,8 @@ public class HttpTest {
 
         // We want the second request
         assertEquals("There should have been two requests", 2, mockWebServer.getRequestCount());
-        mockWebServer.takeRequest();
-        RecordedRequest rr = mockWebServer.takeRequest();
+        MockWebServerResources.takeRequestWithTimeout(mockWebServer);
+        RecordedRequest rr = MockWebServerResources.takeRequestWithTimeout(mockWebServer);
         assertNotNull("The request should have been recorded", rr);
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream((int) rr
                 .getBodySize());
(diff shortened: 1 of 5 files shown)


ricellis:
Reviewers can hold off on this for a little while I investigate the CI hang on Java 1.7.

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -79,86 +84,109 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
         HttpURLConnection connection = context.connection.getConnection();
 
-        if (shouldAttemptCookieRequest) {
-            if (cookie == null) {
-                cookie = getCookie(connection.getURL(), context);
+        // First time we will have no cookies
+        if (cookieManager.getCookieStore().getCookies().isEmpty() && shouldAttemptCookieRequest
+                .get()) {
+            if (!requestCookie(connection.getURL(), context)) {
+                // Requesting a cookie failed, set a flag if we failed so we won't try again
```
tomblench: Log a warning, "Failed to request cookie, won't attempt again"?

tomblench: Never mind, I see there is logging in `requestCookie`

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -79,86 +84,109 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
+                                break;
+                            default:
+                                break;
+                        }
+                    } finally {
+                        IOUtils.closeQuietly(errorStream);
+                    }
+                    if (renewCookie) {
+                        boolean success = requestCookie(connection.getURL(), context);
+                        // Don't resend request, failed to get cookie
```
tomblench: comment doesn't match code immediately below

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -79,86 +84,109 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
+                        if (success) {
+                            context.replayRequest = true;
+                        } else {
+                            // Didn't successfully renew, maybe creds are expired
```
tomblench: "expired" is confusing in the context of this comments because we think of cookies expiring. Do you mean "maybe creds are incorrect"?

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: FIxed in 4e77de7

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: The thought I had in my mind was expired in the sense of having been changed on the server since this is a renewal and they must have been valid previously if we are trying to renew, but I see your point! Fixed in b1ef8a8.

ricellis:
@rhyshort @tomblench the tests are passing on this now, but as well as changing the tests I added some more logging so might be worth another look

rhyshort:
+1

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -316,27 +324,34 @@ public void cookieRenewal() throws Exception {
...
         // also assert that there were 3 calls
         assertEquals("The server should have received 3 requests", 3, mockWebServer
                 .getRequestCount());
 
-        mockWebServer.takeRequest(); // cookie
-        mockWebServer.takeRequest(); // actual get
-
         // this is the request that should have the new cookie.
-        RecordedRequest request = mockWebServer.takeRequest();
+        RecordedRequest request = mockWebServer.takeRequest(10, TimeUnit.SECONDS);
```
tomblench: Can we factor out all of these timeouts to a common variable if the timeout is always going to be the same?

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -864,7 +879,8 @@ public void testHttpConnectionRetries() throws Exception {
     }
 
     /**
-     * Test that an IllegalArgumentException is thrown if an https proxy address is used. SSL proxies
+     * Test that an IllegalArgumentException is thrown if an https proxy address is used. SSL
```
tomblench: Formatting

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -79,86 +84,123 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
+                                    // Set the flag for deserialization
+                                    toThrow.deserialize = true;
+                                    throw toThrow;
+                                } else {
+                                    // Was expired - set boolean to renew cookie
+                                    renewCookie = true;
+                                }
+                                break;
+                            case HttpURLConnection.HTTP_UNAUTHORIZED: //401
+                                IOUtils.toString(errorStream, "UTF-8");
```
tomblench: Calling `toString` and throwing away the result looks odd. Presumably this is to consume the errorStream. Out of interest does findbugs warn about discarding return values?

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -79,86 +84,123 @@ public HttpConnectionInterceptorContext interceptRequest(HttpConnectionIntercept
...
-                // Don't resend request, failed to get cookie
-                if (cookie != null) {
-                    context.replayRequest = true;
                 } else {
-                    context.replayRequest = false;
+                    // Store any cookies provided on the response
+                    storeCookiesFromResponse(connection);
                 }
+            } catch (IOException e) {
+                logger.log(Level.SEVERE, "Failed to get response code from request", e);
```
tomblench: It's possible we could have got here because we failed to read the error stream as well - I don't know if that makes the failure too verbose

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -245,7 +285,33 @@ private boolean sessionHasStarted(InputStream responseStream) throws IOException
...
+        }
+    }
+
+    private String listToSemicolonSeparatedString(List<String> cookieStrings) {
+        // RFC 6265 says multiple cookie pairs should be "; " separated
+        StringBuilder builder = new StringBuilder();
+        int index = 1; // Count from 1 since we will compare to size
+        for (String cookieString : cookieStrings) {
+            builder.append(cookieString);
+            if (index != cookieStrings.size()) {
```
tomblench: I think you meant `index++` here?

Also do we still support java 7? Java 8 has `String.join()`

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
ricellis: Moved into a static method and added constants in 77b429e (sadly `MockWebServer` is final so I couldn't just override `takeRequest` to always avoid tests hanging for ages).

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
ricellis: Fixed in  7b4cc00

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: Yes, I think findbugs normally does, not sure why it didn't for this. In 632cc86 I have moved the string extraction higher up and added some logging for the value which might be useful anyway.

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: I tweaked it in 632cc86

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
ricellis: Well I was counting from 1 so I didn't, but I did mean to put it at the end of the loop, so good spot. Changed to count from zero and used `index++` in the comparison in 37770a3.

ricellis: Oh and yes, we are still supporting Java 7 for Android.

tomblench:
+1

PR #289, opened 2016-09-07
270 dns log

### LINKED ISSUE #270: Warn / log if DNS TTL is long/infinite
When debugging connection issues, it would be useful to know the DNS caching policy. Perhaps we could log the value of `java.security.Security.getProperty("networkaddress.cache.ttl");` when the library is initialised?

ricellis:
## What

Added log warnings for long JVM DNS cache TTLs.

## How

Added warning level log messages if the TTL is too high or can't be checked.
* If permission is denied to read the configuration property log a warning.
* If the property can be read and parsed, but is longer than 30 s log a warning.
* If the value is the default setting and a `SecurityManager` is active then log a warning because the TTL is infinite if unconfigured with a `SecurityManager`.

## Testing

Added tests to validate log output:
dnsNoWarningLessThan30
dnsNoWarning0
dnsWarningForever
dnsNoWarning30
dnsWarning31
dnsWarningPermissionDenied
dnsWarningDefaultWithSecurityManager

## Issues

Fixes #270

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 - [FIX] `java.lang.StringIndexOutOfBoundsException` when trying to parse `Set-Cookie` headers.
 - [FIX] `NullPointerException` in `CookieInterceptor` when no body was present on response.
 - [UPGRADED] Upgraded GSON to 2.7
+- [IMPROVED] Added warning messages for JVM DNS cache configuration settings that could impede
+  client operation during cluster failover.
 
 # 2.5.1 (2016-07-19)
 - [IMPROVED] Made the 429 response code backoff optional and configurable. To enable the backoff add
--- a/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
@@ -20,6 +20,7 @@ import java.net.Authenticator;
 import java.net.MalformedURLException;
 import java.net.PasswordAuthentication;
 import java.net.URL;
+import java.security.Security;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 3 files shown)


cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -53,6 +53,7 @@
 import java.net.Socket;
 import java.net.SocketTimeoutException;
 import java.net.URL;
+import java.security.Security;
```
rhyshort: I can't see a change that requires this to be imported.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -378,4 +379,6 @@ public void testBasicAuth() throws IOException {
         // Test passes if there are no exceptions
         client.getAllDbs();
     }
+
+
```
rhyshort: Why have these two lines been added?

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: I changed IDEs and import cleanup must now be turned off :(

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: I changed IDEs and I suspect a code style issue.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: Fixed in ff73d2e

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: Fixed in ff73d2e

rhyshort:
+1

ricellis:
@willholley are you happy with the log messages:
>WARNING: DNS cache lifetime may be too long. DNS cache lifetimes in excess of 30 seconds may impede client operation during cluster failover.

and

>WARNING: Permission denied to check Java DNS cache TTL. If the cache lifetime is too long cluster failover will be impeded.

?

willholley:
@ricellis the error messages look fine. Ideally we shouldn't need to prescribe a value (though 30 seconds is probably a good one) and the client should respect the DNS TTL that we return (e.g. see http://mxtoolbox.com/SuperTool.aspx?action=a%3awillholley.cloudant.com&run=toolpage). From what I can tell, this varies between DNS entries - sometimes it's 30 seconds and sometimes up to 5 minutes.

ricellis:
Thanks, this is logging only as we can't change the behaviour of the JVM - so 30 s is the barrier chosen for the message.

PR #291, opened 2016-09-08
290 cookie error npe

### LINKED ISSUE #290: NullPointerException in CookieInterceptor.java
When calling Cloudant.contains() method in cloudant-java API, got:

java.lang.NullPointerException
        at com.cloudant.http.interceptors.CookieInterceptor.interceptResponse(CookieInterceptor.java:128)
        at com.cloudant.http.HttpConnection.execute(HttpConnection.java:357)
        at com.cloudant.client.org.lightcouch.CouchDbClient.execute(CouchDbClient.java:470)
        at com.cloudant.client.org.lightcouch.CouchDbClient.executeToInputStream(CouchDbClient.java:560)
        at com.cloudant.client.org.lightcouch.CouchDbClient.head(CouchDbClient.java:336)
        at com.cloudant.client.org.lightcouch.CouchDatabaseBase.contains(CouchDatabaseBase.java:177)
        at com.cloudant.client.api.Database.contains(Database.java:652)
        at com.ibm.lex.services.matches.database.Cloudant.exists(Cloudant.java:49)

cloudant-java 2.5.0, no okhttp dependancy, JDK 1.8 IBM

Most likely it was temporary issue, but NPE is not good way to handle such situation.

ricellis:
## What

Prevent NPE when intercepting response with no body.

## How

* Added a `null` check for the error stream.
* Stopped converting the stream to a string if it was `null`.
* Added `null` check and reversed condition for `errorString.matches` to avoid NPE if the `errorString` was `null`.
* Changed value of  `toThrow.deserialize` to be `false` if the `errorString` was `null` in the case where a non-expiry 403 needs propagating.

## Testing

Added new tests, which failed with NPE before fix and now pass:
* `noErrorStream403`
* `noErrorStream401`

Note that the NPE does not happen when using okhttp, but our parameterized tests exercise both options anyway.

## Issues

Fixes #290

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
@@ -998,14 +998,17 @@ public class HttpTest {
     }
 
     /**
-     * Test that having no body and hence no error stream correctly results in the 403 causing
-     * a CouchDbException with no NPEs.
+     * Test that having no body and hence no error stream on a 403 response correctly results in the
+     * 403 causing a CouchDbException with no NPEs.
      *
      * @throws Exception
      */
     @Test(expected=CouchDbException.class)
     public void noErrorStream403() throws Exception {
+
+        // Respond with a cookie init to the first request to _session
         mockWebServer.enqueue(MockWebServerResources.OK_COOKIE);
+        // Respond to the executeRequest GET of / with a 403 with no body
         mockWebServer.enqueue(new MockResponse().setResponseCode(403));
         CloudantClient c = CloudantClientHelper.newMockWebServerClientBuilder(mockWebServer)
                 .username("a")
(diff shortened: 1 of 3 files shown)


cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -996,4 +996,44 @@ public MockResponse dispatch(RecordedRequest request) throws InterruptedExceptio
...
+                .username("a")
+                .password("b")
+                .build();
+
+        String response = c.executeRequest(Http.GET(c.getBaseUri())).responseAsString();
+        fail("There should be an exception, but received response " + response);
+    }
+
+    /**
+     * Test that having no body and hence no error stream correctly results in a 401 cookie renewal.
```
tomblench: comment says "no body and hence no error stream" but there is clearly a body

tomblench: Actually I think I understand what is going on here but it's not always when the enqueued responses will be streamed, especially because they are all declared at the top. Perhaps some comments around each group of enqueues might help maintainability.

ricellis: Yeah I see the confusion, tried to make it more clear with 1ccb8bf

tomblench:
+1

PR #300, opened 2016-09-26
Remove `LibraryVersion` class.

tomblench:
Use the `META-INF/client.properties` file to derive the main part of the
user agent (the library name) and the version.

Also add support for detecting Android runtimes. This means this code
can be used by sync-android and will also correctly report java-cloudant
usage on Android.


ricellis:
I think this is most of the way there, but as we discussed briefly there is still a bit of an edge case around the UA identifying whether requests going through `cloudant-http` originated from `sync-android` code or `cloudant-client` code in the case where both libraries are running in the same environment.

As it currently stands I don't think we can be sure which of `cloudant-client*.jar!META-INF/client.properties` or `cloudant-sync-datastore-core*.jar!META-INF/client.properties` will be used as it will depend on the classpath ordering.

I think there are two choices that would solve that problem:

1. Replacing the static `HTTP` with an instance initialized from each library such that the appropriate UA (or partial UA) could be passed through.
2. Retain the shared platform UA information in `cloudant-http`, but move the library specific elements into an interceptor in each library and augment the UA.

If we decided that we don't mind not having the calls from each library being distinguishable then we could just use different file names for `client.properties` (or move them in the package hierarchy) and just choose a deterministic order in which to load them, so that we are at least ensuring consistency in what we will see.

ricellis:
This has been superseded by #318

PR #315, opened 2016-11-04
Updated doc and test for CVE-2016-5597

### LINKED ISSUE #314: Update proxy documentation for CVE-2016-5597
The JVM fixes for [CVE-2016-5597](https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2016-5597) disabled by default the possibility of using Basic auth to authenticate with an HTTP proxy and then HTTPS tunnel because a badly written subclass of `java.net.Authenticator` that was providing credentials for both the HTTP proxy and the HTTPS server could expose the HTTPS credentials in plain text to the proxy server.

The credentials supplied for servers in java-cloudant are never set using `Authenticator.setDefault()` so the client code itself is not vulnerable, but we do have documentation about configuring proxy auth with an `Authenticator` and that documentation needs to be updated wrt this vulnerability.

Additionally the test that exercises this path needs to be updated to function on newer JVMs.

    - Any using https tunneling through http proxy with basic auth, doc has only existed since 2.6.0
- Java version (including vendor and platform)
    - From these Java 8 versions and other backports of the fix
        - [Oracle 8u111](http://www.oracle.com/technetwork/java/javase/8u111-relnotes-3124969.html)
        - [IBM 8 SR3 FP20](http://www-01.ibm.com/support/docview.wss?uid=swg1IV90160)
    - Only applies when not using okhttp.
    - See [`HttpTest#proxiedRequest`] (https://github.com/cloudant/java-cloudant/blob/2.6.2/cloudant-client/src/test/java/com/cloudant/tests/HttpProxyTest.java#L226)

ricellis:
## What

Updated doc and test for CVE-2016-5597.

## How

* Updated doc to mention `jdk.http.auth.tunneling.disabledSchemes` property and the CVE.

## Testing

`HttpProxyTest.proxiedRequest`

* Updated test to use `jdk.http.auth.tunneling.disabledSchemes` property.
* Updated test `Authenticator` to only send creds to a proxy.

Test now passes locally on the most recent JVM.

## Issues

Fixes #314

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpProxyTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpProxyTest.java
@@ -112,6 +112,8 @@ public class HttpProxyTest {
     String mockProxyUser = "alpha";
     String mockProxyPass = "alphaPass";
 
+    String defaultDisabledList = "";
+
     /**
      * Enables https on the mock web server receiving our requests if useHttpsServer is true.
      *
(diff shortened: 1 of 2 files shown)


cloudant-client/src/test/java/com/cloudant/tests/HttpProxyTest.java
```diff
@@ -217,6 +225,8 @@ public void resetAuthenticator() {
         if (useProxyAuth && useHttpsServer && !okUsable) {
             Authenticator.setDefault(null);
         }
+        // Reset the disabled schemes property
+        System.setProperty("jdk.http.auth.tunneling.disabledSchemes", defaultDisabledList);
```
tomblench: I don't really understand the purpose of this - you don't ever set `jdk.http.auth.tunneling.disabledSchemes` to any different value in this test, so why is the reset needed?

ricellis: Good spot, I've fixed it up in 30a24ac.
Basically my reset was happening outside of the if block, thus setting the value to `""` which had the effect of setting it for the test that needed it.
Unfortunately it turns out the property is unset for the default so we can't read the value to reset it anyway.

PR #316, opened 2016-11-07
Updated to okhttp3

ricellis:
## What

Updated optional okhttp dependency version to 3.4.2.

## How

* Changed dependency version to 3.4.2 (including mockwebserver).
* Updated package names.
* Fixed 2.x calls that don’t work in 3.x.
    * Used new okhttp client builder.
    * Updated authenticator.
* Fix SSLInterceptor NPE
    * The spec states returning a non-null (possibly empty) array of issuers and whilst a null works on the default it causes a NPE in some implementations. Whilst I can’t find doc that clearly states it behaviour suggests that the empty array is a special case that doesn’t mean trust none, but trust all.
* Updated `CHANGES.md`.

## Testing

* Parameterised SSL tests for both HTTP implementation types to ensure that we get SSL coverage of both implementations and don't regress one or the other.
* Added `HttpFactoryParameterizedTest` to simplify tests that need both `HttpURLConnection` implementations.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Unreleased
+- [UPGRADED] Optional okhttp dependency to version 3.4.2.
 - [FIX] `NullPointerException` when calling `AllDocsResponse.getIdsAndRevs` for a request with
   multiple non-existent keys (IDs).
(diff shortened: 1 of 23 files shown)


rhyshort:
Check the copyrights

ricellis:
Fixed the copyrights in 8f0dcb6

PR #318, opened 2016-11-15
User agent generation

rhyshort:
## What

Change how the user-agent is generated and inserted to enable sync-android and java-cloudant to be used in the same application.

## How

- Created new UserAgentInterceptor, which adds the user agent to the request.
- ClientBuilder always adds the same UserAgentInterceptor for all clients.
- Deleted `LibraryVersion` class.

## Testing

No new tests, tests updated to check the interceptor for the UA string rather than the `LibraryVersion` class.

## Issues

https://github.com/cloudant/sync-android/issues/362


cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -106,6 +111,38 @@
  */
 public class ClientBuilder {
 
+
+    private static final UserAgentInterceptor USER_AGENT_INTERCEPTOR;
+
+    //init the string, based on a properties file or fallback to some defaults
+    static {
```
ricellis: I think it would be nicer to keep this block with the `UserAgentInterceptor`. We could accept a prefix string in the constructor (potentially with another constructor or helper method for loading the properties).
I guess that static method means reading the properties fewer times, but we will only new the interceptor up once per client anyway.
Then we can reuse more code in sync-android.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -223,6 +260,9 @@ public CloudantClient build() {
         //Build properties and couchdb client
         CouchDbProperties props = new CouchDbProperties(url);
 
+        props.addRequestInterceptors(USER_AGENT_INTERCEPTOR);
```
ricellis: we should just be able to new the interceptor up here and get rid of the static.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -114,8 +114,19 @@ public void cookieTest() {
      * Assert that the User-Agent header is of the expected form.
      */
     @Test
-    public void testUserAgentHeaderString() {
-        String userAgentHeader = new LibraryVersion().getUserAgentString();
+    public void testUserAgentHeaderString() throws Exception {
+
+        ClientBuilder builder = ClientBuilder.account("test");
+        Class clazz = builder.getClass();
+        Field interceptor = clazz.getDeclaredField("USER_AGENT_INTERCEPTOR");
```
ricellis: I think if we move the logic into the `UserAgentInterceptor` we could new up using the prefix string and avoid all the reflection.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -268,7 +255,9 @@ public HttpConnection execute() throws IOException {
         while (retry && numberOfRetries-- > 0) {
             connection = connectionFactory.openConnection(url);
 
-            connection.setRequestProperty("User-Agent", USER_AGENT);
+            // Set the default user-agent, consuming libraries are expected to set the UA via an
+            // interceptor, see UserAgentInterceptor
+            connection.setRequestProperty("User-Agent", DEFAULT_USER_AGENT);
```
ricellis: I think our libraries should always add the interceptor so I'm not sure it is worth adding this default header as it is an extra call on every single request (although I agree it could be nice in the case things went wrong with our library interceptors but we should have tests covering that).

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
rhyshort: Removed from builder in 821ea4a

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
rhyshort: Wouldn't it be better to just pass the same interceptor around? It doesn't mutate anything internally, so I don't think threading is an issue.

ricellis: I was thinking there was no point in storing it, but I see your point that if there are multiple clients built we need not have multiple copies.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -114,8 +114,9 @@ public void cookieTest() {
      * Assert that the User-Agent header is of the expected form.
      */
     @Test
-    public void testUserAgentHeaderString() {
-        String userAgentHeader = new LibraryVersion().getUserAgentString();
+    public void testUserAgentHeaderString() throws Exception {
+
+        String userAgentHeader = new UserAgentInterceptor(UserAgentInterceptor.class.getClassLoader(), "META-INF/client.properties").getUserAgent();
```
ricellis: I think this could use a comment to point out that the file isn't actually present on the test classloader so it is validating the "unknown" case.
As discussed a second test case that uses a classloader with some test file would be valuable.

cloudant-http/src/main/java/com/cloudant/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Loads the properties file using the classloader provided. Creating a string from the properties
+     * "user.agent.name" and "user.agent.version".
+     * @param loader The class loader to use to load the resource.
+     * @param filename The name of the file to load.
+     * @return A string that represents the first part of the UA string eg java-cloudant/2.6.1
+     */
+    protected static String loadUA(ClassLoader loader, String filename){
+        String ua = "java-cloudant";
```
ricellis: I know we should load the properties correctly and thus never really see the unknown case, but this will mean `java-cloudant/unknown` for a sync-android that didn't load properties.
Previously your unknown case had `cloudant-http` do you think that is preferable?

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
ricellis: Also didn't we decide to fully qualify the properties file names?

cloudant-http/src/main/java/com/cloudant/http/interceptors/UserAgentInterceptor.java
rhyshort: changed to `cloudant-http` in fafb83e

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -105,17 +110,48 @@ public void cookieTest() {
     }
 
     // java-cloudant/n.n.n or java-cloudant/unknown followed by 4 groups of /anything
-    private final String userAgentRegex = "java-cloudant/(?:(?:\\d+.\\d+.\\d+)|(?:unknown))" +
+    private final String userAgentRegex = "java-cloudant/(?:(?:\\d+.\\d+.\\d+))" +
             "(?:/{1}[^/]+){4}";
+
+    private final String userAgentUnknownRegex = "java-cloudant/(?:(?:unknown))" +
```
brynh: This should be `cloudant-http` instead of `java-cloudant` if we can't retrieve the user agent shouldn't it?

rhyshort: yes, fixed in 484576b

ricellis:
Don't forget to change the properties file to a FQN before squashing/merging.

rhyshort:
@ricellis change the properties file name, can you double check I got all the places?

ricellis:
@rhyshort:

1. You should definitely update the comment on line 24 of the build.gradle.

1. You could optionally update a few other things, but they aren't essential:

    1. The tests reference `client.properties`
        * com.cloudant.tests.CloudantClientTests#testUserAgentHeaderString (line 130)
        * com.cloudant.tests.CloudantClientTests#testUserAgentHeaderStringFromFile (lines 145 and 154)
    1. You could optionally update the defaults in lines 101/102 of build.gradle, but we override it anyway.

rhyshort:
done in 7086e5d

PR #319, opened 2016-11-16
Jenkinsfile

ricellis:
## What

Added Jenkinsfile for pipeline builds

## How

* Added Jenkinsfile.
* Added publishing for master branch:
    * Added VERSION file.
    * Added upload credentials and command.
    * Added git tagging for release builds.
    * Added VERSION reading to build.gradle.
* Updated build.gradle to enforce encoding options since some of the new Jenkins executors are not UTF-8 by default.

## Testing

OpenJDK based Jenkins executors need the disabled schemes property set before the test runs. With that fix in place the tests run cleanly, there are no new tests.


Jenkinsfile
```diff
@@ -0,0 +1,95 @@
...
+                        sh "git push origin ${version}"
+                    }
+                } finally {
+                    sh "git config --unset credential.username"
+                    sh "git config --unset credential.helper"
+                }
+            }
+        }
+    }
+}
```
rhyshort: add new line before EOF.

PR #321, opened 2016-11-16
Preserve path for cookie authentication.

### LINKED ISSUE #320: CookieInterceptor doesn't work with server that uses URL path
Following on from issue #305 the `CookieInterceptor` also needs updating to handle paths on the end of a base URL pointing to the database server.

Things to fix:

1. Update `CookieInterceptor` to take a base URL.
1. Add a test to make sure the cookie handling works with a URL that includes a path
1. Fix the error message so that the formatting is correct.

    * <= 2.6.2
    * Any
    * N/A
```java
CloudantClient client = ClientBuilder.url("http://myhost.example.com/somepath/")
        .username("user")
        .password("pass")
        .build();
client.serverVersion();
```
>Failed to get cookie from server, response code %s, cookie authentication will not be attempted again

rhyshort:
## What

Preserve path for cookie authentication.

## How

Modify the constructor for `CookieInterceptor` to take the URL that the client was constructed with
to use as the base URL. 

## Testing

New test `testCookieAuthWithPath` in `HTTPTest`

## Issues

Fixes #320

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
@@ -15,6 +15,7 @@ package com.cloudant.tests;
 
 import static org.hamcrest.CoreMatchers.anyOf;
 import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
@@ -23,7 +24,9 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import com.cloudant.client.api.ClientBuilder;
 import com.cloudant.client.api.CloudantClient;
+import com.cloudant.client.api.Database;
 import com.cloudant.client.org.lightcouch.CouchDbException;
 import com.cloudant.client.org.lightcouch.TooManyRequestsException;
 import com.cloudant.http.Http;
@@ -154,7 +157,7 @@ public class HttpTest extends HttpFactoryParameterizedTest {
     @Category(RequiresCloudant.class)
     public void testCookieAuthWithoutRetry() throws IOException {
         CookieInterceptor interceptor = new CookieInterceptor(CloudantClientHelper.COUCH_USERNAME,
-                CloudantClientHelper.COUCH_PASSWORD);
+                CloudantClientHelper.COUCH_PASSWORD, clientResource.get().getBaseUri().toString());
 
         HttpConnection conn = new HttpConnection("POST", dbResource.get().getDBUri().toURL(),
                 "application/json");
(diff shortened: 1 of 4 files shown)


cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -59,23 +59,28 @@
...
         try {
+            this.sessionURL = new URL(String.format("%s/_session", baseURL));
             username = URLEncoder.encode(username, "UTF-8");
             password = URLEncoder.encode(password, "UTF-8");
             this.sessionRequestBody = String.format("name=%s&password=%s", username, password)
                     .getBytes("UTF-8");
         } catch (UnsupportedEncodingException e) {
             //all JVMs should support UTF-8, so this should not happen
             throw new RuntimeException(e);
+        } catch (MalformedURLException e) {
```
ricellis: We should move the existing logging for this to here and remove the other `MalformedURLException` catch.

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
```diff
@@ -59,23 +59,28 @@
...
+            // this should be a valid URL since the builder is passing it in
+            throw new RuntimeException(e);
```
ricellis: I don't think we should throw here if we didn't before.

cloudant-http/src/main/java/com/cloudant/http/interceptors/CookieInterceptor.java
rhyshort: removed in d4ded9b

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -154,7 +157,7 @@ public void testReadBeforeExecute() throws Exception {
     @Category(RequiresCloudant.class)
     public void testCookieAuthWithoutRetry() throws IOException {
         CookieInterceptor interceptor = new CookieInterceptor(CloudantClientHelper.COUCH_USERNAME,
-                CloudantClientHelper.COUCH_PASSWORD);
+                CloudantClientHelper.COUCH_PASSWORD, dbResource.get().getDBUri().toString());
```
ricellis: This one should be `clientResource` since it is the baseURI that is needed.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -186,6 +189,23 @@ public void testCookieAuthWithoutRetry() throws IOException {
         }
     }
 
+    @Test
+    public void testCookieAuthWithPath() throws Exception {
+        MockWebServer mockWebServer = new MockWebServer();
+        mockWebServer.enqueue(new MockResponse().setResponseCode(200).setBody("{\"ok\":true," +
```
ricellis: I think we have a pre-canned OK cookie response in `MockWebServerResources.OK_COOKIE`

rhyshort: done in 7206f74

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
rhyshort: done in 7206f74

PR #324, opened 2016-12-09
Implement bluemix method to create client from VCAP services env

### LINKED ISSUE #323: Implement method to create Cloudant client from VCAP services environment variable
Add an additional method to parse a `VCAP_SERVICES` environment variable allowing easy binding of a service instance to a Cloud Foundry application.

Example `VCAP_SERVICES` variable is JSON:
```json
{
    "cloudantNoSQLDB": [
        {
            "credentials": {
                "username": "example",
                "password": "xxxxxxx",
                "host": "example.cloudant.com",
                "port": 443,
                "url": "https://example:xxxxxxx@example.cloudant.com"
            },
            "syslog_drain_url": null,
            "label": "cloudantNoSQLDB",
            "provider": null,
            "plan": "Lite",
            "name": "Cloudant NoSQL DB"
        }
    ]
}
```

smithsz:
## What

Implement bluemix method that creates a Cloudant client from a VCAP_SERVICES environment variable. This allows easy binding of a service instance to a Cloud Foundry application.

## How

Parse the VCAP_SERVICES environment variable as JSON for necessary configuration required to create a Cloudant client.

## Testing

Includes additional unit tests.

## Issues

Fixes #323.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
@@ -16,6 +16,7 @@ package com.cloudant.client.api;
 
 import com.cloudant.client.api.views.Key;
 import com.cloudant.client.internal.util.DeserializationTypes;
+import com.cloudant.client.internal.util.CloudFoundryServices;
 import com.cloudant.client.internal.util.IndexDeserializer;
 import com.cloudant.client.internal.util.SecurityDeserializer;
 import com.cloudant.client.internal.util.ShardDeserializer;
@@ -29,7 +30,9 @@ import com.cloudant.http.interceptors.ProxyAuthInterceptor;
 import com.cloudant.http.interceptors.SSLCustomizerInterceptor;
 import com.cloudant.http.interceptors.TimeoutCustomizationInterceptor;
 import com.cloudant.http.internal.interceptors.UserAgentInterceptor;
+import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonParseException;
 
 import java.net.Authenticator;
 import java.net.MalformedURLException;
@@ -607,4 +610,81 @@ public class ClientBuilder {
         return this;
     }
 
+    /**
+     * Sets Cloudant client credentials by inspecting the VCAP_SERVICES Cloud Foundry environment
+     * variable. The VCAP_SERVICES variable contains connection information to access a Bluemix
+     * service instance.
+     * <P>
+     * Note: Only a single Cloudant service must exist in VCAP_SERVICES. If there are multiple bound
+     * Cloudant service instances then use {@link #bluemix(String)} to select an instance by name.
+     * </P>
+     *
+     * @return a new ClientBuilder for the account
+     */
+    public static ClientBuilder bluemix() {
+        return ClientBuilder.bluemix(null);
+    }
+
+    /**
+     * Sets Cloudant client credentials by inspecting the VCAP_SERVICES Cloud Foundry environment
+     * variable. The VCAP_SERVICES variable contains connection information to access a Bluemix
+     * service instance.
+     * <P>
+     * Note: Specifying a service name is only required when multiple Cloudant services are
+     * available. If there is only a single bound Cloudant service instance then {@link #bluemix()}
+     * can also be used.
+     * </P>
+     *
+     * @param serviceName name of Bluemix service instance or {@code null} to try to use the only
+     *                    available Cloudant service
+     * @return a new ClientBuilder for the account
+     *
+     * @throws IllegalArgumentException if the Cloudant service instance is not available
+     * @throws IllegalStateException    if the VCAP_SERVICES environment variable is invalid or missing
+     */
+    public static ClientBuilder bluemix(String serviceName) {
+        Gson gson = new GsonBuilder().create();
+
+        String vcap_env = System.getenv("VCAP_SERVICES");
+        if (vcap_env == null) {
+            throw new IllegalStateException("VCAP_SERVICES is not present.");
+        }
+
+        CloudFoundryServices services;
+        try {
+            services = gson.fromJson(vcap_env, CloudFoundryServices.class);
+        } catch (JsonParseException e){
+            throw new IllegalStateException("VCAP_SERVICES is not valid JSON.", e);
+        }
+
+        if (services.cloudantNoSQLDB == null || services.cloudantNoSQLDB.size() == 0) {
+            throw new IllegalStateException("No Cloudant services present.");
+        }
+
+        if (serviceName == null) {
+            if (services.cloudantNoSQLDB.size() == 1) {
+                CloudFoundryServices.CloudFoundryService service = services.cloudantNoSQLDB.get(0);
+                if (service.credentials == null || service.credentials.url == null) {
+                    throw new IllegalStateException("Invalid VCAP_SERVICES services.");
+                }
+                return ClientBuilder.url(service.credentials.url);
+            } else {
+                throw new IllegalArgumentException("Multiple Cloudant services present. " +
+                        "A service name must be specified.");
+            }
+        }
+
+        for (CloudFoundryServices.CloudFoundryService service : services.cloudantNoSQLDB) {
+            if (serviceName.equals(service.name)) {
+                if (service.credentials == null || service.credentials.url == null) {
+                    throw new IllegalStateException("Invalid VCAP_SERVICES services.");
+                }
+                return ClientBuilder.url(service.credentials.url);
+            }
+        }
+
+        throw new IllegalArgumentException(
+                String.format("Cloudant service matching name %s not found.", serviceName)
+        );
+    }
 }
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/util/CloudFoundryServices.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/util/CloudFoundryServices.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright © 2016 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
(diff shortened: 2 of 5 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -607,4 +610,62 @@ public ClientBuilder readTimeout(long readTimeout, TimeUnit readTimeoutUnit) {
...
+     *
+     * @param serviceName name of Bluemix service instance
+     * @return a new ClientBuilder for the account
+     */
+    public static ClientBuilder bluemix(String serviceName) {
+        Gson gson = new GsonBuilder().create();
+
+        String vcap_env = System.getenv("VCAP_SERVICES");
+        if (vcap_env == null) {
+            throw new IllegalArgumentException("VCAP_SERVICES is not present.");
```
ricellis: I'm in two minds about these first three `IllegalArgumentException`s. The javadoc for `IllegalArgumentException` says:
>Thrown to indicate that a method has been passed an illegal or inappropriate argument.

It seems slightly weird to throw it when the argument is correct, but the environment is wrong. That said it could be argued that the method argument is inappropriate when the environment is not correct. However, in my mind `IllegalArgumentException` says I can change the argument and not get the exception, but in this case (and the case of invalid JSON or no Cloudant services bound) that wouldn't be the case.

I'd appreciate other's thoughts on this, but perhaps these first three should really be an `IllegalStateException`.

I'm happy with the last two being `IllegalArgumentException` because changing the argument used could resolve the exception.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -607,4 +610,62 @@ public ClientBuilder readTimeout(long readTimeout, TimeUnit readTimeoutUnit) {
...
+    /**
+     * Sets Cloudant client credentials by inspecting the VCAP_SERVICES Cloud Foundry environment
+     * variable. The VCAP_SERVICES variable contains connection information to access a Bluemix
+     * service instance.
+     * <P>
+     * Note: Specifying a service name is only required when multiple Cloudant services are
+     * available.
+     * </P>
+     *
+     * @param serviceName name of Bluemix service instance
```
ricellis: > `or {@code null} to try to use the only available Cloudant service.`

cloudant-client/src/main/java/com/cloudant/client/internal/util/CloudFoundryServices.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2016 IBM Corp. All rights reserved.
```
ricellis: This needs to be (C) or ©

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
+/*
+ * Copyright (c) 2016 IBM Corp. All rights reserved.
```
ricellis: Should be (C) or ©, probably better update your new file template :)

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+        new MockUp<System>() {
+            @Mock
+            public String getenv(final String string) {
+                return null;
+            }
+        };
+        ClientBuilder.bluemix();
+    }
+
+    @Test(expected = NullPointerException.class)
```
ricellis: As per findbugs, if the VCAP JSON has no `credentials` this could NPE and it appears this test catches that case. I think it should probably be `IllegalStateException` again.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
smithsz: Rhys and I talked about using `IllegalStateException` for these. It does make more sense I think. I'll update.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
smithsz: Will add.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -607,4 +610,62 @@ public ClientBuilder readTimeout(long readTimeout, TimeUnit readTimeoutUnit) {
         return this;
     }
 
+    /**
+     * Sets Cloudant client credentials by inspecting the VCAP_SERVICES Cloud Foundry environment
+     * variable. The VCAP_SERVICES variable contains connection information to access a Bluemix
+     * service instance.
+     *
```
smithsz: Agreed, will amend.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
smithsz: Sure, will update.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -607,4 +610,76 @@ public ClientBuilder readTimeout(long readTimeout, TimeUnit readTimeoutUnit) {
         return this;
     }
 
+    /**
+     * Sets Cloudant client credentials by inspecting the VCAP_SERVICES Cloud Foundry environment
+     * variable. The VCAP_SERVICES variable contains connection information to access a Bluemix
+     * service instance.
+     * <P>
+     * Note: Only a single Cloudant service must exist in VCAP_SERVICES.
```
rhyshort: I think we should call out that an exception is going to be thrown if there is more than one cloudant service in the VCAP. So We'd say something like:

> @throws IllegalArgumentException If there is more than one cloudant service present in the VCAP_SERVICES.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.tests;
+
+import com.cloudant.client.api.ClientBuilder;
+import com.google.gson.GsonBuilder;
+import mockit.*;
```
rhyshort: As part of our code style, we don't use wildcard imports.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+import org.junit.*;
```
rhyshort: Another wildcard import.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
ricellis: I think we should explicitly point users at the alternative method. Maybe append to the note something like:
> `If there are multiple bound Cloudant service instances then use {@link #bluemix(String)} to select an instance by name.`

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+            @Mock
+            public String getenv(final String string) {
+                return new CloudFoundryServiceTest.VCAPGenerator().toJson();
+            }
+        };
+        ClientBuilder.bluemix();
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void vcapSingleServiceMultipleServiceWithNoNameSpecified() {
```
ricellis: I think this test name is a bit wrong.
Shouldn't it be: `vcapMultipleServiceWithNoNameSpecified`?

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+        new MockUp<System>() {
+            @Mock
+            public String getenv(final String string) {
+                VCAPGenerator vcap = new CloudFoundryServiceTest.VCAPGenerator();
+                vcap.createNewServiceWithoutName(CloudantClientHelper.SERVER_URI_WITH_USER_INFO,
+                        CloudantClientHelper.COUCH_USERNAME, CloudantClientHelper.COUCH_PASSWORD);
+                vcap.createNewServiceWithoutName(CloudantClientHelper.SERVER_URI_WITH_USER_INFO,
```
ricellis: This seems to be duplicated.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+                        CloudantClientHelper.COUCH_USERNAME, CloudantClientHelper.COUCH_PASSWORD);
+                return vcap.toJson();
+            }
+        };
+        ClientBuilder.bluemix();
+    }
+
```
ricellis: I think we could use another test for the other `IllegalArgumentException` case something like:
`vcapMultiServiceMissingNamedService`

smithsz: added.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
smithsz: Each `vcap.createNewServiceWithoutName` call adds a new service to the fake `vcap` so I need to repeat it for the multiple service tests.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
smithsz: Yep, updated.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
smithsz: Added.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
ricellis: ok

ricellis:
@smithsz looks good, but it will need an entry added in the `CHANGES.md` before squash/merge.

PR #332, opened 2017-01-10
Improve documentation for design document synchronization

### LINKED ISSUE #307: Design document synchronization doesn't work
Cloudant Client: 2.5.0 onwards. The latest version also has the same issue
Java Version: 1.7
OKHTTP: No
Code Sample: 

```
ClientBuilder builder = ClientBuilder.account("ssss").username("ssss").password("sss");
CloudantClient client = builder.build();
Database db = client.database("ssss", true);
DesignDocumentManager designDocManager = db.getDesignDocumentManager();
InputStream designDocStream = 
                this.getClassLoader().getResourceAsStream(designDocFileName.toString());
            DesignDocument designDocument = 
                gson.fromJson(
                    new InputStreamReader(designDocStream, StandardCharsets.UTF_8), DesignDocument.class);
            designDocManager.put(designDocument);
```

Problem: 
Irrespective of whether the design document with the same content exists or not, the design document is always updated with a new revision of the document. When the Cloudant database is huge, this change in the design document is causing the database to re-index causing database to fail during the indexing time. The issue is that the Document comparison is expecting a revision even before the revision is set in the code. I will send out a PR - please verify and let me know.

Try with a simple design document and the problem is reproducible.

smithsz:
## What
Document that the _id, _rev and full content must match for local &
remote documents to be equal. If the local & remote documents are not
equal then the local document _will_ be saved to the database.

## How
Documentation update.

## Testing
Nope.

## Issues
Fixes #307.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/DesignDocumentManager.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/DesignDocumentManager.java
@@ -88,18 +88,20 @@ public class DesignDocumentManager {
 
     /**
      * Synchronizes a design document to the Database.
-     * <p>This method will first try to find a document in the database with the same id
+     * <p>This method will first try to find a document in the database with the same {@code _id}
      * as the given document, if it is not found then the given document will be saved to the
-     * database.
-     * </p>
+     * database.</p>
      * <p>If the document was found in the database, it will be compared with the given document
-     * using {@code equals()}. If both documents are not equal, then the given document will be
-     * saved to the database and updates the existing document.
-     * </p>
-     * <p>
-     * If the design document's ID is not prefixed with _design, then the _design prefix will be
-     * added.
-     * </p>
+     * using {@link com.cloudant.client.api.model.DesignDocument#equals(Object)}. Both the
+     * {@code _rev} value and content must match for the given documents to be considered equal.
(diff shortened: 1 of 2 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/DesignDocumentManager.java
```diff
@@ -88,18 +88,20 @@
...
+     * database.</p>
      * <p>If the document was found in the database, it will be compared with the given document
-     * using {@code equals()}. If both documents are not equal, then the given document will be
-     * saved to the database and updates the existing document.
-     * </p>
-     * <p>
-     * If the design document's ID is not prefixed with _design, then the _design prefix will be
-     * added.
-     * </p>
+     * using {@link com.cloudant.client.api.model.DesignDocument#equals(Object)}. Both the
```
tomblench: We should probably provide some documentation for `DesignDocument#equals` even though it only does what you'd expect it to do.

You could use something like this, shamelessly stolen and adapted from `java.lang.String`:

"Compares this design document to the specified object. The result is true if and only if the argument is not null and is a DesignDocument with the same _id, _rev, and content"

smithsz: Fixed in d20b8f8a36d26fbb27de7de076c8ef37237ae7d3.

PR #336, opened 2017-01-16
Add a method to GET standalone attachments

### LINKED ISSUE #331: Add a method to GET standalone attachments
Currently to `GET` the `InputStream` for a standalone attachment requires using `executeRequest` code like that below. It would be nice to have an API method that made this a bit easier and more compact to do.

    - 2.x
    - Any
    - N/A

```java
InputStream in;
try {
   HttpConnection conn = Http.GET(new URL(db.getDBUri() + "/myDoc" + "/myAttachment"));
   in = client.executeRequest(conn).responseAsInputStream();
} catch (CouchException e) {
   // handle exception
} catch (MalformedURLException e){
   // handle exception
} finally {
   in.close();
}
```

smithsz:
## What
Add a method to `GET` standalone attachments.

## How
Reads an attachment from the database, returning an `InputStream`.

## Testing
Updates unit tests.

## Issues
Fixes #331.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
@@ -944,6 +944,33 @@ public class Database {
     }
 
     /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentName the attachment name
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    public InputStream getAttachment(String docId, String attachmentName) {
+        return getAttachment(docId, attachmentName, null);
+    }
+
+    /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentName the attachment name
+     * @param revId the document revision id or {@code null}
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    public InputStream getAttachment(String docId, String attachmentName, String revId) {
+        return db.getAttachment(docId, attachmentName, revId);
+    }
+
+    /**
      * Creates an attachment from the specified InputStream and a new document with a generated
      * document ID.
      * <P>Example usage:</P>
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
@@ -299,6 +299,39 @@ public abstract class CouchDatabaseBase {
     }
 
     /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentName the attachment name
+     * @param revId the document revision id or {@code null}
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    public InputStream getAttachment(String docId, String attachmentName, String revId) {
+        final URI uri = new DatabaseURIHelper(dbUri).attachmentUri(docId, revId, attachmentName);
+        return getAttachment(uri);
+    }
+
+    /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param uri the attachment URI
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    private InputStream getAttachment(URI uri) {
+        HttpConnection connection = Http.GET(uri);
+        couchDbClient.execute(connection);
+        try {
+            return connection.responseAsInputStream();
+        } catch (IOException e) {
+            throw new CouchDbException("Error retrieving response input stream.", e);
+        }
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -299,6 +299,53 @@ public Response remove(String id, String rev) {
...
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param uri the attachment URI
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    private InputStream getAttachment(URI uri) {
+        HttpConnection connection = Http.GET(uri);
+        couchDbClient.execute(connection);
+        try {
+            return connection.responseAsInputStream();
```
smithsz: I was wondering if we should return some new `AttachmentInputStream` type here instead that would also hold attachment metadata (such as id, content-type, content-length, etc)?

rhyshort: I can see the argument for the `content-type` and `content-length` but not so much around the others

ricellis: The `Content-Type` header in particular provides information that could potentially be essential for the attachment stream to be handled correctly by an application, but that information is available from the document itself.

ricellis: That said it would be nice not to have to retrieve that information separately, but if you haven't retrieved that info then you wouldn't know the name of the attachment to retrieve (unless you already knew the name because you uploaded it, in which case you could already also know the content-type).

rhyshort: it's probably best that we just return the InputStream, we can always reconsider and introduce something like `AttachmentInputStream` as a subclass of `InputStream` if it turns out there is demand for the meta information being available on the attachment request.

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -944,6 +944,33 @@ public boolean contains(String id) {
...
     /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentId the attachment name
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    public InputStream getAttachment(String docId, String attachmentId) {
```
rhyshort: Other libraries like `SwiftCloudant` refer to the `attachmentId` as `attachmentName` perhaps we should unify around that name?

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -299,6 +299,53 @@ public Response remove(String id, String rev) {
...
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentId the attachment id
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    public InputStream getAttachment(String docId, String attachmentId) {
+        final URI uri = new DatabaseURIHelper(dbUri).attachmentUri(docId, attachmentId);
```
rhyshort: I think this should call down into the three arg version and default `revId` to `null` or just have the three arg version and have the caller set the `revId` to `null`.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -299,6 +299,53 @@ public Response remove(String id, String rev) {
...
+    }
+
+    /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentId the attachment name
+     * @param revId the document revision id
```
rhyshort: state that rev can be `null`.

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
rhyshort: Depending on the decision for the underlying lightcouch Database, this should call into `getAttachment(String, String, String)`.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -299,6 +299,52 @@ public Response remove(String id, String rev) {
...
     /**
+     * Reads an attachment from the database.
+     *
+     * The stream must be closed after usage, otherwise http connection leaks will occur.
+     *
+     * @param docId the document id
+     * @param attachmentName the attachment id
+     * @return the attachment in the form of an {@code InputStream}.
+     */
+    public InputStream getAttachment(String docId, String attachmentName) {
```
ricellis: I think this method can be deleted now that everything points to the 3-arg version.

smithsz: Fixed in 4822347.

ricellis:
oh also needs an update to `CHANGES.md` before merging.

PR #337, opened 2017-01-17
Include URL in the CouchDbException exception message

### LINKED ISSUE #304: Add urls to exceptions
It would be easier to debug some issues if the URL being connected to was included in the exception.

There are two main examples of this:
1. `com.cloudant.client.org.lightcouch.CouchDbException: Error retrieving server response`
   could be improved to
   `com.cloudant.client.org.lightcouch.CouchDbException: Error retrieving server response for https://example:1234/_all_dbs`
2. Other exception types e.g. those from the `java.net` package would also benefit e.g.
   `java.net.SocketTimeoutException: Read timed out`
   but these messages are initialized outside of our control and cannot be reset or appended to. One possibility is to wrap or re-create the exceptions with an extended message. Alternatively we could ensure that the logging associated with these exceptions is improved e.g. `com.cloudant.http.interceptors.CookieInterceptor interceptResponse
   SEVERE: Failed to get response code from request`
   might be changed to
   `com.cloudant.http.interceptors.CookieInterceptor interceptResponse
   SEVERE: Failed to get response code from request for https://example:1234/_all_dbs`

We should definitely improve 1 and I believe we already sanitize user info from the URL so it should be safe to include it in that exception.
Case 2 I think needs some consideration as to the best solution, but just improving the logging is probably the simplest even though it requires more changes.
  - All
  - Any
  - N/A
  - The stack trace for the test `com.cloudant.tests.CloudantClientTests#connectionTimeout`

FYI @xpqz

smithsz:
## What
Include URL in the `CouchDbException` exception message (where applicable).

## How
Add a private `url` attribute to `CouchDbException`. The URL can be set via a setter or through an additional constructor. `.getMessage()` now includes the URL (if defined).

## Testing
No additional tests.

## Issues
Fixes #304.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbException.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbException.java
@@ -39,6 +39,7 @@ public class CouchDbException extends RuntimeException {
     }
 
     private int statusCode;
+    private String url = null;
     protected String error = null;
     protected String reason = null;
 
@@ -74,9 +75,23 @@ public class CouchDbException extends RuntimeException {
 
     @Override
     public String getMessage() {
-        //improve the message with status code, error and reason if we have them
-        return ((getStatusCode() > 0) ? getStatusCode() + " " : "") + super.getMessage()
-                + ((error != null) ? ": " + getError() : "")
-                + ((reason != null) ? ": " + getReason() : "");
+        String msg = super.getMessage();
+        // trim trailing full stop
+        msg = (msg.endsWith(".")) ? msg.substring(0, msg.length() - 1) : msg;
+
+        // include the status code, URL, error and reason (if available)
+        return ((getStatusCode() > 0) ? getStatusCode() + " " : "") + msg
+                + ((url != null) ? " at " + url : "") + "."
+                + ((error != null) ? " Error: " + getError() + ".": "")
+                + ((reason != null) ? " Reason: " + getReason() + ".": "");
+    }
+
+    /**
+     * Set the URL that resulted in this exception being thrown.
+     *
+     * @param url the {@link String} representation of a URL
+     */
+    public void setUrl(String url) {
+        this.url = url;
     }
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Unreleased
+- [FIXED] Add URL to `CouchDbException` exception message (where applicable) for easier debugging.
 - [NEW] Add additional method to `GET` standalone attachments.
 - [FIXED] Issue with "+" (plus) not being regarded as a reserved character in URI path components.
 - [FIXED] Issue with double encoding of restricted URL characters in credentials when using
--- a/cloudant-client/.gitignore
+++ b/cloudant-client/.gitignore
@@ -1,1 +1,2 @@
 /build
+/bin/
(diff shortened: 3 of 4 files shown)


cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbException.java
```diff
@@ -52,6 +53,11 @@ public CouchDbException(String message, Throwable cause, int statusCode) {
         this.statusCode = statusCode;
     }
 
+    public CouchDbException(String message, Throwable cause, String url) {
```
rhyshort: Since we aren't calling it, we should remove the constructor.

ricellis: I guess it would be nice to do it the same way in both places, so I'm fine with this just calling the setter too.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbException.java
```diff
@@ -76,7 +82,17 @@ public String getReason() {
     public String getMessage() {
         //improve the message with status code, error and reason if we have them
         return ((getStatusCode() > 0) ? getStatusCode() + " " : "") + super.getMessage()
+                + ((url != null) ? ": " + url : "")
```
ricellis: The current message looks like:
`404 Object Not Found: not_found: no_db_file`
I think the new message will look like:
`404 Object Not Found: https://examples.cloudant.com/mydb: not_found: no_db_file`

I wonder if use of the `:` in both the URL scheme and the message separators is a bit confusing.

Maybe we can just tack the URL on the end with "at"?
`404 Object Not Found: not_found: no_db_file at https://examples.cloudant.com/mydb`

Or maybe we should just make the message more human readable altogether, something like:
`404 Object Not Found at https://examples.cloudant.com/mydb. Error: not_found. Reason: no_db_file.`

Thoughts @smithsz @rhyshort

smithsz: +1 for the human readable `404 Object Not Found at https://examples.cloudant.com/mydb. Error: not_found. Reason: no_db_file`.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbException.java
```diff
@@ -81,10 +76,11 @@ public String getReason() {
     @Override
     public String getMessage() {
         //improve the message with status code, error and reason if we have them
-        return ((getStatusCode() > 0) ? getStatusCode() + " " : "") + super.getMessage()
-                + ((url != null) ? ": " + url : "")
-                + ((error != null) ? ": " + getError() : "")
-                + ((reason != null) ? ": " + getReason() : "");
+        String msg = super.getMessage().replaceAll("\\.$", "");
```
ricellis: Whilst this is tidy, I wonder if it might be a bit inefficient to use a regex.
In my simplistic testing using:
`msg = (super.getMessage().endsWith(".")) ? super.getMessage().substring(0, msg.length() - 1) : super.getMessage();`

is around 13 times faster when the replacement needs to be made, and more than 20 times faster in the case when it doesn't. Whilst hopefully we won't be writing exception messages often it might be worth considering.

smithsz: Sure, it does seem overkill to compile this same regex each time. I wonder how a pre-compiled pattern would fare. Either way, I'll update with your suggestion. Thanks.

ricellis: Cool.
Changing my test to use a pre-compiled pattern improves it, but it is still 9 and 14 times slower.

PR #340, opened 2017-01-18
Matrix build

rhyshort:
## What

Introduce matrix builds.

## How

- Modified the Jenkinsfile to split FindBugs from the `cloudant` tests and run in parallel
- Added new nodes for CouchDB 1.6 and CouchDB 2.0. 

## Testing

No new tests, only ReplicatorTest modified to create the `_replicator` database during set-up.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,7 +53,17 @@ stage('QA') {
                 node("couchdb1.6") {
                     unstash name: 'built'
                     try {
-                        sh 'sudo /usr/local/bin/wrapdocker \n docker pull couchdb:latest \n docker run -d -p 5984:5984 --name couchdb couchdb\n ./gradlew test'
+                        sh 'sudo /usr/local/bin/wrapdocker \n docker pull couchdb:1.6.1 \n docker run -d -p 5984:5984 --name couchdb couchdb:1.6.1\n ./gradlew test'
+                    } finally {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -24,23 +24,51 @@ stage('Build') {
...
+                        try {
+                            sh './gradlew -Dtest.with.specified.couch=true -Dtest.couch.username=$DB_USER -Dtest.couch.password=$DB_PASSWORD -Dtest.couch.host=clientlibs-test.cloudant.com -Dtest.couch.port=443 -Dtest.couch.http=https -Dtest.couch.ignore.compaction=true -Dtest.couch.auth.headers=true cloudantServiceTest'
+                        } finally {
+                            junit '**/build/test-results/*.xml'
+                        }
+                    }
+                }
+            },
+            couchdb1dot6: {
+                node("couchdb1.6") {
```
rhyshort: Although this uses DinD (Docker in Docker) to run the server we are testing against it appears to not work with the standard slave image, however it works with the image created to run CouchDB from our Artifactory. We can change this for DinD tag instead but i thought i'd get some feedback before making the changes.

ricellis:
Suggest we close this for now, given #348.
Hopefully we can pick it up again sometime when maybe we can do it without DinD. Maybe using peer containers or a multi docker test topology of some kind.

PR #343, opened 2017-01-25
Fixed connect test

ricellis:
## What

Fixed long delay in the `connectionTimeout()` test.

## How

The `connectionTimeout()` test was passing, but taking a lot longer than it should. Refactored to use the loopback address for that test to resolve the issue on some JVMs. On OpenJDK there is no difference between the connect and read timeouts (at least for the way this test provokes a connect timeout), so had to add a read timeout as well to make sure it didn't take 5 minutes.

Removed `SleepingDispatcher` class in favour of using MockWebServer’s `SocketPolicy.NO_RESPONSE`, which was able to recreate the behaviour needed for the timeout tests using the sleeping dispatcher. Refactored those tests as necessary.

## Testing

Test only changes.

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
@@ -16,7 +16,6 @@ package com.cloudant.tests;
 
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.createPost;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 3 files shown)


cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -312,17 +320,12 @@ public void connectionTimeout() throws Throwable {
...
     @Test(expected = SocketTimeoutException.class)
     public void readTimeout() throws Throwable {
-
-        final Long READ_TIMEOUT = 250l;
-
-        //start a simple http server
-        server.setDispatcher(new MockWebServerResources.SleepingDispatcher(READ_TIMEOUT * 2,
-                TimeUnit.MILLISECONDS));
-
+        // Don't response so the read will timeout
```
smithsz: *respond

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientTests.java
```diff
@@ -312,17 +320,12 @@ public void connectionTimeout() throws Throwable {
      */
     @Test(expected = SocketTimeoutException.class)
     public void readTimeout() throws Throwable {
```
smithsz: docstring needs amending to remove SleepingDispatcher reference

Issue #166, opened 2015-12-11
Support text index with Query


rhyshort:
Support the use of text indexes with the `_find` and `_index` endpoints. 

See [Cloudant Docs](https://docs.cloudant.com/cloudant_query.html)

ricellis:
`Database#createIndex(String indexDefinition)` does a `POST` of the `indexDefinition` string to `_index` so it should be possible to create a text index already by that means.

`Database#createIndex(String indexName, String designDocName, String indexType, IndexField[] fields)` would need some modifications to support the different `field` syntax of a text index, but there is an additional option `default_field` for  text indexes that still would not be supported by this signature. At that point we end up with a different method for text indexes, which makes the `indexType` parameter pretty pointless.

The big limitation is not being able to specify and index name or design doc for creating a text index so that could be improved. We could also possibly make improvements for constructing the index JSON string for a text index or have a function for creating the recommended (all fields) text index. However, to do that currently is not too challenging
```java
db.createIndex("{\"type\":\"text\",\"index\": {}}");
```

I can't see anything about either of the `Database#findByIndex(...)` methods that would not work with a text index.

@rhyshort was there something in particular that you thought was not supported or was it mainly the doc that said "only json as of now" that made you raise this issue?

rhyshort:
I can't really remember, (this was a while ago) probably the only json for now, plus you need to be able to get bookmarks back from text index queries, which I don't think is support when using findByIndex iirc

ricellis:
I've moved this to the later milestone, because I think there is a lot more API consideration that needs to happen here. As it currently stands it feels like we would have two awkward APIs and I'd rather consider a unified approach that works for either index type.

ricellis:
A workaround for creating text indexes is documented in a duplicate issue https://github.com/cloudant/java-cloudant/issues/375#issuecomment-333768204

PR #349, opened 2017-03-22
Add createTextIndex

tomblench:
See #166

_What_

Add support for creating text indexes in Cloudant Query

_How_

Add `createTextIndex` method to `Database` and `TextIndexField` class.

_Testing_

Add `testTextIndex` test method.

_TODO_
- ~~javadoc~~
- other documentation
- more tests?


ricellis:
As discussed if we want to put this in version 2.x I think we could register a custom deserializer, but I think that in itself could be reasonably simple (i.e. convert to a `JsonObject` check the necessary values then initialize an appropriate `IndexField`).

PR #351, opened 2017-04-20
Support faceted search with multiple values for the drill down

### LINKED ISSUE #350: Search.drilldown  doesn't take multi value for each drilldown

2.8.0
1.8  IBM
yes

The current  drilldown api in Search  only takes 1 value per drilldown.  There is no way to specify multiple values for the drilldown.

for Example ,  I would want to drilldown on "tags" with 2 vaules
drilldown=["tags", "tag1", "tag2"]

Cloudant Rest api  does support this but current java api doesn't allow this  operation.

  /**
     * @param fieldName  the name of the field
     * @param fieldValue the value of the field
     * @return this for additional parameter setting or to query
     * @see <a target="_blank" href="https://docs.cloudant.com/search.html#faceting">drilldown
     * query parameter</a>
     */
    public Search drillDown(String fieldName, String fieldValue) {
        assertNotEmpty(fieldName, "fieldName");
        assertNotEmpty(fieldValue, "fieldValue");
        JsonArray drillDownArray = new JsonArray();
        JsonPrimitive fieldNamePrimitive = new JsonPrimitive(fieldName);
        drillDownArray.add(fieldNamePrimitive);
        JsonPrimitive fieldValuePrimitive = new JsonPrimitive(fieldValue);
        drillDownArray.add(fieldValuePrimitive);
        databaseHelper.query("drilldown", drillDownArray, false);
        return this;
    }

  
There should be a method as below which can take multiple values  as it takes in for counts.

  /**
     * @param fieldName  the name of the field
     * @param fieldValues array of  values of the field
     * @return this for additional parameter setting or to query
     * @see <a target="_blank" href="https://docs.cloudant.com/search.html#faceting">drilldown
     * query parameter</a>
     */
    public Search drillDown(String fieldName, String[] fieldValues) {
        assertNotEmpty(fieldName, "fieldName");
        assert (fieldValues.length > 0);
        JsonArray drillDownArray = new JsonArray();
        JsonPrimitive fieldNamePrimitive = new JsonPrimitive(fieldName);
        drillDownArray.add(fieldNamePrimitive);
        for(String fieldValue: fieldValues) {
             JsonPrimitive fieldValuePrimitive = new JsonPrimitive(fieldValue);
            drillDownArray.add(fieldValuePrimitive);
        }
        databaseHelper.query("drilldown", drillDownArray, false);
        return this;
    }

smithsz:
## What
Add faceted search variable argument to `drillDown` method allowing multiple drill down values to be specified.

_This is a minor version change to the API as it does not break binary compatibility._

## How
`drillDown` method now takes variable argument as `fieldValues`.

## Testing
Includes additional until test.

## Issues
- Fixes #350

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# UNRELEASED
+- [NEW] Add faceted search variable argument to `drillDown` method allowing multiple drill down
+  values to be specified for a single field name.
+- [DEPRECATED] The `drillDown(String, String)` method. Use new `drillDown(String, String...)` which
+  allows multiple drill down values.
+
 # 2.8.0 (2017-02-15)
 - [NEW] Added `bluemix` method to the client builder allowing service credentials to be passed using
   the CloudFoundry VCAP_SERVICES environment variable.
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Search.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Search.java
@@ -365,7 +365,9 @@ public class Search {
      * @return this for additional parameter setting or to query
      * @see <a target="_blank" href="https://docs.cloudant.com/search.html#faceting">drilldown
      * query parameter</a>
+     * @deprecated Use {@link #drillDown(String, String...)}
      */
+    @Deprecated
     public Search drillDown(String fieldName, String fieldValue) {
         assertNotEmpty(fieldName, "fieldName");
         assertNotEmpty(fieldValue, "fieldValue");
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# UNRELEASED
+- [NEW] Add faceted search variable argument to `drillDown` method allowing multiple drill down
+  values to be specified.
```
ricellis: maybe add `for a single field name.` since it was already possible to have multiple `drillDown` values with different field names.

cloudant-client/src/main/java/com/cloudant/client/api/Search.java
```diff
@@ -360,25 +360,26 @@ public Search counts(String[] countsfields) {
...
-     * @param fieldName  the name of the field
-     * @param fieldValue the value of the field
+     * @param fieldName   the name of the field
+     * @param fieldValues field values to add
      * @return this for additional parameter setting or to query
      * @see <a target="_blank" href="https://docs.cloudant.com/search.html#faceting">drilldown
      * query parameter</a>
      */
-    public Search drillDown(String fieldName, String fieldValue) {
+    public Search drillDown(String fieldName, String... fieldValues) {
```
tomblench: I wrote a quick test and found that this will break API compatibility and code compiled against the old API will fail at runtime with a `java.lang.NoSuchMethodError` because the signature has changed.

You'll need to provide two overloads of the same method to ensure API compatibility.

smithsz: Thanks - I've deprecated the old method rather than outright removing it.

PR #358, opened 2017-06-26
Add IAM session interceptor

tomblench:
_What_

- Move shared functionality into CookieInterceptorBase

- Add iamApiKey option to ClientBuilder

_Testing_

- Add mock test

- TODO add more tests...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# 2.10.0 (Unreleased)
+- [NEW] Add IAM cookie authentication method.
+
 # 2.9.0 (2017-04-26)
 - [NEW] Add faceted search variable argument to `drillDown` method allowing multiple drill down
   values to be specified for a single field name.
--- a/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015, 2016 IBM Corp. All rights reserved.
+ * Copyright © 2015, 2017 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -26,6 +26,7 @@ import com.cloudant.http.HttpConnectionInterceptor;
 import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
 import com.cloudant.http.internal.interceptors.CookieInterceptor;
+import com.cloudant.http.internal.interceptors.IamCookieInterceptor;
 import com.cloudant.http.internal.interceptors.ProxyAuthInterceptor;
 import com.cloudant.http.internal.interceptors.SSLCustomizerInterceptor;
 import com.cloudant.http.internal.interceptors.TimeoutCustomizationInterceptor;
@@ -160,6 +161,8 @@ public class ClientBuilder {
     private TimeUnit connectTimeoutUnit = TimeUnit.MINUTES;
     private long readTimeout = DEFAULT_READ_TIMEOUT;
     private TimeUnit readTimeoutUnit = TimeUnit.MINUTES;
+    private String iamApiKey;
+    private URL iamServer;
 
     /**
      * Constructs a new ClientBuilder for building a CloudantClient instance to connect to the
@@ -233,6 +236,7 @@ public class ClientBuilder {
         }
     }
 
+
     /**
      * Build the {@link CloudantClient} instance based on the endpoint used to construct this
      * client builder and the options that have been set on it before calling this method.
@@ -248,9 +252,27 @@ public class ClientBuilder {
 
         props.addRequestInterceptors(USER_AGENT_INTERCEPTOR);
 
+        if (this.iamApiKey != null) {
 
+            // read iamServer from system property
+            try {
+                this.iamServer = new URL(System.getProperty("com.cloudant.client.iamserver",
+                        "https://iam.bluemix.net/oidc/token"));
+            } catch (MalformedURLException mue) {
+                throw new CouchDbException("IAM server property was not a valid URL", mue);
+            }
+
+            // Create IAM cookie interceptor and set in HttpConnection interceptors
+            IamCookieInterceptor cookieInterceptor = new IamCookieInterceptor(this.iamApiKey,
+                    this.iamServer, this.url.toString());
+
+            props.addRequestInterceptors(cookieInterceptor);
+            props.addResponseInterceptors(cookieInterceptor);
+            logger.config("Added IAM cookie interceptor");
+
+        }
         //Create cookie interceptor
-        if (this.username != null && this.password != null) {
+        else if (this.username != null && this.password != null) {
             //make interceptor if both username and password are not null
 
             //Create cookie interceptor and set in HttpConnection interceptors
@@ -711,4 +733,18 @@ public class ClientBuilder {
                 String.format("Cloudant service instance matching name %s was not found.", instanceName)
         );
     }
+
+    /**
+     * Sets the
+     * <a href="https://console.bluemix.net/docs/services/Cloudant/guides/iam.html#ibm-cloud-identity-and-access-management"
+     * target="_blank">IAM</a> API key for the client connection.
+     *
+     * @param iamApiKey the IAM API key for the session
+     * @return this ClientBuilder object for setting additional options
+     */
+    public ClientBuilder iamApiKey(String iamApiKey) {
+        this.iamApiKey = iamApiKey;
+        return this;
+    }
+
 }
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
@@ -1,0 +1,426 @@
+/*
+ * Copyright © 2017 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.cloudant.tests;
+
+import static com.cloudant.tests.HttpTest.takeN;
+import static com.cloudant.tests.util.MockWebServerResources.EXPECTED_OK_COOKIE;
+import static com.cloudant.tests.util.MockWebServerResources.EXPECTED_OK_COOKIE_2;
+import static com.cloudant.tests.util.MockWebServerResources.IAM_TOKEN;
+import static com.cloudant.tests.util.MockWebServerResources.IAM_TOKEN_2;
+import static com.cloudant.tests.util.MockWebServerResources.OK_IAM_COOKIE;
+import static com.cloudant.tests.util.MockWebServerResources.OK_IAM_COOKIE_2;
+import static com.cloudant.tests.util.MockWebServerResources.iamSession;
+import static com.cloudant.tests.util.MockWebServerResources.iamSessionUnquoted;
+import static com.cloudant.tests.util.MockWebServerResources.setCookie;
+import static org.hamcrest.CoreMatchers.anyOf;
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+import com.cloudant.client.api.CloudantClient;
+import com.cloudant.client.org.lightcouch.CouchDbException;
+import com.cloudant.http.Http;
+
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import mockit.Invocation;
+import mockit.Mock;
+import mockit.MockUp;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+
+import java.nio.charset.Charset;
+
+/**
+ * Created by tomblench on 30/06/2017.
+ */
+
+public class HttpIamTest {
+
+    @Rule
+    public MockWebServer mockWebServer = new MockWebServer();
+
+    @Rule
+    public MockWebServer mockIamServer = new MockWebServer();
+
+    final static String hello = "{\"hello\":\"world\"}\r\n";
+    final static String iamApiKey = "iam";
+    final static String iamTokenEndpoint = "/oidc/token";
+
+    /**
+     * Mock System.getProperty to fake the value of the IAM server
+     *
+     * Do this instead of setting the property in the real System because it pollutes the JVM
+     */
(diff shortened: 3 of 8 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -711,4 +726,16 @@ public static ClientBuilder bluemix(String vcapServices, String instanceName) {
                 String.format("Cloudant service instance matching name %s was not found.", instanceName)
         );
     }
+
+    public ClientBuilder iamApiKey(String iamApiKey) {
```
ricellis: Needs documentation.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -345,6 +345,91 @@ public void cookieRenewal() throws Exception {
...
     /**
+     * This test check that the cookie is renewed if the server presents a Set-Cookie header
+     * after the cookie authentication.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void iamCookieRenewal() throws Exception {
+        final String hello = "{\"hello\":\"world\"}\r\n";
+        final String authSession = "AuthSession=";
```
ricellis: I believe the IAM session is going to have a different name, should be `IAMSession=...`

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -345,6 +345,91 @@ public void cookieRenewal() throws Exception {
     }
 
     /**
+     * This test check that the cookie is renewed if the server presents a Set-Cookie header
+     * after the cookie authentication.
+     *
```
ricellis: It might be worth documenting the flow expected in this test. As it stands I don't think it is valid. If the (couch) server sends a new `Set-Cookie` then it is effectively providing us with a new session, under which circumstances we should have no need to contact the IAM server. I don't think that an IAMSession will ever be renewed in that way, but if it was we should just use the new cookie, i.e. we should assert 0 `_iam_session` requests.

As stated in the PR description I think more tests are needed. Equivalent mock tests for doing:
* a successful initial session
* a successful renewal on a 401
* failure to get a token initially
* failure to get a token on renewal

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
```diff
@@ -0,0 +1,294 @@
...
+        if (shouldAttemptCookieRequest.get()) {
+
+            HttpURLConnection connection = context.connection.getConnection();
+
+            // If we got a 401 or 403 we might need to renew the cookie
+            try {
+                boolean renewCookie = false;
+                int statusCode = connection.getResponseCode();
+
+                if (statusCode == HttpURLConnection.HTTP_FORBIDDEN || statusCode ==
```
ricellis: I think it may be worth factoring parts of this out. The `403` `credentials_expired` checks are very specific to Cloudant I think and it is probably not worth doing all the JSON extraction etc for `403` in IAM cases.

As there is still a decent amount of common functionality that makes it frustrating to override an entire `checkResponse` method we can probably break this out into something like `getRenewalStatusCodes` (which can default to 401) and `performAdditionalChecks()` which can default to a no-op or whatever fits.

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
```diff
@@ -0,0 +1,294 @@
...
+            logger.log(Level.SEVERE, "Failed to read cookie response", e);
+        }
+        return false;
+    }
+
+    protected boolean requestCookie(HttpConnectionInterceptorContext context) {
+        return requestCookie(context,
+                sessionURL,
+                sessionRequestBody,
+                "application/x-www-form-urlencoded",
```
ricellis: I think we should default to JSON here if we can as the encoding is likely more reliable. I am expecting `_iam_session` to accept JSON **if** the `oidc/token` can as well then it leaves the normal `_session` as the odd one out that can do an override.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -248,9 +252,20 @@ public CloudantClient build() {
 
         props.addRequestInterceptors(USER_AGENT_INTERCEPTOR);
 
+        if (this.iamApiKey != null) {
+            //make interceptor if both username and password are not null
```
tomblench: Fixed in c07ebeee7aab28094fa751034627ddb42a3b0ee7

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
tomblench: Placeholder doc done in c07ebeee7aab28094fa751034627ddb42a3b0ee7

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -711,4 +726,16 @@ public static ClientBuilder bluemix(String vcapServices, String instanceName) {
                 String.format("Cloudant service instance matching name %s was not found.", instanceName)
         );
     }
+
+    public ClientBuilder iamApiKey(String iamApiKey) {
+        this.iamApiKey = iamApiKey;
+        return this;
+    }
+
+    public ClientBuilder iamServer(URL iamServer) {
```
tomblench: See c07ebeee7aab28094fa751034627ddb42a3b0ee7 - the exact property name is up for grabs

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
tomblench: See the test changes introduced in c9c6e145d6e110fc3f77abe7f2677119381d7bee - the comments and the use of a few helper methods should make the flow a bit more clear.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
tomblench: Done in c07ebeee7aab28094fa751034627ddb42a3b0ee7

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
tomblench: Not sure what the actual format of the JSON will be, once that's clear I can change this

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
tomblench: This part is specific to the Cloudant session cookie and is invoked either after getting the first IAM token or before getting a refreshed IAM token. So unless the IAM session cookie code is going to be implemented in a radically different way I think most (all?) of it is still relevant.

Of course, we'll have to see once we run it against a real server.

ricellis: My view is that we should never try and parse the JSON from a 403 response looking for `credentials_expired` except for the `_session` endpoint. Neither the `oidc/token` nor `_iam_session` are going to return that.

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
tomblench: See 1259141 - the session cookie is forwarded as-is in JSON format

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
tomblench: See 34da68e - it was too awkward to share any behaviour in the super and sub-classes so each has their own behaviour

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# 2.9.1 (Unreleased)
```
ricellis: Adding new API methods will make this `2.10.0`

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
```diff
@@ -0,0 +1,434 @@
...
+     * - GET a resource on the cloudant server
+     * - Cookie jar empty, so get IAM token followed by session cookie
+     * - GET now proceeds as normal, expected cookie value is sent in header
+     * @throws Exception
+     */
+    @Test
+    public void iamTokenAndCookieSuccessful() throws Exception {
+
+        // Request sequence
+        // _session request to get Cookie
```
ricellis: `_iam_session`?

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
```diff
@@ -0,0 +1,434 @@
...
+    @Before
+    public void setupSystemMock() {
+        new SystemMock();
+    }
+
+    /**
+     * Test IAM token and cookie flow:
+     * - GET a resource on the cloudant server
+     * - Cookie jar empty, so get IAM token followed by session cookie
+     * - GET now proceeds as normal, expected cookie value is sent in header
```
ricellis: cookie value is sent in replayed request header?

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
```diff
@@ -0,0 +1,434 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void iamTokenAndCookieSuccessful() throws Exception {
+
+        // Request sequence
+        // _session request to get Cookie
+        // GET request -> 200 with a Set-Cookie
+        mockWebServer.enqueue(OK_IAM_COOKIE);
+        mockWebServer.enqueue(new MockResponse().setResponseCode(200).addHeader("Set-Cookie",
```
ricellis: I don't think there will be a `Set-Cookie` here, I think this will just a be a normal response, the `Set-Cookie` has already happened on the previous `_iam_session` response.

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
```diff
@@ -0,0 +1,434 @@
...
+     * - third GET on cloudant server, cookie expired, get IAM token and session cookie and replay
+     *   request
+     * @throws Exception
+     */
+    @Test
+    public void iamTokenAndCookieWithExpirySuccessful() throws Exception {
+
+        // Request sequence
+        mockWebServer.enqueue(OK_IAM_COOKIE);
+        mockWebServer.enqueue(new MockResponse().setResponseCode(200).addHeader("Set-Cookie",
```
ricellis: More unnecessary/unexpected `Set-Cookie`s - appears to affect later tests as well.

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
```diff
@@ -37,7 +38,37 @@
...
 
     public static final String EXPECTED_OK_COOKIE =
-            "AuthSession=\"a2ltc3RlYmVsOjUxMzRBQTUzOtiY2_IDUIdsTJEVNEjObAbyhrgz\"";
+            "a2ltc3RlYmVsOjUxMzRBQTUzOtiY2_IDUIdsTJEVNEjObAbyhrgz";
+    public static final String EXPECTED_OK_COOKIE_2 =
+            "dG9tYmxlbmNoOjU5NTM0QzgyOhqHa60IlqPmGR8vTVIK-tzhopMR";
+
+    // these IAM tokens are treated as opaque strings by the interceptor so it doesn't matter if
+    // they're not representative of the real thing which is a json object with an expiry, etc
+    public static final String IAM_TOKEN = "{\"access_token\": \"IamToken1\"}";
```
ricellis: We should probably mock out the rest of the token content here, given that we know what it looks like.

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
tomblench: The request isn't replayed because we know ahead of time that we need to get the cookie (because the cookie jar is empty). Does that make sense?

ricellis: Sure, we have the optimization for no cookie. Guess I was thinking about a renewal flow, sorry!

CHANGES.md
tomblench: Done in 77bab8d

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
tomblench: Done in d59bfbe

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
tomblench: Done in 4951698

cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java
tomblench: Done in 4951698

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
tomblench: Done in 1ff9ff7

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
```diff
@@ -0,0 +1,269 @@
...
+                    if (errorStream != null) {
+                        // Consume the error stream to avoid leaking connections
+                        String error = IOUtils.toString(errorStream, "UTF-8");
+                        // Log the error stream content
+                        logger.fine(error);
+                    }
+                } finally {
+                    IOUtils.closeQuietly(errorStream);
+                }
+                // TODO these messages should be customised so we can tell if bearer or session token failed
```
ricellis: Let's resolve this TODO by adding the URL to the message which will make it clear which endpoint caused a problem.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -248,9 +252,27 @@ public CloudantClient build() {
 
         props.addRequestInterceptors(USER_AGENT_INTERCEPTOR);
 
+        if (this.iamApiKey != null) {
+
+            // read iamServer from system property
+            try {
+                this.iamServer = new URL(System.getProperty("com.cloudant.client.iamserver",
+                        "https://iam.ng.bluemix.net/oidc/token"));
```
smithsz: should be `https://iam.bluemix.net/oidc/token`

tomblench: Thanks for spotting, fixed in d6cb7ab

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptor.java
```diff
@@ -192,7 +130,7 @@ public HttpConnectionInterceptorContext interceptResponse(HttpConnectionIntercep
                     }
 
                     if (renewCookie) {
-                        logger.finest("Cookie was invalid attempt to get new cookie.");
+                        logger.finest("Cookie was invalid. Will attempt to get new cookie.");
```
smithsz: As discussed, we should consider checking if the cookie is expired prior to making the request.

ricellis:
@tomblench Many of the existing files need a latest copyright date update to 2017. Sorry I failed to spot that in my first couple of passes.

PR #360, opened 2017-07-03
Update Gradle and gitignore

amdelamar:
## What

- Updated Gradle from 2.14.1 to 4.0
- Update gitignore.io

## How

Just saw the older version of Gradle and thought it would be helpful to contribute. Also, as I built the project on my local I noticed git wanted to commit .class files, so I updated the gitignore from [gitignore.io](http://gitignore.io/). (mainly to prevent .class files from being committed but there are some other nice additions as well).
If either of these are not necessary, please ignore.

## Testing

No changes to tests.
I ran `./gradlew check` and got `413 tests completed, 5 failed, 1 skipped`. The 5 failures are from replicator tests, which appears probably because I used local CouchDB without a `_replicator` database. You might want to run the tests yourself on this PR branch just in case.

## Issues

No issues.


ricellis:
Thanks for this PR and sorry for the delay in looking at it. Unfortunately it looks like there is a test failure in the [Travis CI check](https://travis-ci.org/cloudant/java-cloudant/builds/249676432?utm_source=github_status&utm_medium=notification) for `com.cloudant.tests.HttpTest#okUsableClassLoad`. It seems likely this could be due to changes in class output directories introduced in Gradle 4, but given that there is currently no specific requirement to upgrade to Gradle 4 I didn't really dig into it. The changes in output directories might explain why the class files were missed by the current .gitignore as well.

Anyway please feel free to open a new PR if you can resolve the failure and get a green Travis check.

amdelamar:
Thanks. And not a problem. Maybe I'll get back to this in the future. Gradle is on 4.2 now, but with gradle wrapper this isn't entirely necessary to update right away.

PR #365, opened 2017-08-15
363 cookie expiry check

### LINKED ISSUE #363: Cookie interceptors should check for expiry
We can optimise the cookie interceptors by checking if the existing cookie has expired in the request interceptor. This can be done by storing the cookie expiry date and checking it for each request, and we can avoid having to replay a request we know will fail.

ricellis:
- [ ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://github.com/cloudant/java-cloudant/blob/master/DCO1.1.txt)
- [ ] You have added tests for any code changes
- [ ] You have updated the [CHANGES.md](https://github.com/cloudant/java-cloudant/blob/master/CHANGES.md) 
- [ ] You have completed the PR template below:

## What

Added test for expired cookies.

## How

Added new tests (see below).
Added handling in `CookieInterceptorBase` to avoid potentially having two consecutive `/` separators in URLs.

## Testing

* Added `SessionInterceptorExpiryTests` for preemptive renewal of expired cookies.
    * Test with and without OkHttp.
    * Using both `_session` and `_iam_session`.
* Updated test utilities to suit.

## Issues

Fixes #363


cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
```diff
@@ -42,63 +43,85 @@
...
 
-    // helper for asserts etc
-    public static String authSession(String cookie) {
-        return String.format(Locale.ENGLISH, "AuthSession=\"%s\"", cookie);
+    // helpers for making Cookie responses
+    public static final String AUTH_COOKIE_NAME = "AuthSession";
+    public static final String IAM_COOKIE_NAME = "IAMSession";
+
+    private static final String quoteValue(String value) {
```
emlaver: I believe `final` can be removed.  IDEA resource: `As private methods cannot be meaningfully overridden, declaring them final is redundant.`

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
```diff
@@ -42,63 +43,85 @@
...
+        return "\"" + value + "\"";
+    }
+
+    public static String authSession(String value) {
```
emlaver: This method is not being called by any tests but can replace the `AuthSession` variable in HttpTest: https://github.com/cloudant/java-cloudant/blob/0e4ae1da8fccf9222719490bc1a27e23f36ccb5c/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java#L305

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
```diff
@@ -42,63 +43,85 @@
...
+        return authSessionUnquoted(quoteValue(value));
     }
 
-    public static String authSessionUnquoted(String cookie) {
-        return String.format(Locale.ENGLISH, "AuthSession=%s", cookie);
+    public static String authSessionUnquoted(String value) {
```
emlaver: Should this be removed? I did not find any tests that called this method

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
ricellis: Fixed in eceba97d5889e1809dcb57092d7a78b4a3243b5a

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
ricellis: Agreed. done in c882cc27e237ac1255af786032276c5ebb165cf5.

cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java
ricellis: Tom made the matching versions for IAM and non-IAM cases. I guess it never got used, but my change in c882cc27e237ac1255af786032276c5ebb165cf5 now uses it :)

cloudant-client/src/test/java/com/cloudant/tests/SessionInterceptorExpiryTests.java
```diff
@@ -0,0 +1,166 @@
...
+     * Test that if a cookie is expired it does not cause a replay cycle. That is we should not
+     * retrieve an expired cookie from the store, so a new session request should be made before
+     * any subsequent request.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testNewCookieRequestMadeAfterExpiry() throws Exception {
+        // Make a GET request and get a cookie valid for 2 seconds
+        executeTest(System.currentTimeMillis() + 2000, MockWebServerResources.EXPECTED_OK_COOKIE);
```
tomblench: Would it be possible to create a cookie which is already invalid? For example
```
executeTest(System.currentTimeMillis() - 10000, MockWebServerResources.EXPECTED_OK_COOKIE)
```
That way you don't need to have the sleep statement and the test can run more quickly.

ricellis: Yes, although I guess it depends what we are trying to verify since I think the expired cookie is never added to the store vs being in the store, but not being returned on the next request.

ricellis: Yeah so it doesn't work because the already expired cookie is never added to the store.
GET / -> intercepted because no cookie
POST _session -> returns already expired cookie
GET / fails because no cookie added since it was already expired.

Leaving as is.

tomblench: Ah shame, fair enough

PR #371, opened 2017-08-30
Session connection leaks

ricellis:

## What

Fixed various connection leaks.

## How

Fixed connection leaks:
* Ensured the error stream is consumed in `CookieInterceptorBase`.
* Added a defensive attempt to cleanup to the `HttpConnection` so any missed consumes will be handled as a last resort.

Renamed some fields from `*Session*` to `*Token*` in the `IamCookieInterceptor` to avoid confusion since they were associated with requests to the token endpoint not the `_iam_session` endpoint.

## Testing

Existing tests using OkHttp exposed messages in the logs for these leaks.
Fixed tests that were themselves leaking connections:
* `com.cloudant.tests.CloudantClientTests#testUserInfoWithSpecials`.
* `com.cloudant.tests.CloudantClientTests#testUserInfoInUrlWithSpecials`.
* `com.cloudant.tests.CloudantClientTests#testUserInfoInUrl`.
* `com.cloudant.tests.HttpTest` inputStreamRetry tests.
* `com.cloudant.tests.SessionInterceptorExpiryTests`

Fixed NPE in `com.cloudant.tests.HttpIamTest`.

## Issues

N/A

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
@@ -50,6 +50,7 @@ import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
 
+import org.apache.commons.io.IOUtils;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
@@ -562,6 +563,9 @@ public class HttpTest extends HttpFactoryParameterizedTest {
                 try {
                     if (444 == context.connection.getConnection().getResponseCode()) {
                         context.replayRequest = true;
+                        // Close the error stream
+                        InputStream errors = context.connection.getConnection().getErrorStream();
+                        if (errors != null) IOUtils.closeQuietly(errors);
                     }
                 } catch (IOException e) {
                     e.printStackTrace();
--- a/cloudant-client/src/test/java/com/cloudant/tests/SessionInterceptorExpiryTests.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/SessionInterceptorExpiryTests.java
@@ -17,14 +17,17 @@ package com.cloudant.tests;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import com.cloudant.http.Http;
 import com.cloudant.http.HttpConnection;
 import com.cloudant.http.HttpConnectionRequestInterceptor;
 import com.cloudant.http.HttpConnectionResponseInterceptor;
+import com.cloudant.http.internal.DefaultHttpUrlConnectionFactory;
 import com.cloudant.http.internal.interceptors.CookieInterceptor;
 import com.cloudant.http.internal.interceptors.IamCookieInterceptor;
+import com.cloudant.http.internal.ok.OkHttpClientHttpUrlConnectionFactory;
 import com.cloudant.tests.util.HttpFactoryParameterizedTest;
 import com.cloudant.tests.util.MockWebServerResources;
 
@@ -39,6 +42,7 @@ import okhttp3.mockwebserver.RecordedRequest;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 public class SessionInterceptorExpiryTests extends HttpFactoryParameterizedTest {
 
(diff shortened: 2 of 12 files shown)


cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
```diff
@@ -211,6 +222,15 @@ public boolean call(HttpConnection connection) throws IOException {
                         if (sessionHasStarted(connection.responseAsInputStream())) {
                             return storeCookiesFromResponse(connection.getConnection());
                         } else {
+                            // If the session did not start, consume the error stream
+                            InputStream errorStream = connection.getConnection().getErrorStream();
+                            if (errorStream != null) {
```
emlaver: This same block of code is used on this [line](https://github.com/cloudant/java-cloudant/pull/371/files#diff-fe23e1df4307844289ddf96bd367bae8R131) and in [HttpConnection](https://github.com/cloudant/java-cloudant/pull/371/files#diff-03970475f79bf0e33474af285ca0cb6fR363).  Maybe put this in a utils class to reduce duplicate code?

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
```diff
@@ -562,6 +563,9 @@ private void testInputStreamRetry(HttpConnection request, byte[] expectedContent
                 try {
                     if (444 == context.connection.getConnection().getResponseCode()) {
                         context.replayRequest = true;
+                        // Close the error stream
+                        InputStream errors = context.connection.getConnection().getErrorStream();
+                        if (errors != null) IOUtils.closeQuietly(errors);
```
tomblench: This could be troublesome if we want to update our commons-io dependency, which is a bit out of date.

https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/IOUtils.html#closeQuietly(java.io.InputStream)

```
Deprecated. As of 2.6 removed without replacement. Please use the try-with-resources statement or handle suppressed exceptions manually
```

It looks like this method is also null-guarded, although that feature isn't documented.

cloudant-client/src/test/java/com/cloudant/tests/SessionInterceptorExpiryTests.java
```diff
@@ -108,9 +112,17 @@ private void queueResponses(Long expiry, String cookieValue) {
     private void executeTest(Long expiryTime, String cookieValue) throws Exception {
         queueResponses(expiryTime, cookieValue);
         HttpConnection conn = Http.GET(mockWebServer.url("/").url());
+        conn.connectionFactory = (okUsable) ? new OkHttpClientHttpUrlConnectionFactory() :
```
tomblench: Do we have a matrix of these tests which run with OK and without? I haven't really thought about that until now.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
```diff
@@ -354,6 +355,25 @@ public HttpConnection execute() throws IOException {
...
+            // If we're going to retry we should consume any existing error streams to avoid
+            // leaking connections. Consuming the stream is preferable to just closing it as it
+            // makes the connection eligible for re-use.
+            if (retry && numberOfRetries > 0) {
+                InputStream errorStream = connection.getErrorStream();
+                if (errorStream != null) {
+                    try {
+                        // Copy the stream to a null destination
+                        IOUtils.copy(errorStream, NullOutputStream.NULL_OUTPUT_STREAM);
+                        logger.log(Level.WARNING,"Consumed unused HTTP response error stream.");
```
tomblench: Will the stream be set to `null` if it has been already consumed? Is there a chance that it can still be non-null but return 0 bytes from the copy?

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
```diff
@@ -211,6 +222,15 @@ public boolean call(HttpConnection connection) throws IOException {
                         if (sessionHasStarted(connection.responseAsInputStream())) {
                             return storeCookiesFromResponse(connection.getConnection());
                         } else {
+                            // If the session did not start, consume the error stream
```
tomblench: Should probably include "to avoid leaking connections"

cloudant-client/src/test/java/com/cloudant/tests/SessionInterceptorExpiryTests.java
ricellis: Basically anything that extends `HttpFactoryParameterizedTest`. They work "automatically" if using the `ClientBuilder`. These tests that I added recently work directly with the `HttpConnection` though so needed this `connectionFactory` set to actually use both connection types, something that slipped my mind when I originally added the tests.

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
ricellis: Fair point. We use this in quite a few other places as well. I'll common them up.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
ricellis: It is only `null` if there was no error or the "server provided no useful information". Generally the stream will have been closed if it was consumed, but there is a chance it could return 0 bytes from the copy.

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
ricellis: Fixed in 86353f01a4170131ae71f2debd75069b0e4ead2d

cloudant-client/src/test/java/com/cloudant/tests/HttpTest.java
ricellis: Done in 0687f06cc3dee69903fac1e48661151656056a32. Note just for the cloudant-http module we still have some usages of `closeQuietly` in tests and cloudant-client.

cloudant-http/src/main/java/com/cloudant/http/internal/interceptors/CookieInterceptorBase.java
ricellis: Done in 0687f06cc3dee69903fac1e48661151656056a32.

cloudant-http/src/main/java/com/cloudant/http/HttpConnection.java
tomblench: So should the warning log entry only be made if `IOUtils.copy` returns non-0?

ricellis: Also done in the utility method I added in 0687f06.

PR #374, opened 2017-09-11
Add test.couch.iam.api.key -D option

tomblench:
_What_
Add option to allow execution of tests with  IAM authentication enabled

_How_
Add `test.couch.iam.api.key` `-D` option

_Testing_
Some failing tests to be investigated and fixed or marked ignored in a follow up PR

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/CloudantClientHelper.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/CloudantClientHelper.java
@@ -40,6 +40,7 @@ public abstract class CloudantClientHelper {
     private static final String COUCH_PORT;
     private static final String HTTP_PROTOCOL;
     private static final URL SERVER_URL;
+    private static final String COUCH_IAM_API_KEY;
 
     static {
 
@@ -67,6 +68,8 @@ public abstract class CloudantClientHelper {
                 HTTP_PROTOCOL = System.getProperty("test.couch.http", "http"); //should either be
                 // http or https
             }
+
+            COUCH_IAM_API_KEY = System.getProperty("test.couch.iam.api.key");
 
             //now build the URLs
             SERVER_URL = new URL(HTTP_PROTOCOL + "://"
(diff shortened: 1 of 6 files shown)


cloudant-client/src/test/java/com/cloudant/tests/CloudantClientHelper.java
```diff
@@ -118,7 +121,8 @@ public static ClientBuilder newMockWebServerClientBuilder(MockWebServer mockServ
     public static ClientBuilder getClientBuilder() {
         return ClientBuilder.url(SERVER_URL)
                 .username(COUCH_USERNAME)
-                .password(COUCH_PASSWORD);
+                .password(COUCH_PASSWORD)
+                .iamApiKey(COUCH_IAM_API_KEY);
```
ricellis: It may be worth adding a comment here that this test helper is relying on the behaviour of the `ClientBuilder` to do the right thing. I wonder if we should make this return a different builder in each case, and/or add some additional tests to validate that the behaviour remains as expected when combinations are supplied (since we're not throwing exceptions if someone specifies both username/pw and an IAM API key).

e.g.
```java
ClientBuilder builder = ClientBuilder.url(SERVER_URL);
if (COUCH_IAM_API_KEY != null) {
  return builder..iamApiKey(COUCH_IAM_API_KEY);
} else {
  return builder.username(COUCH_USERNAME)
    .password(COUCH_PASSWORD);
}
```
and tests for `testIAMApiKeyPreferredToUsernamePassword()` `testUsernamePasswordFallbackIfNoAPIKey()` or similar.

I think as it stands there is scope for the `build()` behaviour to change and we wouldn't realise that we were testing with the wrong creds.

We should probably also document the behaviour of specifying credentials and a key if we didn't already.

tomblench: @ricellis my preference would be that this helper remains as-is and we have a test like `testIAMApiKeyPreferredToUsernamePassword` as suggested

Issue #184, opened 2016-01-18
Easier generation of selectors


ricellis:
It would be nice to have an easier way to generate `Selector` JSON.
Currently the JSON string is either composed or serialized from a `Map` or some other object.

The discussion in #137 proposed some ideas, such as a DSL or some specific helper object for exposing the available options.

rhyshort:
Since it's going to be a lot of work to create a DSL, We could just recommend in the interim that users use groovy to build up `Map` objects which we could adapt our API to accept. See http://groovy-lang.org/groovy-dev-kit.html#Collections-Maps.

PR #377, opened 2017-10-12
Query selector builder

tomblench:
## What

Easier generation of `selector` for [Cloudant Query](https://docs.cloudant.com/cloudant_query.html)

## How

Add `QueryBuilder` class and classes for operations and expressions with static helpers for compact and fluent creation of selectors aided by IDE auto-complete. 

## Testing

See `QueryTests` class (TODO rename?)

## Issues

See #184

## TODO:

other stuff in the query body:
- ~~fields~~
- ~~sort~~
- ~~limit~~
- ~~skip~~

documentation
- javadoc
- CHANGES
- README

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/query/QueryBuilder.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/query/QueryBuilder.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright © 2017 IBM Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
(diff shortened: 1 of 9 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/query/QueryBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class QueryBuilder {
+
+    OperationOrExpression query;
+
+    public QueryBuilder(OperationOrExpression query) {
+        this.query = query;
+    }
+
+    public String build() {
```
ricellis: Maybe we could reduce the boiler plate even further by making this a static, since I don't think we actually need the `query` to be stored in an instance

```java
public static String build(OperationOrExpression query) {
    return String.format("{\"selector\": {%s}}", query.toString());
}
```

I know naming isn't final, but we could also possibly call the method `selector` to perhaps indicate better how it maps to the HTTP API and maybe drop the `Builder` suffix (if we don't have an explicit `build` method anyway) then:
`new QueryBuilder(gt("year", 2018)).build());`
could be
`Query.selector(gt("year", 2018));`

It's not a massive saving, but is a bit more concise.

PR #380, opened 2017-10-23
Improve handling of non-existent doc ids in view response

### LINKED ISSUE #379: Improve handling of non-existent doc id in view response
The test below gets a list of docs using the `keys` option where the doc id is non-existent:
```
        String id1 = db.save(new Foo()).getId();
        String id2 = "non-existing-doc";

        List<JsonObject> allDoc = db.getAllDocsRequestBuilder().keys(id1, id2)
                .includeDocs(true)
                .build()
                .getResponse()
                .getDocsAs(JsonObject.class);
```
The result `allDoc` size is two where you'd expect the size to be one.  Also, in RowImpl class, we are not checking if the result documents have the `error` ket.

We should check that row has the `doc` key and does not have `error` key in [RowImpl.getDocument](https://github.com/cloudant/java-cloudant/blob/dde30d749c19eb90e2b1b7f25b26d58a934d8613/cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java#L61) and [RowImpl.getDocumentAsType](https://github.com/cloudant/java-cloudant/blob/dde30d749c19eb90e2b1b7f25b26d58a934d8613/cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java#L65).
I think returning `null` if `{"error":"not_found"}` is valid.  We'll need to consider what the right behavior should be for other errors.

emlaver:
## What

Improved handling of non-existent doc ids in view response.

## How

- Check that row has `doc` key and does not have `error` key

## Testing
Added `_all_docs` tests with a) only non-existing doc ids and b) an existing and non-existing doc id.

## Issues

fixes #379

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
@@ -111,6 +111,14 @@ public interface ViewResponse<K, V> extends Iterable<ViewResponse<K, V>> {
          * @since 2.0.0
          */
         <D> D getDocumentAsType(Class<D> docType);
+
+        /**
+         * Gets the error message if an error exists.
+         *
+         * @return the error message for this row or null if there is no error
+         * @since 2.10.0
+         */
+        String getError();
     }
 
     /**
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/views/AllDocsRequestResponse.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/views/AllDocsRequestResponse.java
@@ -72,6 +72,17 @@ public class AllDocsRequestResponse implements AllDocsRequest, AllDocsResponse {
         return response.getKeys();
     }
 
+    @Override
+    public Map<String, String> getErrors() {
+        Map<String, String> errors = new HashMap<String, String>();
+        for (ViewResponse.Row<String, Revision> row : response.getRows()) {
+            if(row.getError() != null) {
+                errors.put(row.getKey(), row.getError());
+            }
+        }
+        return errors;
+    }
+
     /*
          * Object representation of rev field from a JSON object.
          * <P>
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
@@ -58,11 +58,23 @@ public class RowImpl<K, V> implements ViewResponse.Row<K, V> {
 
     @Override
     public Document getDocument() {
-        return gson.fromJson(row.get("doc"), Document.class);
+        return getDocumentAsType(Document.class);
     }
 
     public <D> D getDocumentAsType(Class<D> docType) {
-        return gson.fromJson(row.get("doc"), docType);
+        D doc = null;
+        if(row.has("doc")) {
+            doc = gson.fromJson(row.get("doc"), docType);
+        }
+        return doc;
     }
 
+    @Override
+    public String getError() {
+        String error = null;
+        if(row.has("key") && row.has("error")) {
+            error = row.get("error").getAsString();
+        }
+        return error;
+    }
 }
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/views/ViewResponseImpl.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/views/ViewResponseImpl.java
@@ -163,7 +163,10 @@ class ViewResponseImpl<K, V> implements ViewResponse<K, V> {
             if (docs == null) {
                 docs = new ArrayList<Document>();
                 for (Row row : getRows()) {
-                    docs.add(row.getDocument());
+                    Document doc = row.getDocument();
+                    if(doc != null) {
+                        docs.add(doc);
+                    }
                 }
             }
             return docs;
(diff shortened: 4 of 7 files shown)


cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
```diff
@@ -58,11 +58,19 @@ public V getValue() {
 
     @Override
     public Document getDocument() {
-        return gson.fromJson(row.get("doc"), Document.class);
+        Document doc = null;
```
ricellis: I just noticed that it would be a lot easier if this just called `return getDocumentAsType(Document.class);`
then the `row` and `error` checks need only happen in one place.

cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
```diff
@@ -58,11 +58,19 @@ public V getValue() {
...
+        return doc;
     }
 
     public <D> D getDocumentAsType(Class<D> docType) {
-        return gson.fromJson(row.get("doc"), docType);
+        D doc = null;
+        if(row.has("doc") && !row.has("error")) {
+            doc = gson.fromJson(row.get("doc"), docType);
+        }
+        return doc;
```
ricellis: If we're returning `null` here we'll need `com.cloudant.client.internal.views.ViewResponseImpl#getDocsAs` to perform a `null` check before adding to the list too.

cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
emlaver: Fixed in 21bfc3c.

cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
emlaver: Fixed in 7f24237.

cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -68,6 +69,15 @@
     <D> List<D> getDocsAs(Class<D> docType);
 
     /**
+     * Gets a map of the document id and error message for all the of the result rows on this page.
+     * For example, if a doc id does not exist the error message will show "not_found".
+     *
+     * @return a map with an entry for key of the doc id and value of error
+     * @since 2.10.0
+     */
+    Map<String, String> getErrors();
```
ricellis: We should just use the Rows for the general view case, sorry if that wasn't clear yesterday e.g.
```
response.getRows().get(0).getError();
```

cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -111,6 +121,14 @@
...
          */
         <D> D getDocumentAsType(Class<D> docType);
+
+        /**
+         * Gets a map of the document id and error message for this row.
+         *
+         * @return a map with an entry for key of the doc id and value of error
+         * @since 2.10.0
+         */
+        Map<String, String> getError();
```
ricellis: I'd prefer this to be `String getError()` to avoid forcing a binding of the error to the ID, which can be returned independently by `getId()` anyway.

cloudant-client/src/main/java/com/cloudant/client/internal/views/RowImpl.java
```diff
@@ -58,11 +61,23 @@ public V getValue() {
...
-        return gson.fromJson(row.get("doc"), docType);
+        D doc = null;
+        if(row.has("doc")) {
+            doc = gson.fromJson(row.get("doc"), docType);
+        }
+        return doc;
     }
 
+    @Override
+    public Map<String, String> getError() {
```
ricellis: As per above this can be simplified if we're just getting the error string.

cloudant-client/src/main/java/com/cloudant/client/internal/views/ViewResponseImpl.java
```diff
@@ -186,6 +192,14 @@
     }
 
     @Override
+    public Map<String, String> getErrors() {
```
ricellis: Remove

cloudant-client/src/main/java/com/cloudant/client/internal/views/AllDocsRequestResponse.java
```diff
@@ -72,6 +72,11 @@ public AllDocsResponse getResponse() throws IOException {
         return response.getKeys();
     }
 
+    @Override
+    public Map<String, String> getErrors() {
+        return response.getErrors();
```
ricellis: this will need to iterate over the rows and build the `Map`

cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -123,12 +113,12 @@
         <D> D getDocumentAsType(Class<D> docType);
 
         /**
-         * Gets a map of the document id and error message for this row.
+         * Gets the error message if an error exists.
          *
-         * @return a map with an entry for key of the doc id and value of error
+         * @return the error message for this row
```
ricellis: or `null` if there is no error

PR #381, opened 2017-10-24
[WIP] 166 text indexes

### LINKED ISSUE #166: Support text index with Query

ricellis:

## What

Added new APIs for creating, deleting and listing indexes, including text index support.

## How

* Deprecated old index list and create methods.
* Added new `Builder`s for creating JSON and text index definitions.
    * `new JsonIndex.Builder()`
    * `new TextIndex.Builder()`
* Added common index types
    * `Index` and `Field` interfaces
    * `InternalIndex`, `NamedField` and `Sort`
* Added specialised models for text and JSON indexes
    * `JsonIndex`
    * `TextIndex`
* Added new `Indexes` class for listing multiple types.
* Added new `Database#deleteIndex(... String type)`, redirected old `deleteIndex` to be for
JSON index only to preserve old behaviour.
* Removed now unused internal code.
* Added `SelectorUtils` for parsing selector strings.
* Added custom serializers for field types and API asymmetry.
* Added `partial_filter_selector` support.

## Testing

* Updated existing `IndexTests` (server) for new API
* Added new `com.cloudant.client.api.query` test package with
    * IndexCreationTests (mocks)
    * IndexDeletionTests (mocks)
    * IndexListTests (mocks)
    * IndexLifecycleTests (server)

## Issues

Fixes #166

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
@@ -15,7 +15,6 @@
 package com.cloudant.client.api;
 
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.assertNotEmpty;
-import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.assertNotNull;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.close;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.createPost;
 import static com.cloudant.client.org.lightcouch.internal.CouchDbUtil.getAsString;
@@ -29,11 +28,14 @@ import com.cloudant.client.api.model.IndexField;
 import com.cloudant.client.api.model.Params;
 import com.cloudant.client.api.model.Permissions;
 import com.cloudant.client.api.model.Shard;
+import com.cloudant.client.api.query.Indexes;
+import com.cloudant.client.api.query.JsonIndex;
 import com.cloudant.client.api.views.AllDocsRequestBuilder;
 import com.cloudant.client.api.views.ViewRequestBuilder;
 import com.cloudant.client.internal.DatabaseURIHelper;
 import com.cloudant.client.internal.URIBase;
 import com.cloudant.client.internal.util.DeserializationTypes;
+import com.cloudant.client.internal.util.SelectorUtils;
 import com.cloudant.client.internal.views.AllDocsRequestBuilderImpl;
 import com.cloudant.client.internal.views.AllDocsRequestResponse;
 import com.cloudant.client.internal.views.ViewQueryParameters;
@@ -48,7 +50,6 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import com.google.gson.JsonParseException;
 import com.google.gson.JsonParser;
 import com.google.gson.JsonPrimitive;
 
@@ -264,7 +265,7 @@ public class Database {
     }
 
     /**
-     * Create a new index
+     * Create a new JSON index
      * <P>
      * Example usage creating an index that sorts ascending on name, then by year.
      * </P>
@@ -287,20 +288,57 @@ public class Database {
      *
      * @param indexName     optional name of the index (if not provided one will be generated)
      * @param designDocName optional name of the design doc in which the index will be created
-     * @param indexType     optional, type of index (only "json" as of now)
+     * @param indexType     optional, type of index (only "json" for this method)
      * @param fields        array of fields in the index
+     * @see Database#createIndex(String)
      */
+    @Deprecated
     public void createIndex(String indexName, String designDocName, String indexType,
                             IndexField[] fields) {
-        JsonObject indexDefn = getIndexDefinition(indexName, designDocName, indexType, fields);
-        createIndex(indexDefn.toString());
+        if (indexType == null || "json".equalsIgnoreCase(indexType)) {
+            JsonIndex.Builder b = JsonIndex.builder().name(indexName).designDocument(designDocName);
+            for (IndexField f : fields) {
+                switch (f.getOrder()) {
+                    case desc:
+                        b.desc(f.getName());
+                        break;
+                    case asc:
+                    default:
+                        b.asc(f.getName());
+                        break;
+                }
+            }
+            createIndex(b.definition());
+        } else {
+            throw new CouchDbException("Unsupported index type " + indexType);
+        }
     }
 
     /**
-     * Create a new index from a JSON string
+     * <P>
+     * Create a new index from a string of JSON representing the index definition
+     * </P>
+     * <P>
+     * Helpers are available to construct the index definition string for JSON and text indexes.
+     * </P>
+     * <P>
+     * Example usage:
+     * </P>
+     * <pre>
+     * {@code
+     * // Create a JSON index with generated names for the field named "Movie_year" with the default
+     * // ascending sort order
+     * db.createIndex(new JsonIndex.Builder().fields(new JsonIndex.Field("Movie_year")).definition());
+     *
+     * // Create a text index for the string field named "Movie_title"
+     * db.createIndex(new TextIndex.Builder().fields(new TextIndex.Field("Movie_title",
+     *     TextIndex.Field.Type.STRING)).definition());
+     * }
+     * </pre>
      *
      * @param indexDefinition String representation of the index definition JSON
-     * @see #createIndex(String, String, String, IndexField[])
+     * @see JsonIndex.Builder
+     * @see com.cloudant.client.api.query.TextIndex.Builder
      * @see <a
      * href="https://console.bluemix.net/docs/services/Cloudant/api/cloudant_query.html#creating-an-index"
      * target="_blank">index definition</a>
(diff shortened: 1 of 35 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/query/Sort.java
```diff
@@ -0,0 +1,81 @@
...
+import com.cloudant.client.internal.query.NamedField;
+import com.google.gson.annotations.SerializedName;
+
+public class Sort extends NamedField {
+
+    public enum Order {
+        /**
+         * ascending
+         */
+        @SerializedName("asc")
```
tomblench: I wonder if this and `Type` could use
`toString → return lowercase`
and
`constructor → construct from lowercase`
rather than `SerializedName`

This is more portable if we ever move away from gson and also allows easy construction of queries/selectors in my code which currently avoids explicitly building the JSON tree.

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -279,20 +281,61 @@ public Shard getShard(String docId) {
...
+     * <P>
+     * Example usage:
+     * </P>
+     * <pre>
+     * {@code
+     * // Create a JSON index with generated names for the field named "Movie_year" with the default
+     * // ascending sort order
+     * db.createIndex(new JsonIndex.Builder().fields(new JsonIndex.Field("Movie_year")).definition());
+     *
+     * // Create a text index for the string field named "Movie_title"
```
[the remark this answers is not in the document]
ricellis: I initially had statics on the `Builder` superclass as short cuts for instantiating the builder types, but it didn't make sense when it moved into the internal package. If we had Java 8 minimum we could add static methods to the interface to instantiate the builders e.g.:
`Index.json().fields(...)`
`Index.text().feilds(...)`
otherwise we'll have to add another class or add it to each index type
`JsonIndex.configure/define/create/builder().fields(...)`

I tried to stay close to the JSON form (as we did with the proposed QueryBuilder). I can definitely see the plus sides of being less like the JSON and easier to use so if we're going to move away from that I'd be inclined to go even further and just move the `asc` `desc` etc onto the `Builder` and make the `fields(...)` method internal e.g.

`JsonIndex.builder().asc("Movie_year").done()`
`TextIndex.builder().string("Movie_title").done()`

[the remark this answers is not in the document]
tomblench: OK so maybe the internal/external API split isn't quite right. I appreciate that you don't want to create even more classes but it doesn't seem quite right to me that `Builder.newInstance()` actually returns (eg) a `JsonIndex`, which is doing double-duty as builder pattern class model/value object.

The naming a bit misleading here, your `Builder` is really a factory of sorts, and the thing it instantiates is a builder of sorts. I just wonder if this isn't very intuitive to an API user.

cloudant-client/src/main/java/com/cloudant/client/api/query/Sort.java
ricellis: I think we'd need both in any case, otherwise we'd still need a custom JSON adapter to call the lower case constructor for deserialization. The way to avoid that would be to break from the Java convention and use `Sort.asc` and `Sort.desc`. Either way I think your proposed query code only goes one way so can cope with a `toString() {return lowercase;}` that we can easily add.

If we go with `.asc("fieldName")` methods on the index builders, we could do similar statics on the `Sort` class to new up the instances instead of using public constructors. It would be shorter and open the avenue to hiding the enum entirely or replacing the values with strings internally instead. It just depends if we think there is any value to the end user in being able to compare to some known set of options vs getting a string.

As an aside the "lock-in" is an annoyance, but we're always going to need some internal JSON adapters for API models even if we allow free choice of deserializer elsewhere (e.g. for user doc models). It's the only way to guarantee things like `serializeNulls()` or similar are set the way we need and to cope with some of the idiosyncrasies of the Couch API. The Java EE JSON specs may provide an avenue to using binding annotations that can be interpreted by multiple JSON parsers, but at the moment I think there are only Jackson and the RI to choose from and I believe they even have some behaviour differences!

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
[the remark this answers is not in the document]
ricellis: The `newInstance()` is entirely internal and not visible to the API user, it's a utility for creating an instance of the correct type of the thing that needs to be built.

cloudant-client/src/test/java/com/cloudant/tests/IndexTests.java
```diff
@@ -22,9 +22,9 @@
 import com.cloudant.client.api.CloudantClient;
 import com.cloudant.client.api.Database;
 import com.cloudant.client.api.model.FindByIndexOptions;
-import com.cloudant.client.api.model.Index;
 import com.cloudant.client.api.model.IndexField;
-import com.cloudant.client.api.model.IndexField.SortOrder;
+import com.cloudant.client.api.query.JsonIndex;
```
tomblench: Not related but this file appears to have incorrect line endings (probably some ancient legacy...)

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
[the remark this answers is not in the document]
tomblench: Sorry, I got confused about `newInstance`, now I understand what that does.

Here is a suggestion for a different way of constructing `Builder`s without the user having to directly know about the internal static `Builder` class. It's only a suggestion and you might not think it's worth the extra code.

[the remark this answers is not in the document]
tomblench: Forgot to add the actual link - [here](https://github.com/cloudant/java-cloudant/pull/382/files)

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
```diff
@@ -0,0 +1,249 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+public class IndexListTests extends MockedServerTest {
+
+    private static String SELECTOR_STRING = "{\"year\":{\"$gt\":2010}}";
+
+    // _all_docs
+    private static String ALL_DOCS = "{\"ddoc\":null,\"name\":\"_all_docs\",\"type\":\"special\"," +
```
tomblench: What's your opinion on moving these out into test fixture files? On the downside it means we have lots of small files, but as it stands these are quite hard to read with the line breaks and escaped quotes

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private void assertJsonIndex(JsonIndex index, String name, String selector, JsonIndex
+            .Field... fields) throws Exception {
+        assertIndex(index, name, "json", selector);
+        // Assert the fields
+        assertEquals("The fields should be correct", Arrays.asList(fields), index.getFields());
+    }
+
+    private void assertTextIndex(TextIndex index, String name, String selector, String analyzer,
```
tomblench: Looking at this test helper makes me realise I hadn't thought about how/if we deserialise selectors. Obviously we have the builder stuff on my PR to help people build them if they want to use that to help them. What do we do about deserialisation?

I can see your approach is to keep it as a String, which is probably the most sensible, and if users want to use a JSON mapper etc to modify it they can. Or we could deserialise as a Map<String,Object> or whatever which is about as descriptive as we can get.

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
```diff
@@ -0,0 +1,249 @@
...
+public class IndexListTests extends MockedServerTest {
+
+    private static String SELECTOR_STRING = "{\"year\":{\"$gt\":2010}}";
+
+    // _all_docs
+    private static String ALL_DOCS = "{\"ddoc\":null,\"name\":\"_all_docs\",\"type\":\"special\"," +
+            "\"def\":{\"fields\":[{\"_id\":\"asc\"}]}}";
+    private static String JSON_SIMPLE = "{\"ddoc\":\"_design/testindexddoc\"," +
+            "\"name\":\"simplejson\",\"type\":\"json\"," +
+            "\"def\":{\"fields\":[{\"Person_dob\":\"asc\"}],\"partial_filter_selector\":{}}}";
```
tomblench: Since we're testing deserialisation, there ought to be one example which uses `selector` instead of `partial_filter_selector`

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
ricellis: Would probably be nicer, we did for some design doc tests.

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
ricellis: The rationale for String was that it could be passed straight back in again (or into a new index builder) without any further manipulation.
I hope that we will be able to tweak your query builder slightly to allow output of a selector alone (since it's not a full query) to help with the building up. I don't think we need to deserialize into a model especially given that it wouldn't be supported by the proposed query builder (for editing as it were).

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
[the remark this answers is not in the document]
ricellis: Done in 1abf58a
`JsonIndex.builder()`
`TextIndex.builder()`
to avoid the new'ing up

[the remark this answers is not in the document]
ricellis: Done in b1a418a
Changed to builder field name methods `asc` `desc` `string` `number` `bool`
each being a var args of field names, but can be called multiple times as well.

cloudant-client/src/main/java/com/cloudant/client/api/query/Sort.java
ricellis: Done in 5a3343d and 5e3f9ba respectively for `toString`
I don't think we need the constructor at this point.

cloudant-client/src/test/java/com/cloudant/tests/IndexTests.java
ricellis: Fixed in 7034ae0

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
ricellis: Added in b36b7b4

cloudant-client/src/test/java/com/cloudant/api/query/IndexListTests.java
ricellis: Done in af864b6

PR #382, opened 2017-10-25
Suggestion for builder alternative

tomblench:
DO NOT MERGE


ricellis:
yep this is what I was trying to suggest [here](https://github.com/cloudant/java-cloudant/pull/381#discussion_r146779533) too so we're on the same page.

PR #384, opened 2017-10-25
Ensured IllegalStateException thrown for view multibuilder

### LINKED ISSUE #383: Creating and executing a ViewMultipleRequest query that has no actual queries added to it -> NPE
When using a `ViewMultipleRequest`, the individual queries are separated by a call to `add()`.

If no actual `add()` call is performed (the set of queries to be executed in this multiple request is empty), the java-cloudant library crashes with a `NullPointerException` when executing the query, i.e. calling `getViewResponses()`.

`Exception in thread "main" java.lang.NullPointerException at com.cloudant.client.internal.views.ViewMultipleRequester.getViewResponses(ViewMultipleRequester.java:47)`

Tested with version 2.9.0.

```java
package bugtest;

import java.net.URL;
import java.util.List;

import com.cloudant.client.api.ClientBuilder;
import com.cloudant.client.api.CloudantClient;
import com.cloudant.client.api.Database;
import com.cloudant.client.api.views.Key;
import com.cloudant.client.api.views.MultipleRequestBuilder;
import com.cloudant.client.api.views.ViewMultipleRequest;
import com.cloudant.client.api.views.ViewRequestBuilder;
import com.cloudant.client.api.views.ViewResponse;
import com.google.gson.JsonNull;

public final class EmptyMultipleQueryBugTest {
	
	public static void main(final String[] args) throws Exception {
		final String urlString = "http://localhost:8080/";
		final URL url = new URL(urlString);
		final String username = "admin";
		final String password = "pass";
		final CloudantClient cloudantClient = ClientBuilder.url(url).username(username).password(password).build();
		final String databaseName = "emptymultiplequerybugtestdatabase";
		final boolean createDatabaseIfNotExist = true;
		final Database database = cloudantClient.database(databaseName, createDatabaseIfNotExist);
		final String designDocumentName = "designdocument";
		final String viewName = "view";
		final ViewRequestBuilder viewRequestBuilder = database.getViewRequestBuilder(designDocumentName, viewName);
		final Key.Type<String> keyType = Key.Type.STRING;
		final Class<JsonNull> valueType = JsonNull.class;
		final MultipleRequestBuilder<String, JsonNull> multipleRequestBuilder = viewRequestBuilder.newMultipleRequest(keyType, valueType);
		// Note, there is no call to multipleRequestBuilder.add()
		final ViewMultipleRequest<String, JsonNull> viewMultipleRequest = multipleRequestBuilder.build();
		final List<ViewResponse<String, JsonNull>> viewResponsesList = viewMultipleRequest.getViewResponses();
	}
	
}
```

Another question then is in what way to fix this. Should java-cloudant silently accept these zero queries, perform these zero queries (= i.e. no http call to Cloudant), and return an empty result list? Or should an exception be thrown that the query is malformed? Or should Cloudant be queried in every case, returning whatever Cloudant returns (empty result set or error)? Anyways, the current behaviour is broken.

ricellis:

## What

Fixed `IllegalStateException` not thrown for `build()` before `add()` when no request parameters were used.

## How

Set `isBuildable=true` in instances after `add()`, but `false` on initial creation
and after adding parameters.

## Testing

Added new `com.cloudant.tests.ViewsTest#multiRequestBuildOnlyAfterAddNoParams` that is similar to `multiRequestBuildOnlyAfterAdd` but tries to build with no request parameters or add called first.
Added some additional permutations.

## Issues

Fixes #383

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/views/MultipleRequestBuilderImpl.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/views/MultipleRequestBuilderImpl.java
@@ -21,7 +21,7 @@ public class MultipleRequestBuilderImpl<K, V> extends
         CommonViewRequestBuilder<K, V, MultipleRequestBuilder<K, V>> implements
         MultipleRequestBuilder<K, V> {
 
-    private boolean isBuildable = false;
+    private boolean isBuildable;
 
     private final ViewMultipleRequester<K, V> multiViewRequester;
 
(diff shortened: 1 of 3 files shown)


cloudant-client/src/main/java/com/cloudant/client/internal/views/MultipleRequestBuilderImpl.java
```diff
@@ -47,6 +45,8 @@ private MultipleRequestBuilderImpl(ViewQueryParameters<K, V> params, ViewMultipl
 
     public MultipleRequestBuilder<K, V> add() {
         multiViewRequester.add(this.viewQueryParameters);
+        //one at least one request has been added, then it is buildable
```
tomblench: Grammatical mistake carried over from previous comment

tomblench: Also, are lines 41 and 42 necessary?
```
         //parameters have been set, not buildable until add() is called
         isBuildable = false;
         return this;
 ```
Since `isBuildable` is set to `false` in the ctor

PR #392, opened 2017-11-09
Use pre-fixed size array list during bulk call

neekrish:
When there are large number of documents, the array list will resize several times if size is not prefixed. Since the size is already known, the size needs to be fixed while creating the array list.

- [ ] You have added tests for any code changes
- [ ] You have updated the [CHANGES.md](https://github.com/cloudant/java-cloudant/blob/master/CHANGES.md) 
- [ ] You have completed the PR template below:

## What

Added initial size to response array construction.

## Testing

Bulk method covered by existing code path, core Java ctor does not need testing.

## Issues

N/A


ricellis:
I've completed the PR template questions. In future please ensure to do this, even if the answer is N/A. I agree we can omit the CHANGES entry and a new test for this small change.

PR #394, opened 2017-11-13
Fix bugs in Changes.readNextRow() for continuous changes feeds

### LINKED ISSUE #390: Changes.hasNext() never return false when last_seq row is received

2.90

java version "1.8.0_131"
Java(TM) 2 Runtime Environment, Standard Edition (IBM build 1.8.0_131-b11 24_Apr_2017_14_30 Mac OS X x64(SR4 FP5))

No

```
while (changeFeed.hasNext()) {
    // process change feed
}
```
The feed is stopped but we could never get out of the while loop because changeFeed.hasNext() never returns false.  

`Changes.readNextRow()` does not handle the case when `row.startsWith("{\"last_seq\":")` is true.  It never break the `while(true)` loop when it receives an end of feed signal from cloudant.

tomblench:
## What

Fix bugs in `Changes.readNextRow()`. Previously it was possible for  `readNextRow()` to hang indefinitely for continuous changes feeds.

## How

Exit the while loop early if the line we read from the stream was null or contains `{"last_seq":`.

In theory `row` should never be null because the stream should block if there is no more data or
send "last_seq" when we get to the end. However, it's a case worth handling so that the loop still
exits even if "last_seq" isn't set.

## Testing

Added mock server test `changesFeedLastSeq`.

## Issues

Fixes #390.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# 2.11.0 (Unreleased)
+- [FIXED] An issue where `Changes.hasNext()` never returns on receipt
+  of `last_seq` for continuous changes feeds.
+
 # 2.10.0 (2017-11-07)
 - [NEW] Added API for upcoming IBM Cloud Identity and Access Management support for Cloudant on IBM
   Cloud. Note: IAM API key support is not yet enabled in the service.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-# 2.11.0 (Unreleased)
+# (Unreleased)
 - [FIXED] An issue where `Changes.hasNext()` never returns on receipt
   of `last_seq` for continuous changes feeds.
 
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
@@ -278,6 +278,7 @@ public class Changes {
             return true;
         }
         // we were stopped, end of changes feed
+        terminate();
         return false;
     }
 
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2011 lightcouch.org
- * Copyright (c) 2015, 2017 IBM Corp. All rights reserved.
+ * Copyright © 2015, 2017 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2011 lightcouch.org
- * Copyright (c) 2015 2016 IBM Corp. All rights reserved.
+ * Copyright (c) 2015, 2017 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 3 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
```diff
@@ -1,309 +1,312 @@
...
+                return false;
+            } else if (row.isEmpty()) {
+                // heartbeat
+                continue;
+            }
+            setNextRow(gson.fromJson(row, ChangesResult.Row.class));
+            return true;
+        }
+        // we were stopped, end of changes feed
+        return false;
```
smithsz: do we need to call `terminate();` here too?

tomblench: Good point, done in 653c4cb

cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
```diff
@@ -1,309 +1,313 @@
...
-        this.nextRow = nextRow;
-    }
-
-    private void terminate() {
-        CouchDbUtil.close(getReader());
-    }
-}
+/*
+ * Copyright (C) 2011 lightcouch.org
+ * Copyright (c) 2015, 2017 IBM Corp. All rights reserved.
```
ricellis: This should be (C) or ©

cloudant-client/src/test/java/com/cloudant/tests/ChangeNotificationsTest.java
```diff
@@ -152,4 +152,44 @@ public void changesCustomParameter() throws Exception {
...
+                "{\"seq\":\"14-g1AAAAGjeJzLYWBgYMlgTmGQT0lKzi9KdUhJMtNLykxPyilN1UvOyS9NScwr0ctLLckBKmRKZEiy____f1YGcyJrLlCA3dwkLcXczISwdlQrjHBbkeQAJJPqobYwgm0xNjdJNTRPJGwC0R7JYwGSDA1ACmjRfoR_DM2NzUwtLUn1jzEhmw5AbAL7iRniJ2NT06Q0U8KmZAEAd6SGpQ\",\"id\":\"giraffe\",\"changes\":[{\"rev\":\"3-7665c3e66315ff40616cceef62886bd8\"}]}\n" +
+                "{\"last_seq\":\"14-g1AAAAGjeJzLYWBgYMlgTmGQT0lKzi9KdUhJMtNLykxPyilN1UvOyS9NScwr0ctLLckBKmRKZEiy____f1YGcyJrLlCA3dwkLcXczISwdlQrjHBbkeQAJJPqobYwgm0xNjdJNTRPJGwC0R7JYwGSDA1ACmjRfoR_DM2NzUwtLUn1jzEhmw5AbAL7iRniJ2NT06Q0U8KmZAEAd6SGpQ\",\"pending\":0}";
+        mockWebServer.enqueue(new MockResponse().setBody(body));
+        Changes c = db.changes().continuousChanges();
+        int nChanges = 0;
+        // check against regression where hasNext() will hang
+        while(c.hasNext()) {
+            nChanges++;
```
ricellis: I know it doesn't matter because of the way `Changes` calls `setNextRow` during `hasNext` (and it is one of the horrible things about this design), but in a "correct" `Iterator` the return value from `hasNext()` should not change unless `next()` has been called. I'm not suggesting we change the implementation since we have longer term plans to refactor changes anyway, but this test should at least call `c.next()` in the loop so that if our `Iterator` was correct this test wouldn't break, and in case anyone uses it as an example.

tomblench: Huh, good point and I didn't even notice we were breaking the `Iterator` contract.

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# 2.11.0 (Unreleased)
```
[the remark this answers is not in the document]
tomblench: Fixed in b70c587

cloudant-client/src/main/java/com/cloudant/client/api/Changes.java
tomblench: Fixed in b70c587

cloudant-client/src/test/java/com/cloudant/tests/ChangeNotificationsTest.java
tomblench: Fixed in b70c587

ricellis: Yes, it's sad. We'll rectify that legacy abomination in the next major.

PR #395, opened 2017-11-14
Allow custom service name in CloudFoundry VCAP_SERVICES environment variable.

### LINKED ISSUE #391: ClientBuilder.bluemix issue with dedicated cloud
I need to retrieve credentials from VCAP_SERVICES.
Its throwing 
_java.lang.IllegalArgumentException: No Cloudant services information present.
  	at com.cloudant.client.api.ClientBuilder.bluemix(ClientBuilder.java:696)_

The reason being 
   ```
 if (services.cloudantNoSQLDB == null || services.cloudantNoSQLDB.size() == 0) {
            throw new IllegalArgumentException("No Cloudant services information present.");
        }
```
The array is named as cloudantNoSQLDB  and for a dedicated cloud env the JSON looks like this 
`{
   **"cloudantNoSQLDB Dedicated**":[
      {
         "credentials":{
            "username":"xxxx",
            "password":"yyyyyy",
            "host":"kkkkk",
            "port":000,
            "url":"xxxxxxxx"
         },
      .............................
...................
         ]
      }
   ]
}
`

The name has "space" Dedicated. and hence the exception.

### PR DESCRIPTION
**Pull request by smithsz**

## What
Previously the service name was hard coded as 'cloudantNoSQLDB'. This can now be
customised by passing a `serviceName` parameter to the `ClientBuilder.bluemix`
method.

## Testing
Includes additional unit tests.

## Issues
Fixes #391.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
-# (Unreleased)
-- [FIXED] An issue where `Changes.hasNext()` never returns on receipt
-  of `last_seq` for continuous changes feeds.
+# UNRELEASED
+- [NEW] Added an extra bluemix method to the client builder allowing a custom service name to be
+  used with the VCAP_SERVICES environment variable content.
+- [FIXED] An issue where `Changes.hasNext()` never returns on receipt of `last_seq` for continuous
+  changes feeds.
 
 # 2.10.0 (2017-11-07)
 - [NEW] Added API for upcoming IBM Cloud Identity and Access Management support for Cloudant on IBM
--- a/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
@@ -15,8 +15,8 @@
 package com.cloudant.client.api;
 
 import com.cloudant.client.api.views.Key;
+import com.cloudant.client.internal.util.CloudFoundryService;
 import com.cloudant.client.internal.util.DeserializationTypes;
-import com.cloudant.client.internal.util.CloudFoundryServices;
 import com.cloudant.client.internal.util.IndexDeserializer;
 import com.cloudant.client.internal.util.SecurityDeserializer;
 import com.cloudant.client.internal.util.ShardDeserializer;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java [hunk 1]
```diff
@@ -672,51 +677,93 @@ public static ClientBuilder bluemix(String vcapServices) {
...
+     *     exist in {@code vcapServices}.</li>
+     * </ul>
+     */
+    public static ClientBuilder bluemix(String vcapServices, String serviceName, String instanceName) {
         if (vcapServices == null) {
             throw new IllegalArgumentException("The vcapServices JSON information was null.");
         }
 
-        CloudFoundryServices services;
+        JsonObject obj;
```
tomblench: Do we also need a `null` check for `serviceName` (and an extra bullet point in the `@throws` documentation)?

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# UNRELEASED
+- [NEW] Allow custom service name in CloudFoundry VCAP_SERVICES environment variable.
```
ricellis: This makes it sound like we're changing the content of the environment variable. Maybe:
>Added an extra `bluemix` method to the client builder allowing a custom service name to be used with the VCAP_SERVICES environment variable content.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
```diff
@@ -672,51 +677,93 @@ public static ClientBuilder bluemix(String vcapServices) {
...
+        return bluemix(vcapServices, "cloudantNoSQLDB", instanceName);
+    }
 
+    /**
+     * Sets Cloudant client credentials by inspecting a service information JSON string. This object
+     * takes the form of a {@code VCAP_SERVICES} environment variable which is a JSON object that
+     * contains information that you can use to interact with a service instance in Bluemix.
+     * <P>
+     * Note: Specifying an instance name is only required when multiple Cloudant service instances
+     * are present. If there is only a single Cloudant service instance then
```
ricellis: "If there is only a single Cloudant service instance then {@link #bluemix(String)} can also be used."
I don't think this is correct in the case where the single Cloudant service instance has a custom service name, as you'd need to provide `String, String`.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
smithsz: I'll add it in. The additional null check is belt and braces really. I guess it'll save us from parsing the json and then throwing.

CHANGES.md
smithsz: Will change.

cloudant-client/src/main/java/com/cloudant/client/api/ClientBuilder.java
smithsz: Yeah, I guess if you've specified a custom `serviceName` and you don't want to specify an `instanceName` then you'd need `.bluemix(vcapServices, serviceName, null)`.

smithsz: I think I'll just remove the note entirely from this bluemix constructor.

PR #398, opened 2017-11-28
Combine query builder & index apis

ricellis:
- [x] You have updated the [CHANGES.md](https://github.com/cloudant/java-cloudant/blob/master/CHANGES.md) - these changes already recorded from previous merges

## What

Common up parts of the `QueryBuilder` and `Index` improvements.

## How

* Added statics to construct `Sort`
* Refactored `QueryBuilder` to use `Sort`
* Renamed `OperationOrExpression` -> `Selector`
* Used a single `Type` enum instead of having two versions

## Testing

Refactored the new tests to accommodate these changes.

## Issues

N/A


cloudant-client/src/main/java/com/cloudant/client/internal/util/SelectorUtils.java
```diff
@@ -17,18 +17,89 @@
...
+import com.cloudant.client.api.query.Selector;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 
 public class SelectorUtils {
 
-    public static JsonObject getPartialFilterSelectorFromString(String selectorJson) {
-        return getSelectorFromString("partial_filter_selector", selectorJson);
+    // Selector object keys
```
tomblench: I wonder whether this has overlap with my `Helpers` class

cloudant-client/src/main/java/com/cloudant/client/api/query/Selector.java
```diff
@@ -0,0 +1,49 @@
+/*
```
tomblench: Did you remove `OperationOrExpression.java`? I couldn't see it in the diff but not sure if it's just a github quirk.

ricellis: Good spot, I thought I'd done a rename, but it looks like I somehow kept the original. Fixed in 3c4aa85cdc3ec79f01851f02027f3af454c2d09b

cloudant-client/src/main/java/com/cloudant/client/internal/util/SelectorUtils.java
ricellis: Moved `SelectorUtils` methods into `Helpers`, unfortunately it didn't look like I could re-use any of the methods. See b2c9a25a46f7754a1a681737a1694b5b0dd0c6eb

ricellis:
Also moved the Predicate classes to the internal package in b4a7072ebbb9e57c9d802367925b76fb6b06a108

PR #420, opened 2018-02-21
Add extra scheduler endpoints and fix tests

tomblench:
## What

Add support for `/_scheduler/docs` and `/_scheduler/jobs` endpoints.

Tests will now run correctly on Couch 2.0 with update_docs=false.

## Testing

Fixed tests in `ReplicatorTest`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 # Unreleased
 
 - [NEW] Add `CloudantClient.metaInformation()`.
+- [NEW] Add the following methods:
+  - `CloudantClient.metaInformation()`,
+  - `CloudantClient.schedulerJobs()`,
+  - `CloudantClient.schedulerDocs()`,
+  - `CloudantClient.schedulerDocs(docId)`.
 - [FIXED] An issue where `getReason()` returned an incorrect value for
   `Response` objects returned by `Database.bulk()`.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
   - `CloudantClient.metaInformation()`,
   - `CloudantClient.schedulerJobs()`,
   - `CloudantClient.schedulerDocs()`,
-  - `CloudantClient.schedulerDocs(docId)`.
+  - `CloudantClient.schedulerDoc(docId)`.
 - [FIXED] An issue where `getReason()` returned an incorrect value for
   `Response` objects returned by `Database.bulk()`.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,6 @@
 
 - [NEW] Add `CloudantClient.metaInformation()` to get meta information from the welcome page.
 - [NEW] Add methods for interacting with the replicator's `_scheduler` endpoint:
-  - `CloudantClient.metaInformation()`,
   - `CloudantClient.schedulerJobs()`,
   - `CloudantClient.schedulerDocs()`,
   - `CloudantClient.schedulerDoc(docId)`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # Unreleased
 
-- [NEW] Add `CloudantClient.metaInformation()`.
-- [NEW] Add the following methods:
+- [NEW] Add `CloudantClient.metaInformation()` to get meta information from the welcome page.
+- [NEW] Add methods for interacting with the replicator's `_scheduler` endpoint:
   - `CloudantClient.metaInformation()`,
   - `CloudantClient.schedulerJobs()`,
   - `CloudantClient.schedulerDocs()`,
--- a/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -355,8 +355,8 @@ public class CloudantClient {
      * @param docId The replication document ID
      * @return Replication document state
      */
-    public SchedulerDocsResponse.Doc schedulerDocs(String docId) {
-        return couchDbClient.schedulerDocs(docId);
+    public SchedulerDocsResponse.Doc schedulerDoc(String docId) {
+        return couchDbClient.schedulerDoc(docId);
     }
 
     /**
--- a/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -22,6 +22,8 @@ import com.cloudant.client.api.model.ApiKey;
 import com.cloudant.client.api.model.IndexField;
 import com.cloudant.client.api.model.Membership;
 import com.cloudant.client.api.model.Task;
+import com.cloudant.client.api.scheduler.SchedulerDocsResponse;
+import com.cloudant.client.api.scheduler.SchedulerJobsResponse;
 import com.cloudant.client.internal.URIBase;
 import com.cloudant.client.internal.util.DeserializationTypes;
 import com.cloudant.client.org.lightcouch.CouchDbClient;
@@ -321,6 +323,40 @@ public class CloudantClient {
         com.cloudant.client.api.Replicator replicator = new com.cloudant.client.api.Replicator
                 (couchDbReplicator);
         return replicator;
+    }
+
+    /**
+     * Lists replication jobs. Includes replications created via /_replicate endpoint as well as
+     * those created from replication documents. Does not include replications which have
+     * completed or have failed to start because replication documents were malformed. Each job
+     * description will include source and target information, replication id, a history of
+     * recent event, and a few other things.
+     *
+     * @return List of replication jobs
+     */
+    public SchedulerJobsResponse schedulerJobs() {
+        return couchDbClient.schedulerJobs();
+    }
+
+    /**
+     * Lists replication document states. Includes information about all the documents, even in
+     * completed and failed states. For each document it returns the document ID, the database,
+     * the replication ID, source and target, and other information.
+     *
+     * @return List of replication document states
+     */
+    public SchedulerDocsResponse schedulerDocs() {
+        return couchDbClient.schedulerDocs();
+    }
+
+    /**
+     * Get replication document state for a given replication document ID.
+     *
+     * @param docId The replication document ID
+     * @return Replication document state
+     */
+    public SchedulerDocsResponse.Doc schedulerDocs(String docId) {
+        return couchDbClient.schedulerDocs(docId);
     }
 
     /**
--- a/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -332,18 +332,18 @@ public class CloudantClient {
      * description will include source and target information, replication id, a history of
      * recent event, and a few other things.
      *
-     * @return List of replication jobs
+     * @return All current replication jobs
      */
     public SchedulerJobsResponse schedulerJobs() {
         return couchDbClient.schedulerJobs();
     }
 
     /**
-     * Lists replication document states. Includes information about all the documents, even in
+     * Lists replication documents. Includes information about all the documents, even in
      * completed and failed states. For each document it returns the document ID, the database,
      * the replication ID, source and target, and other information.
      *
-     * @return List of replication document states
+     * @return All replication documents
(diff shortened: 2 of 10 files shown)


CHANGES.md
```diff
@@ -1,6 +1,10 @@
 # Unreleased
 
-- [NEW] Add `CloudantClient.metaInformation()`.
+- [NEW] Add the following methods:
```
emlaver: Maybe add a piece that the methods are part of `_scheduler` endpoint? e.g. `Added the following _scheduler endpoint methods`

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -323,6 +325,40 @@ public MetaInformation metaInformation() {
         return replicator;
     }
 
+    /**
```
ricellis: Do you think there is advantage in hiving these off to a class of their own? Similarly to what happens with View/Query etc.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -323,6 +325,40 @@ public MetaInformation metaInformation() {
...
+     * Lists replication jobs. Includes replications created via /_replicate endpoint as well as
+     * those created from replication documents. Does not include replications which have
+     * completed or have failed to start because replication documents were malformed. Each job
+     * description will include source and target information, replication id, a history of
+     * recent event, and a few other things.
```
ricellis: event**s**

`and other information`?

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -323,6 +325,40 @@ public MetaInformation metaInformation() {
...
+     *
+     * @return List of replication jobs
+     */
+    public SchedulerJobsResponse schedulerJobs() {
```
ricellis: Your doc says this returns a list, so I'm expecting something like `List<SchedulerJob>`
Wouldn't it be preferable to unpack into something like that, or is there some top level meta information shared between jobs too.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -323,6 +325,40 @@ public MetaInformation metaInformation() {
...
+    }
+
+    /**
+     * Lists replication document states. Includes information about all the documents, even in
+     * completed and failed states. For each document it returns the document ID, the database,
+     * the replication ID, source and target, and other information.
+     *
+     * @return List of replication document states
+     */
+    public SchedulerDocsResponse schedulerDocs() {
```
ricellis: Again you're making me expect a list :)

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -32,6 +32,7 @@
 import com.cloudant.client.api.model.Shard;
 import com.cloudant.client.api.query.Indexes;
 import com.cloudant.client.api.query.JsonIndex;
+import com.cloudant.client.api.scheduler.SchedulerDocsResponse;
```
ricellis: I'm seeing an import here, but no usage?

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDbClient.java
```diff
@@ -277,6 +279,46 @@ public Replicator replicator() {
...
 
+
+    /**
+     * Lists replication jobs. Includes replications created via /_replicate endpoint as well as
+     * those created from replication documents. Does not include replications which have
+     * completed or have failed to start because replication documents were malformed. Each job
+     * description will include source and target information, replication id, a history of
+     * recent event, and a few other things.
+     *
+     * @return List of replication jobs
```
ricellis: Same doc comments as above, given this class is an internal thing it probably doesn't need this duplication anyway.

cloudant-client/src/main/java/com/cloudant/client/api/scheduler/SchedulerJobsResponse.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public String getStartTime() {
+        return startTime;
+    }
+
+    static class History {
+        private String timestamp;
+        private String type;
+
+        public History(String timestamp, String type) {
```
ricellis: Is this constructor needed?

cloudant-client/src/main/java/com/cloudant/client/api/scheduler/SchedulerDocsResponse.java
```diff
@@ -0,0 +1,131 @@
...
+        return offset;
+    }
+
+    public long getTotalRows() {
+        return totalRows;
+    }
+
+    public class Doc {
+
+        public Doc(String database, String docId, long errorCount, String id, Object info,
```
ricellis: Is this constructor needed?

CHANGES.md
ricellis: Yes, I think this should be something more like:
```
- [NEW] Added `CloudantClient.metaInformation()` to get more meta information from the welcome page.
- [NEW] Added methods for interacting with the replicator's `_scheduler` endpoint: ...
```

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
tomblench: @ricellis I've copied and pasted from the CouchDB docs where possible/applicable but I agree that this is not the best wording.

cloudant-client/src/main/java/com/cloudant/client/api/scheduler/SchedulerDocsResponse.java
tomblench: It's not needed for deserialisation. I had thought it was needed to silence findbugs but just double-checked and it's not needed so I'll remove them.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
tomblench: There is top level information in the response, so it can't be a `List<SchedulerJob>`. Should I change the name `SchedulerJobsResponse` to something different and/or amend the javadoc?

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
tomblench: If there were lots of different endpoints I might see the point in having
```java
Scheduler sch = CloudantClient.scheduler();
sch.jobs();
sch.docs();
sch.other(); // ...
```

But it seems hardly worth the bother for two endpoints which are pretty much stateless.

_However_

One thing I have noticed is that both of these endpoints seem to share the same traits as the various other view/all_docs type of endpoints, like `offset`, `total_rows`, and presumably some mechanism for paging through.

So do you think I should try and re-use the existing mechanisms we have for doing that?

ricellis: I thought this made sense, but (from the doc at least) the `_scheduler` endpoints appear to allow `skip` and `limit`, but not `start_key` etc, which would make the key based paging we use for views useless.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
ricellis: I think amending the javadoc would be fine.

CHANGES.md
```diff
@@ -1,6 +1,11 @@
 # Unreleased
 
-- [NEW] Add `CloudantClient.metaInformation()`.
+- [NEW] Add `CloudantClient.metaInformation()` to get meta information from the welcome page.
+- [NEW] Add methods for interacting with the replicator's `_scheduler` endpoint:
+  - `CloudantClient.metaInformation()`,
```
ricellis: Don't think `metaInformation` is needed here as it is called out on the line above (and unrelated to the `_scheduler` endpoint).

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
ricellis: As you say it isn't really vital for a couple of methods, although I think we are up to 3 methods now. I don't know of any proposals to add more functionality to the scheduler, doing it now would potentially protect the `CloudantClient` from an overwhelming number of methods, but there isn't really a lot of boilerplate or anything to package away in a `Scheduler` class so I agree it doesn't seem worth it.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
ricellis: No, because it still isn't a `List<SchedulerJobsRepsonse.Job>`. What we're returning is an object containing metadata and a list of replication jobs. Maybe 
`@return replication jobs information`
is sufficient here?

I think in the `SchedulerJobsResponse` class itself it is fine to say the list of jobs i.e.
```java
/**
* @return List of replication jobs
*/
public List<Job> getJobs() {
  return jobs;
}
```

etc for the other similar methods

CHANGES.md
```diff
@@ -1,6 +1,11 @@
...
+  - `CloudantClient.schedulerJobs()`,
+  - `CloudantClient.schedulerDocs()`,
+  - `CloudantClient.schedulerDocs(docId)`.
```
ricellis: this should probably be singular:
`schedulerDoc(docId)`

or maybe we could just varargs the `schedulerDocs(String... docIds)` to allow 1 or more specific documents to be fetched, with the no-args getting them all, but the annoying thing then is that we'd have to wrap a single document in a `SchedulerDocsResponse`.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -323,6 +325,40 @@ public MetaInformation metaInformation() {
...
+        return couchDbClient.schedulerDocs();
+    }
+
+    /**
+     * Get replication document state for a given replication document ID.
+     *
+     * @param docId The replication document ID
+     * @return Replication document state
+     */
+    public SchedulerDocsResponse.Doc schedulerDocs(String docId) {
```
ricellis: As per earlier comment, this should probably be the singular `schedulerDoc` or rolled into the other method

cloudant-client/src/main/java/com/cloudant/client/api/scheduler/SchedulerDocsResponse.java
```diff
@@ -0,0 +1,119 @@
...
+
+package com.cloudant.client.api.scheduler;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+public class SchedulerDocsResponse {
```
ricellis: I wonder if it is worth this (and its friend `SchedulerJobsResponse`) extending a `SchedulerResponse` class that contains the `offset` and `total_rows` fields and methods.

It would be nice to take it even further and do `SchedulerResponse<T>` and typing with `Doc` or `Job` as needed, but sadly the `getDocs` `getJobs` method naming would be a bit unfriendly then.

tomblench: Well there are other things which `offset` and `total_rows`, like views, as well.

CHANGES.md
tomblench: I changed it to `schedulerDoc(docId)` in 3946f8e as suggested

CHANGES.md
tomblench: Good point, must have been a sloppy merge. Fixed in 5303fc7

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
tomblench: Done in 3946f8e

CHANGES.md
tomblench: How does 41debcc look for the `return` part of the javadoc? I'm wary of javadoc-ing the `Response` classes but I could copy the descriptions from the HTTP API if you think it's useful.

ricellis: No, what you've done is sufficient, the method signatures are self-explanatory

PR #422, opened 2018-02-28
Junit 5

tomblench:
## What

Upgrade tests to JUnit 5 (JUnit Jupiter)

## How

* Refactor tests to use JUnit 5 (`org.junit.jupiter.api.*`) APIs

* Convert existing Resources/Rules etc to JUnit 5 Extensions

* Add base classes in `com.cloudant.tests.base package` for common test
  patterns eg remote-data-base-per-class, remote-database-per-method

* Convert parameterised tests to TestTemplates which use helper
  classes to provide parameters

* Convert JUnit 4 Categories to JUnit 5 Tags

* Upgrade gradle to 4.6

* Add custom JavaExec task to build.gradle to invoke JUnit Platform
  Console Launcher, due to lack of flexibility in current gradle
  support for JUnit Platform.

## Testing

All tests pass, with current exceptions:
* `DesignDocumentTest` - number of tests run is exactly half on this branch. It appears that the existing code ran every test twice, maybe due to the use of `@RunWith(Enclosed.class)`
* `HttpProxyTest` - two test combinations currently disabled due to failure. Previously it was determined that all test combinations would pass if run in _some_ orderings, but now these two parameter combinations always fail. Needs further investigation.

## TODO

* ~~Remove `if (env.BRANCH_NAME == "master" || env.BRANCH_NAME == "junit5")` from `Jenkinsfile` before merging.~~
* ~~Re-visit java version requirements. 1.8 is required to build and run tests now, although production code should still only require 1.6~~ Fixed in  62e82add35d0333e0e41e321fd0f63ea7e125bf0
* ~~Update copyrights~~
* ~~Format source files which have been automatically updated to switch order of assert arguments, which now have long lines~~


ricellis:
Regarding

cloudant-client/src/test/java/com/cloudant/test/main/RequiresCloudant.java
```diff
@@ -14,8 +14,19 @@
...
- * JUnit category to label tests which require Cloudant Service or Cloudant Local
+ * JUnit tag to label tests which require Cloudant Service or Cloudant Local
  */
-public class RequiresCloudant extends RequiresDB {
-}
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Tag("RequiresCloudant")
+@Tag("RequiresDB")
+public @interface RequiresCloudant {
```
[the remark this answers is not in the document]
tomblench: These tags were straight out of the [JUnit guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations), so it's probably considered best practice.

cloudant-client/src/test/java/com/cloudant/api/query/IndexLifecycleTest.java
```diff
@@ -24,22 +24,20 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 import java.util.Collections;
 import java.util.List;
 
 
 // This really requires Couch2.0 + text index support, but we don't have a way of expressing that
-@Category(RequiresCloudant.class)
+@RequiresCloudant
```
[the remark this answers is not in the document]
tomblench: Probably best done in a follow-on PR. I wanted to get the tags exactly the same so that I could check each of the matrix builds executes the correct number of tests, before thinking about changing the tags.

cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
```diff
@@ -241,26 +221,42 @@ public void attachmentStandaloneGivenId() throws IOException, URISyntaxException
         }
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void attachmentStandaloneNullDocNonNullRev() {
-        byte[] bytesToDB = "binary data".getBytes();
-        ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytesToDB);
-        Response response = db.saveAttachment(bytesIn, "foo.txt", "text/plain", null, "1-abcdef");
+        assertThrows(IllegalArgumentException.class, new Executable() {
```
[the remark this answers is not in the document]
tomblench: There are 49 instances of `assertThrows` and wrapping the whole thing in an `Executable` was easier, because:
- That was what the IDE auto-prompted for (lazy, I know)
- It's difficult to tell with each test which statement is the exact one responsible for throwing the exception, which i guess is one of the drawbacks of the old `@Throws` annotation.

So it takes time to understand each test and put the `assertThrows` in the right place, but I agree that the way you have suggested doing it is a lot clearer and more expressive.

Possibly something to look at towards the end of the PR?

build.gradle
```diff
@@ -28,8 +28,8 @@ subprojects {
     // If the version says "snapshot" anywhere assume it is not a release
     ext.isReleaseVersion = !version.toUpperCase(Locale.ENGLISH).contains("SNAPSHOT")
 
-    sourceCompatibility = 1.6
-    targetCompatibility = 1.6
+    sourceCompatibility = 1.8
```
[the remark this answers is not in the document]
tomblench: @ricellis good idea, see 62e82add35d0333e0e41e321fd0f63ea7e125bf0

cloudant-client/src/test/java/com/cloudant/tests/BulkDocumentTest.java
```diff
@@ -15,39 +15,20 @@
...
-
-
-    private static Database db;
-
-    @BeforeClass
-    public static void setUp() {
-        db = dbResource.get();
-    }
+@RequiresDB
+public class BulkDocumentTest extends TestWithDb {
```
ricellis: Couldn't `TestWithDb` reasonably include the `@RequiresDB` annotation and avoid the need for it to be present throughout so many test classes. I'm assuming anything that extends `TestWithDb` also `RequiresDB` though I guess which DB could be important. Something to think about.

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientHelper.java
```diff
@@ -47,6 +48,8 @@
         try {
             //a URL might be supplied, otherwise use the separate properties
             String URL = System.getProperty("test.couch.url");
+            System.out.println(System.getProperties());
```
ricellis: If you think this and its counterpart on the line below are necessary for debugging can you make it a JUL style thing so we can turn it off if we want. Or if it was just overlooked delete it.

cloudant-client/src/test/java/com/cloudant/tests/DatabaseTest.java
```diff
@@ -48,35 +49,32 @@
...
+    private static MockWebServer mockWebServer;
 
-    private static Database db;
-    private static CloudantClient account;
-
-    @BeforeClass
-    public static void setUp() throws Exception {
-        account = clientResource.get();
-        db = dbResource.get();
+    // TODO before class?
```
ricellis: Does this need doing?

cloudant-client/src/test/java/com/cloudant/tests/BulkDocumentTest.java
tomblench: It's a nice idea but we have test classes like `CloudantClientTests` which extend `TestWithDb` but only need the database for a couple of tests, so the whole class need not be marked as `@RequiresDB`

cloudant-client/src/test/java/com/cloudant/tests/CloudantClientHelper.java
tomblench: I'd left some debugging in, especially for when test parameters weren't being passed in properly from gradle. This is no longer needed, so I've got rid of all of these in 96ff132

cloudant-client/src/test/java/com/cloudant/tests/DatabaseTest.java
tomblench: See 1d42697

Per the commit comment "mockWebServer needs to be fetched from extension before each test because the extension is a "per test" extension (overrides beforeEach/afterEach)".

It's a bit wasteful to stop/start the mock web server for each test when currently only one test uses it, but I think the overhead is negligible.

cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
```diff
@@ -241,26 +221,42 @@ public void attachmentStandaloneGivenId() throws IOException, URISyntaxException
...
+            public void execute() throws Throwable {
+                byte[] bytesToDB = "binary data".getBytes();
+                ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytesToDB);
+                Response response = db.saveAttachment(bytesIn, "foo.txt", "text/plain", "", "1-abcdef");
+            }
+        });
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test()
```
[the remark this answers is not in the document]
tomblench: Done in 0b3645d

cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
```diff
@@ -241,26 +221,42 @@ public void attachmentStandaloneGivenId() throws IOException, URISyntaxException
...
-        ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytesToDB);
-        Response response = db.saveAttachment(bytesIn, "foo.txt", "text/plain", null, "1-abcdef");
+        assertThrows(IllegalArgumentException.class, new Executable() {
+            @Override
+            public void execute() throws Throwable {
+                byte[] bytesToDB = "binary data".getBytes();
+                ByteArrayInputStream bytesIn = new ByteArrayInputStream(bytesToDB);
+                Response response = db.saveAttachment(bytesIn, "foo.txt", "text/plain", null,
+                        "1-abcdef");
+            }
```
emlaver: Can we consider moving the code in the block above into a helper util method since it occurs several times in this test class?

cloudant-client/src/test/java/com/cloudant/tests/DocumentsCRUDTest.java
```diff
@@ -93,13 +74,14 @@ public void findJsonObject() {
         assertNotNull(jsonObject);
     }
 
+    /*
     @Test
```
emlaver: Is this commented because the test is failing? If so, should we instead use `@Disabled`?

emlaver:
I'm seeing test classes that import `org.junit.jupiter.api.Assertions` and others that import the utility method such as `org.junit.jupiter.api.Assertions.assertEquals`.  Should we pick one way to import for consistency in all the test classes?

cloudant-client/src/test/java/com/cloudant/api/query/IndexLifecycleTest.java
```diff
@@ -49,14 +47,14 @@
  */
 public class IndexLifecycleTest {
```
emlaver: Something for a future PR to possibly add clarity in the logs: adding `@DisplayName` for test classes e.g. `@DisplayName("Index lifecycle test...")`

cloudant-client/src/test/java/com/cloudant/tests/UnicodeTest.java
```diff
@@ -70,13 +54,10 @@
...
-
-    private Database db;
-
-    @Before
-    public void setup() {
-        db = dbResource.get();
-    }
+    private static final String TESTJSON = "{\"_id\":\"literal\"," +
+            "\"_rev\":\"1-39933759c7250133b6039d94ea09134f\",\"foo\":\"Grüße 日本語 \uD834\uDD1E.\"}\n";
+    private static final String TESTJSON_ESCAPED = "{\"_id\":\"escaped\"," +
```
emlaver: Shouldn't this have `\\` similar to TESTSTRING_ESCAPED above?

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -878,11 +872,10 @@ public void multiRequest() throws IOException {
                 .build();
         int i = 1;
         List<ViewResponse<String, Object>> responses = multi.getViewResponses();
-        assertEquals("There should be 3 respones for 3 requests", 3, responses.size());
+        assertEquals(3, responses.size(), "There should be 3 respones for 3 requests");
```
emlaver: respones -> responses

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -999,15 +1013,15 @@ public void multiRequestMixedReduced() throws IOException {
                 .build();
 
         List<ViewResponse<String, Object>> responses = multi.getViewResponses();
-        assertEquals("There should be 2 respones for 2 requests", 2, responses.size());
+        assertEquals(2, responses.size(), "There should be 2 respones for 2 requests");
```
emlaver: respones -> responses

cloudant-client/src/test/java/com/cloudant/tests/base/TestWithDb.java
```diff
@@ -0,0 +1,30 @@
...
+import com.cloudant.client.api.Database;
+import com.cloudant.tests.extensions.CloudantClientExtension;
+import com.cloudant.tests.extensions.DatabaseExtension;
+import com.cloudant.tests.extensions.MultiExtension;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+// Base class for tests which require a DB which is used throughout the lifetime of the test class
+public class TestWithDb {
```
emlaver: What about having a `TestWithDbPerClass` that would extend a base `TestWithDb ` class? 
`TestWithDbPerTest` would also extend `TestWithDb`.

cloudant-client/src/test/java/com/cloudant/tests/base/TestWithDb.java
```diff
@@ -0,0 +1,30 @@
+package com.cloudant.tests.base;
```
emlaver: Add copyright

cloudant-client/src/test/java/com/cloudant/tests/base/TestWithDbPerTest.java
```diff
@@ -0,0 +1,35 @@
+package com.cloudant.tests.base;
```
emlaver: Add copyright

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -623,19 +622,14 @@ public void scalarValues() throws Exception {
     @Test
     public void viewWithNoResult_emptyList() throws IOException {
         init();
-        assertEquals("The results list should be of length 0", 0, db.getViewRequestBuilder
-                ("example", "by_tag").newRequest(Key.Type.STRING, Object.class).keys
-                ("javax").build().getResponse().getKeys().size());
+        assertEquals(0, db.getViewRequestBuilder("example", "by_tag").newRequest(Key.Type.STRING, Object.class).keys("javax").build().getResponse().getKeys().size(), "The results list should be of length 0");
```
ricellis: It seems like some of these code replacements have produced very long lines. Perhaps at the end of the review process it is just worth applying the formatter to all the classes before the final merge?

cloudant-client/src/test/java/com/cloudant/tests/extensions/CloudantClientMockServerExtension.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright © 2017 IBM Corp. All rights reserved.
```
ricellis: As you said in the TODOs description for this PR there are a bunch of copyright updates to do.

cloudant-client/src/test/java/com/cloudant/tests/extensions/MultiExtension.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
+import org.junit.jupiter.api.extension.Extension;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class MultiExtension implements Extension, AfterAllCallback, AfterEachCallback,
```
ricellis: Remind me why this is necessary, was it that `@RegisterExtension` doesn't currently guarantee an order and `@ExtendWith` doesn't allow arguments to be passed?

tomblench: Yes, see https://github.com/junit-team/junit5/issues/506#issuecomment-368789516

cloudant-client/src/test/java/com/cloudant/tests/UnicodeTest.java
tomblench: Thanks for spotting this inconsistency. These are actually used as "expected" values for assertions later, and the json deserialiser will always unescape these values, so they are correct in that regard.

However, the logic and naming of some variables is confusing so I've fixed that in e8a6acb

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
tomblench: Fixed in f378a13

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
tomblench: Fixed in f378a13

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
tomblench: Good idea, i've added it to the TODOs

cloudant-client/src/test/java/com/cloudant/tests/DocumentsCRUDTest.java
tomblench: Fixed in 3e9d231

cloudant-client/src/test/java/com/cloudant/tests/base/TestWithDb.java
tomblench: @emlaver good idea, done in fe8add3

cloudant-client/src/test/java/com/cloudant/api/query/IndexLifecycleTest.java
tomblench: Probably future PR, but it would be useful

cloudant-client/src/test/java/com/cloudant/tests/AttachmentsTest.java
tomblench: Probably outside the scope of what I'm doing, which is just translating the tests to junit5 without considering too closely how these tests are written.

cloudant-client/src/test/java/com/cloudant/tests/SessionInterceptorExpiryTests.java
```diff
@@ -30,59 +30,114 @@
...
+        @Override
+        public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context) {
+            return Stream.of(invocationContext(false,"/_iam_session"),
+                    invocationContext(false, "/_session"),
+                    invocationContext(true, "/_iam_session"),
+                    invocationContext(true,"/_session"));
+        }
 
-    // Note Parameter(0) okUsable is inherited
+        // because we fill in the args from the left, we can fill in the single argument "okUsable"
```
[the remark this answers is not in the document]
tomblench: I've just removed it in 5d5abf5 because it was wrong and there's no comment needed here

cloudant-client/src/test/java/com/cloudant/test/main/RequiresCloudant.java
```diff
@@ -14,8 +14,19 @@
...
+
 /**
- * JUnit category to label tests which require Cloudant Service or Cloudant Local
+ * JUnit tag to label tests which require Cloudant Service or Cloudant Local
  */
-public class RequiresCloudant extends RequiresDB {
-}
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Tag("RequiresCloudant")
```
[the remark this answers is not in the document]
ricellis: This can get bundled with the "improve tagging" follow on.

cloudant-client/src/test/java/com/cloudant/api/query/IndexLifecycleTest.java
```diff
@@ -49,14 +47,14 @@
...
-    private DatabaseResource dbResource = new DatabaseResource(clientResource);
-    @Rule
-    public RuleChain chain = RuleChain.outerRule(clientResource).around(dbResource);
+    @RegisterExtension
+    public static CloudantClientExtension clientResource = new CloudantClientExtension();
+    @RegisterExtension
+    public static DatabaseExtension.PerTest dbResource = new DatabaseExtension.PerTest(clientResource);
 
-    private Database db;
+    private static Database db;
```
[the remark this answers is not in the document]
tomblench: Fixed in fe2ecb7 by extending TestWithDbPerTest

PR #425, opened 2018-03-09
419 support iam auth in replications

### LINKED ISSUE #419: Support IAM authentication in replication documents.
_Parent Issue: https://github.com/cloudant/cloudant-sync/issues/31_

_Work Required:_

Both client replicator classes need amending to support IAM authentication:
- Via the `POST /_replicate` endpoint...
https://github.com/cloudant/java-cloudant/blob/2.12.0/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/Replication.java#L67
- Via the `replicator` database...
https://github.com/cloudant/java-cloudant/blob/2.12.0/cloudant-client/src/main/java/com/cloudant/client/api/Replicator.java#L62

smithsz:

## What
Support IAM authentication in replications. For example:

- Via `replicator` database:
```java
client.replicator()
        .source("https://foo.cloudant.com/src")
        .sourceIamApiKey("mysourcedbiamapikey")
        .target("https://bar.cloudant.com/tgt")
        .targetIamApiKey("mytargetdbiamapikey")
        .save();
```
- Via `/_replicate` endpoint:
```
client.replication()
        .source("https://foo.cloudant.com/src")
        .sourceIamApiKey("mysourcedbiamapikey")
        .target("https://bar.cloudant.com/tgt")
        .targetIamApiKey("mytargetdbiamapikey")
        .trigger();
```

## How
Expose new methods on the underlying replication document classes. These add the required IAM API key fields to the generated replication document JSON.

Example IAM authenticated replication document:
```java
{
  "source": { "url": "...", "auth": { "iam": { "api_key": "<key>" } } },
  "target": { "url": "...", "auth": { "iam": { "api_key": "<key>" } } },
}
```

## Testing
Includes additional mock unit tests.

## Issues
Fixes #419.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
@@ -151,6 +151,14 @@ public class ReplicatorDocument {
         return replicatorDocument.getRetriesPerRequest();
     }
 
+    public String getSourceIamApiKey() {
+        return replicatorDocument.getSourceIamApiKey();
+    }
+
+    public String getTargetIamApiKey() {
+        return replicatorDocument.getTargetIamApiKey();
+    }
+
     public void setSource(String source) {
         replicatorDocument.setSource(source);
     }
@@ -227,6 +235,13 @@ public class ReplicatorDocument {
         replicatorDocument.setSinceSeq(sinceSeq);
     }
 
+    public void setSourceIamApiKey(String iamApiKey) {
+        replicatorDocument.setSourceIamApiKey(iamApiKey);
+    }
+
+    public void setTargetIamApiKey(String iamApiKey) {
+        replicatorDocument.setTargetIamApiKey(iamApiKey);
+    }
 
     public static class UserCtx {
         private com.cloudant.client.org.lightcouch.ReplicatorDocument.UserCtx userCtx;
--- a/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/ReplicatorDocument.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/ReplicatorDocument.java
@@ -77,6 +77,9 @@ public class ReplicatorDocument extends Document {
     }
 
     private String getEndpointUrl(JsonElement element) {
+        if (element == null) {
+            return null;
+        }
         JsonPrimitive urlString = null;
         if (element.isJsonPrimitive()) {
             urlString = element.getAsJsonPrimitive();
@@ -84,7 +87,54 @@ public class ReplicatorDocument extends Document {
             JsonObject replicatorEndpointObject = element.getAsJsonObject();
             urlString = replicatorEndpointObject.getAsJsonPrimitive("url");
         }
+        if (urlString == null) {
+            return null;
+        }
         return urlString.getAsString();
+    }
+
+    private JsonElement getDestination(JsonElement oldDestination, String newDestination) {
+        if (oldDestination == null || oldDestination.isJsonPrimitive()) {
+            return new JsonPrimitive(newDestination);
+        }
+        JsonObject json = oldDestination.getAsJsonObject();
+        json.remove("url");
+        json.addProperty("url", newDestination);
+        return json;
+    }
+
+    private String getIamApiKey(JsonElement destination) {
+        if (destination.isJsonPrimitive()) {
+            return null;
+        }
+        JsonObject json = destination.getAsJsonObject();
+        JsonObject authJson = json.getAsJsonObject("auth");
+        if (authJson == null) {
+            return null;
+        }
+        JsonObject iamJson = authJson.getAsJsonObject("iam");
+        if (iamJson == null) {
+            return null;
+        }
+        return iamJson.getAsJsonPrimitive("api_key").getAsString();
+    }
+
+    private JsonObject getDestinationIamJson(String url, String iamApiKey) {
+        JsonObject iamJson = new JsonObject();
+        if (iamApiKey != null) {
+            iamJson.addProperty("api_key", iamApiKey);
+        }
+
+        JsonObject authJson = new JsonObject();
+        authJson.add("iam", iamJson);
+
+        JsonObject json = new JsonObject();
+        if (url != null) {
+            json.addProperty("url", url);
+        }
+        json.add("auth", authJson);
+
+        return json;
     }
 
     public Boolean getContinuous() {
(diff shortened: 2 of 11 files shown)


cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/ReplicatorDocument.java
```diff
@@ -77,16 +77,65 @@ public String getTarget() {
...
+        }
         return urlString.getAsString();
     }
 
+    private JsonElement getDestination(JsonElement oldDestination, String newDestination) {
+        if (oldDestination == null || oldDestination.isJsonPrimitive()) {
+            return new JsonPrimitive(newDestination);
+        }
+        JsonObject json = oldDestination.getAsJsonObject();
+        json.addProperty("url", newDestination);
```
ricellis: Should this remove any old `url` property first?
I am reasonably sure that this will work as is (at least currently), but there is a small risk here because of two factors:
i) As per [RFC8259 Section 4](https://tools.ietf.org/html/rfc8259#section-4) `The names within an object SHOULD be unique.`
ii) As per the[javadoc for JsonObject](https://static.javadoc.io/com.google.code.gson/gson/2.8.2/com/google/gson/JsonObject.html#addProperty-java.lang.String-java.lang.String-) `The member elements of this object are maintained in order they were added.`

So as it currently stands your new `url` would, I believe, supersede any previous `url` property on the object when it is serialized. The risk is that a change in GSON behaviour or a swap to another JSON provider could make this break.

cloudant-client/src/test/java/com/cloudant/tests/ReplicationMockTest.java
```diff
@@ -0,0 +1,151 @@
...
+    @Test
+    public void createAndTriggerReplicationWithSourceIamAuth() throws Exception {
+        mockWebServer.enqueue(JSON_OK);
+
+        CloudantClient c = CloudantClientHelper.newMockWebServerClientBuilder(mockWebServer)
+                .build();
+
+        c.replication()
+                .source(sourceDbUrl)
+                .sourceIamApiKey(EXPECTED_OK_COOKIE)
```
ricellis: I think we should use something that has the format of the IAM API key to help us identify any possible issues with encoding etc
Looking at it the `com.cloudant.tests.HttpIamTest#iamApiKey` would probably be an appropriate thing to make public and share between tests. Of course the value would need updating to something useful instead of the current `"iam"` string value.

cloudant-client/src/test/java/com/cloudant/tests/ReplicationMockTest.java
```diff
@@ -0,0 +1,151 @@
...
+        c.replication()
+                .source(sourceDbUrl)
+                .target(targetDbUrl)
+                .trigger();
+
+        RecordedRequest[] requests = takeN(mockWebServer, 1);
+
+        assertEquals(requests[0].getPath(), "/_replicate");
+
+        String body = requests[0].getBody().readString(Charset.forName("UTF-8"));
```
ricellis: this can be simplified throughout to:
`getBody().readUtf8()`

cloudant-client/src/test/java/com/cloudant/tests/ReplicatorMockTest.java
```diff
@@ -0,0 +1,239 @@
...
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+
+import java.nio.charset.Charset;
+
+/**
+ * Created by samsmith on 08/03/2018.
+ */
+
+public class ReplicatorMockTest {
```
ricellis: I think these two test classes could share a superclass, since they share some constants and a `@Rule`

cloudant-client/src/test/java/com/cloudant/tests/ReplicationMockTest.java
smithsz: The `EXPECTED_OK_COOKIE` is imported from ` MockWebServerResources`.

It's a bit confusing as it is an IAM API token. Badly named perhaps? There is an `IAM_TOKEN` too but that's the whole JSON response.

See [here](https://github.com/cloudant/java-cloudant/blob/master/cloudant-client/src/test/java/com/cloudant/tests/util/MockWebServerResources.java#L46).

cloudant-client/src/test/java/com/cloudant/tests/ReplicatorMockTest.java
smithsz: Probably not warranted now given the Junit 5 changes.

cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/ReplicatorDocument.java
smithsz: Added the remove.

cloudant-client/src/test/java/com/cloudant/tests/ReplicationMockTest.java
ricellis: The `EXPECTED_OK_COOKIE` is representative of the cookie response to `_session` or `_iam_session` requests. The API key is the thing that gets sent to the IAM server to get an IAM token to send to `_iam_session`. The thing that gets put in the document should be an IAM API key, which is neither the cookie nor the token.

That said, it makes little difference here, they are both the cookie and an IAM API key use Base64 (URL style) strings (i.e `[A-Z][a-z][0-9][-_]`). AFAICT an IAM API key is 45 characters while the cookie is longer, but I think on further investigation I'm not worried about other characters being encoded. It might be beneficial to have something like
```java
// IAM API keys are also Base64 URL strings so use the cookie as an example key
public static final String IAM_API_KEY = EXPECTED_OK_COOKIE;
```
and then using that constant in the tests to avoid it looking like the replicator document should contain a cookie.

cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
```diff
@@ -151,6 +151,14 @@ public Integer getRetriesPerRequest() {
         return replicatorDocument.getRetriesPerRequest();
     }
 
+    public String getSourceIamApiKey() {
+        return replicatorDocument.getSourceIamApiKey();
```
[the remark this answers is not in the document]
ricellis: This is ok.

cloudant-client/src/test/java/com/cloudant/tests/ReplicationMockTest.java
ricellis: @smithsz did this get lost somewhere?

smithsz: Whoops. Must have missed these.

Issue #424, opened 2018-03-07
Replication Document is not being created 


mansura-habiba:
Please include the following information in your ticket.

- CloudantClient (java-cloudant) version(s) that are affected by this issue. - 2.0.12
- Java version (including vendor and platform). - 1.8
- If you're using the optional okhttp dependency. - no
- A small code sample that demonstrates the issue.

ReplicationResult replication = client.replication()
        .source(sourceDb)
        .target(targetDb)
        .createTarget(true)
        .trigger();
List<ReplicationHistory> histories = replication.getHistories();

Expected Outcome

- [ ]  The replication job would create target db and copy data from source db
- [ ] There would be a replication document created on .cloudant.com/dashboard.html#/replication

Actual Outcome
- [ ]  The replication job would create target db and copy data from source db successfully completed.
- [ ] There is no replication document created on .cloudant.com/dashboard.html#/replication

Could you please suggest why the replication  document is not being created in cloudant dashboard for replication
cloudant.com/dashboard.html#/replication

tomblench:
@mansura-habiba if you want a replicator DB document you should use `db.replicator()` as per the example shown here:

https://static.javadoc.io/com.cloudant/cloudant-client/2.12.0/com/cloudant/client/api/Replicator.html

Note that this call is asynchronous so you will probably want to poll the result if you have any logic which is dependent on the replication completing or erroring.

Sorry for the confusion, we have two similarly named methods for replicating:
- `replication` which uses the older [`_replicate`](http://docs.couchdb.org/en/2.1.1/api/server/common.html#replicate) endpoint
- `replicator` which uses [the replicator database](http://docs.couchdb.org/en/2.1.1/replication/replicator.html#replicator)

mansura-habiba:
Thanks @tomblench for your kind reply.
I tried the following code but I am getting the following error 

> The method replicator() is undefined for the type Database

Database db =client.database("database"+tenant, false);
		ReplicationResult replication = db.replicator()
        .source(sourceDb)
        .target(targetDb)
        .createTarget(true)
        .trigger();

Do you have any suggestion

tomblench:
@mansura-habiba it looks like our sample code in the javadoc is incorrect. The replicator API is at the account level, not the database level.

For example:
```java
CloudantClient client = ClientBuilder.account("my-cloudant-account-name").username
        ("username").password("password").build();
Response response = client.replicator().source("https://source").target("https://target").createTarget(true)
        .save();
while (true) {
[...]

PR #426, opened 2018-03-09
Fix javadoc on Replicator

tomblench:
## What

Fix javadoc on the `Replicator` class.

## How

The documentation is misleading because it implies the API is on the `Database` object rather than
the `CloudantClient` object.

## Testing

N/A

## Issues

#424 - user was confused by documentation


smithsz:
Needs https://github.com/cloudant/java-cloudant/pull/425/commits/61ab7c3afa117a147b2625ee25e016773a0901f8 to fix tests.

tomblench:
This has been rebased onto master so should go green since master has the same fix @smithsz mentioned.

Issue #428, opened 2018-03-22
Bulk ignores cloudant responses


al-tr:
- CloudantClient (java-cloudant) version(s) that are affected by this issue - 2.12.0 (and earlier)
- Java version (including vendor and platform) - java 8 - IBM J9 VM
- If you're using the optional okhttp dependency - yes
- A small code sample that demonstrates the issue - below

Validation design document:
```js
function (newDoc, oldDoc, userCtx, secObj) {
    //...
    if (!(typeof newDoc[field] === "boolean") && (!(typeof newDoc[field] === "number"))) {
        if (!newDoc[field]) {
            throw({forbidden: 'field cannot be empty' });
        }
    }
    //...
}
```

JavaModel contains `field`.
documents - a list of JavaModels
code:
```java
        try {
            JavaModel object = documents.get(0);
            object.setField(null);
            Response response = database.post(object);
            // validation function returns 403, exception is thrown
            log.info("{} {} {} {} {}", response.getId(), response.getRev(), response.getStatusCode(), response.getError(), response.getReason());
        } catch (CouchDbException ex) {
            log.info("{} {}: {}", ex.getStatusCode(), ex.getError(), ex.getReason());
        }

        List<Response> responses = database.bulk(documents);

        List<PostResult> postResults = responses.stream()
                .peek(e -> log.info("{} - {} - {} - {} - {}", e.getId(), e.getRev(), e.getStatusCode(), e.getError(), e.getReason()))
                .map(e -> new PostResult(e.getStatusCode(), e.getId(), e.getReason()))
                .collect(Collectors.toList());
```

logs:
```text
CloudantRepository - 403 forbidden: field cannot be empty
CloudantRepository - id1 - null - 201 - forbidden - Created
```
PostResult object:
```json
{
    "httpCode": 201,
    "id": "id1",
    "moreInformation": "Created"
}
```

response.getStatusCode() - always `201`
response.getReason() - always `Created`
response.getError() - `null` when there's no actual error, reason when there's

---
nasty workaround:
```java
if (e.getError() == null) {
    return new PostResult(e.getStatusCode(), e.getId(), "Successfully posted");
} else {
     return new PostResult(400, e.getId(), e.getError());
}
```

emlaver:
@al-tr could you give me more details on what you are expecting? I don't understand this part that you wrote above: "response.getError() - null when there's no actual error, reason when there's".

[...]

PR #436, opened 2018-04-11
Don't over-write reason for bulk()

tomblench:
## What

Preserve the original value of the `reason` value from the response array in the json returned from `_bulk_docs`. This can be queried through `getReason()`

This allows users to retrieve the message from failed validation functions, etc.

Note that the overall HTTP status code is still preserved in each result document and can be queried through `getStatusCode()`. It is therefore the same across all result documents. It should not be relied upon to detect failure:

"The return code from a successful bulk insertion is 201. The content of the returned structure indicates success or other information messages on a per-document basis." (from the IBM Cloudant documentation).

## Testing

New test `bulkRejectedByValidation`

## Issues

See #428

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 # Unreleased
 
 - [NEW] Add `CloudantClient.metaInformation()`.
+- [FIXED] An issue where `getReason()` returned an incorrect value for
+  `Response` objects returned by `Database.bulk()`.
 
 # 2.12.0 (2018-02-08)
 - [NEW] Index creation APIs and builders including support for text and partial indexes.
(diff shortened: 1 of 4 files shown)


cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/CouchDatabaseBase.java
```diff
@@ -287,7 +287,6 @@ public Response remove(String id, String rev) {
                     DeserializationTypes.LC_RESPONSES);
             for(Response response : bulkResponses) {
                 response.setStatusCode(connection.getConnection().getResponseCode());
-                response.setReason(connection.getConnection().getResponseMessage());
```
ricellis: Think this method deserves a javadoc updating regarding the response code, just for clarity.

tomblench: Fixed in 3002eb0

CHANGES.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 
 - [NEW] Add `CloudantClient.metaInformation()`.
+- [FIXED] An issue where `getResponseMessage()` returned an incorrect
```
ricellis: Isn't it `getReason()` rather than `getResponseMessage()`?

Issue #147, opened 2015-11-19
complex key high sentinel value


rabbah:
In the couchdb documentation (http://wiki.apache.org/couchdb/View_collation) a high value sentinel for complex keys is {} (as in end key = ["foo", {}]) but there appears no way to set this on a Key.ComplexKey in the cloudant API v 2.0. Am I missing something?

The Cloudant REST API supports {}.

ricellis:
Thanks for pointing this out you are correct there is no way to set this. We'll look at getting something added to `ComplexKey` to support this.

eli-jordan:
This is something I just hit too. I just invoked the addObjects method on ComplexKey using reflection (for not). SO maybe just exposing that method as public is good enough.

ricellis:
It is also possible to use the String `"\ufff0"` as a high value sentinel.
For example `Key.complex("foo", "\ufff0")` should work.

eli-jordan:
Cool I didn't know about that Unicode trick. I can remove my hacks. 

Since keys can be arbitrary json, I still think it would be useful to be able to express that with the client api

HarikaSSS:
Hi, 
with complex key using startkey and endkey, can i extract a single record of any field?
For Instance,
startKey(Key.complex(null, "Title1", null)).endKey(Key.complex("\ufff0", "Title1", "\ufff0"))
as i have mentioned startkey and endkey both with the same value, it should return only one row with Title1. But it is returning all fields having value greater than Title1.
Is this functionality implemented wrong? or is my understanding wrong? how does it work?
how can i match a single field value when there is a complex key defined in my view?

ricellis:
@HarikaSSS 
> with complex key using startkey and endkey, can i extract a single record of any field?

No, you cannot.

There is a [specific collation order for results in a view](http://docs.couchdb.org/en/latest/ddocs/views/collation.html). As per that collation order arrays are sorted element by element starting with the first element.

The `startkey` and `endkey` identify the bounds of a slice of the ordered results. So this means you can only filter by in-order combinations of elements. For example if your first element was identical (e.g. `"a"`) you could then filter to all the keys `["a","title1"...]`

An example ordering is:
```
"key":["a","title1","a"]
"key":["a","title1","b"]
"key":["a","title1","c"]
"key":["a","title2","a"]
"key":["a","title2","b"]
"key":["a","title2","c"]
"key":["b","title1","a"]
"key":["b","title1","b"]
"key":["b","title1","c"]
[...]

PR #437, opened 2018-04-18
Add ComplexKey.addHighSentinel()

tomblench:
## What

Syntactic sugar to allow use of complex key high sentinel value.

## How

`Add ComplexKey.addHighSentinel()`.

## Testing

Added
`queryPageWithComplexStartEndKeyAndJsonObjectValueUsingEndKey`
`addingAfterEndKeyThrowsError`

## Issues

#147

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
   - `CloudantClient.schedulerJobs()`,
   - `CloudantClient.schedulerDocs()`,
   - `CloudantClient.schedulerDoc(docId)`.
+- [NEW] Add `ComplexKey.addHighSentinel()` to allow matching of all
+  values as part of a complex key range.
 - [FIXED] An issue where `getReason()` returned an incorrect value for
   `Response` objects returned by `Database.bulk()`.
 
--- a/cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 IBM Corp. All rights reserved.
+ * Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -38,7 +38,9 @@ import com.google.gson.JsonSerializer;
 public class Key {
 
     /**
+     * <p>
      * Key type identifier.
+     * </p>
      * <P>
      * Accessed via the constants:
      * </P>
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -5,6 +5,7 @@
   - `CloudantClient.schedulerJobs()`,
   - `CloudantClient.schedulerDocs()`,
   - `CloudantClient.schedulerDoc(docId)`.
+- [NEW] Add `Key.ComplexKey.END_KEY`.
```
ricellis: Would it be worth saying what this is actually for here? e.g. appending "for an array high sentinel value".

cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
```diff
@@ -131,6 +150,9 @@ private ComplexKey(JsonArray json) {
         }
 
         private <T> Key.ComplexKey addObjects(T... keys) {
+            if (!canAddKeys) {
+                throw new IllegalArgumentException("Can't add keys after EndKey has been added");
```
ricellis: I think this actually an `IllegalStateException`?

cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
```diff
@@ -170,6 +192,26 @@ private ComplexKey(JsonArray json) {
             return addObjects(numberKeys);
         }
 
+        public Key.ComplexKey add(EndKey endKey) {
```
ricellis: As you have gone down the method route then the `END_KEY` constant need not be exposed and the method doesn't have to take an argument.
e.g. it could just be `end()` (although that naming could give the impression it needs to always be called) or `addHigh()` or whatever

cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
```diff
@@ -170,6 +192,26 @@ private ComplexKey(JsonArray json) {
...
+            Key.ComplexKey k = addObjects(endKey);
```
ricellis: Can't this just call `addObjects("\ufff0")`?

cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
```diff
@@ -131,6 +153,9 @@ private ComplexKey(JsonArray json) {
         }
 
         private <T> Key.ComplexKey addObjects(T... keys) {
+            if (!canAddKeys) {
+                throw new IllegalStateException("Can't add keys after EndKey has been added");
```
ricellis: This refers to EndKey, but should refer to addHighSentinel

cloudant-client/src/main/java/com/cloudant/client/api/views/Key.java
```diff
@@ -170,6 +195,17 @@ private ComplexKey(JsonArray json) {
             return addObjects(numberKeys);
         }
 
+        /**
+         * Add the high key sentinel value, which is guaranteed to sort last, to the complex key.
+         * @return this ComplexKey
```
ricellis: This should probably say that no further additions can be made

cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -744,6 +781,17 @@ public void allDocsEmptyListWithNonExistingKeys() throws Exception {
         assertThat(response.size(), is(0));
     }
 
+    /**
+     * Assert that IllegalStateException is thrown if any keys are added after calling
+     * addHighSentinel()
+     */
+    @Test
+    public void addingAfterEndKeyThrowsError() {
```
emlaver: Change to `addingAfterHighSentinelThrowsError`?

Issue #427, opened 2018-03-22
AllDocsRequest doesn't contain deleted


al-tr:
- CloudantClient (java-cloudant) version(s) that are affected by this issue - 2.12.0 (and earlier)
- Java version (including vendor and platform) - Java 8 - IBM J9 VM
- If you're using the optional okhttp dependency - yes
- A small code sample that demonstrates the issue - below

There's a `_deleted` field for tombstones in `_all_docs` when you request doc by key.
![image](https://user-images.githubusercontent.com/18145416/37770557-e102a966-2de5-11e8-8797-f3787a4f5c83.png)
and there's no way to get information about it via current API:
![image](https://user-images.githubusercontent.com/18145416/37770490-a3a3edfa-2de5-11e8-975f-f536556f2afe.png)

al-tr:
Use case:
bulk accepts documents both with and without `_rev`.
`1` if there's no `_rev`, bulk "creates" documents as new documents
`2` if there's `_rev`, bulk "updates" documents as existing documents

we don't know whether the documents exist or not

for successful "create" (`1`) we need to provide no `_rev`, easy
for successful "update" (`2`) documents should contain `_rev`, in case of `bulk` the business is around huge quantities of documents and the best way to get `_rev`isions for all the documents we need is the fastest call -> `_all_docs` with `keys=["...", ..., "..."]&include_docs=false`. So there's a method in API `getIdsAndRevs()`. 

But if the document is deleted it has a tombstone for some time and `_all_docs`+keys will return id+rev pair, but also is the flag.

If we bulk a document with an id, and it's `_deleted`... nothing will happen. it won't appear again. in this case, when we have tombstone we need to "create" (`1`) a document without `_rev`, but there's no way to understand it via current (2.12.0) AllDocsRequest API where `_deleted` flag is ignored.

emlaver:
@al-tr The workaround is to use `executeRequest`. See the [java-cloudant documentation for an example.](https://static.javadoc.io/com.cloudant/cloudant-client/2.11.0/com/cloudant/client/api/CloudantClient.html#executeRequest-com.cloudant.http.HttpConnection-)
e.g.
```
//Simple POJO
Foo foo = new Foo();
foo.set_id("foo-doc");
// Save the document
Response r1 = db.save(foo);
// Set rev from response to foo object
foo.set_rev(r1.getRev());
// _deleted=true
foo.set_deleted(true);
List<Object> deleteDoc = new ArrayList<Object>();
[...]

PR #439, opened 2018-04-23
Allow retrieval of deleted flag on all_docs

tomblench:
## What

Allow retrieval of `deleted` flag on `all_docs`

## How

Introduce "sparse document" information which can be retrieved even when `include_docs=false` is
set. This allows for retrieval of `id`, `rev`, and `deleted` flag.

Also, add `null` check to `getDocsAs()`: this ensures that attempting to deserialise deleted documents does not prevent retrieval of non-deleted ones.

## Testing

Added
- `allDocsWithKeysDocDeleted`
- `allDocsWithKeysDocDeletedIncludeDocs`
- `allDocsWithKeysDocDeletedIncludeDocsGetDocsAs`

## Issues

See #427 
See #440

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
@@ -119,6 +119,13 @@ public interface ViewResponse<K, V> extends Iterable<ViewResponse<K, V>> {
          * @since 2.10.0
          */
         String getError();
+
+        /**
+         * Returns a "sparse" document consisting of only the document ID, the revision ID, and the
+         * deleted flag
+         * @return a sparse document
+         */
+        Document getSparseDocument();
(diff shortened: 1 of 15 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java
```diff
@@ -119,6 +119,13 @@
          * @since 2.10.0
          */
         String getError();
+
+        /**
+         * Returns a "sparse" document consisting of only the document ID, the revision ID, and the
+         * deleted flag
+         * @return a sparse document
+         */
+        Document getSparseDocument();
```
emlaver: I'm not sure sparse is the best word to identify these fields but my idea of `getPartialDeletedDoc` or `getIdRevDeletedFields` might not be much better.

PR #443, opened 2018-05-14
Added support for IAM API key in bluemix method

emlaver:

## What

Added support for IAM API key to `ClientBuilder.bluemix`.

## How

- Added `apikey` and `host` in  `CloudFoundryService.java`  to instantiate IAM client in CloudFoundryServiceCredentials
- Handle cases when IAM or legacy credentials do not exist
- Updated Jenkinsfile to set IAM Key for VCAP testing
- Added CHANGES entry
- Updated copyrights

## Testing

Added mock tests for accessing VCAP with IAM credentials, and tests that assert `IllegalArgumentException` is raised when IAM key is missing.

For reviewers:
- With the added change of pulling the IAM key in Jenkins, should we consider "future-proofing" tests by adding a client extension that has the option to build a Cloudant client with IAM e.g. `CloudantClientExtension(boolean useIAM)`?

### DIFF
--- a/cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
@@ -14,12 +14,33 @@
 
 package com.cloudant.tests;
 
+import static com.cloudant.tests.util.MockWebServerResources.IAM_TOKEN;
+import static com.cloudant.tests.util.MockWebServerResources.OK_IAM_COOKIE;
+import static com.cloudant.tests.util.MockWebServerResources.IAM_API_KEY;
+import static com.cloudant.tests.util.MockWebServerResources.iamTokenEndpoint;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import com.cloudant.client.api.ClientBuilder;
+import com.cloudant.client.api.CloudantClient;
+import com.cloudant.http.Http;
+import com.cloudant.http.HttpConnection;
+import com.cloudant.tests.extensions.MockWebServerExtension;
+import com.cloudant.tests.util.IamSystemPropertyMock;
+import com.cloudant.tests.util.MockWebServerResources;
 import com.google.gson.GsonBuilder;
 
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.api.function.Executable;
+
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


ricellis:
>With the added change of pulling the IAM key in Jenkins, should we consider "future-proofing" tests by adding a client extension that has the option to build a Cloudant client with IAM e.g. CloudantClientExtension(boolean useIAM)?

I think we're fine for now.

ricellis:
@emlaver I recommend replacing with mock server responses. There is no advantage to connecting to a real server from these tests (given that the VCAP is a generated test fixture). We could even add extra asserts on the recorded requests.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -83,14 +148,74 @@ public String toJson() {
...
+        }
+        // 200 with the server info
+        MockResponse serverInfoResponse = new MockResponse().setResponseCode(200)
+                .setBody("{\"couchdb\":\"Welcome\"," +
+                        "\"version\":\"2.1.1\",\"vendor\":{\"name\":\"IBM Cloudant\"," +
+                        "\"version\":\"6919\",\"variant\":\"paas\"},\"features\":[\"geo\",\"scheduler\"," +
+                        "\"iam\"]}");
+        server.enqueue(serverInfoResponse);
+
+        HttpConnection conn = client.executeRequest(Http.GET(client.getBaseUri()));
```
ricellis: I don't understand why we need `executeRequest` here, can't this just be:
```java
client.serverVersion();
```

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -83,14 +148,74 @@ public String toJson() {
...
+        } else {
+            MockWebServerResources.assertMockIamRequests(server, mockIamServer);
+        }
+
+        // assert CouchDB version
+        assertTrue(responseStr.contains("\"version\":\"2.1.1\""));
+
+        // assert same request details for serverVersion method
+        server.enqueue(serverInfoResponse);
+        client.serverVersion();
```
ricellis: Why don't we just do this just the once, instead of `executeRequest`?
Combined with my earlier comments I think this could look like from line 187:
```java
server.enqueue(serverInfoResponse);
assertEquals("2.1.1", client.serverVersion(), "The server version should be returned correctly");

if (isCookieAuth) {
  // _session request
  RecordedRequest cookieRequest = server.takeRequest();
  assertEquals("POST", cookieRequest.getMethod(), "The request method should be POST");
  assertEquals("/_session", cookieRequest.getPath(), "The request should be to the _session path");
} else {
  MockWebServerResources.assertMockIamRequests(server, mockIamServer);
}

// server info request
RecordedRequest request = server.takeRequest();
assertEquals("GET", request.getMethod(), "The request method should be GET");
assertEquals("/", request.getPath(), "The request should be for /");
```

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
```diff
@@ -83,14 +148,74 @@ public String toJson() {
...
+
+        // server info request
+        RecordedRequest request = server.takeRequest();
+        assertEquals("GET", request.getMethod(), "The request method should be GET");
+        assertEquals("/", request.getPath(), "The request should be to the _session path");
```
ricellis: The assertion message is wrong here, but I don't think this is needed anyway, see above.

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
emlaver: @ricellis `assertMockIamRequests` already asserts the `/` request so I'll move the code below `// server info request` into the if cookie auth block.

emlaver:
Should I replace [these lines in HttpIamTest](https://github.com/cloudant/java-cloudant/blob/master/cloudant-client/src/test/java/com/cloudant/tests/HttpIamTest.java#L117-L143) with `MockWebServerResources.assertMockIamRequests`?

cloudant-client/src/test/java/com/cloudant/tests/CloudFoundryServiceTest.java
ricellis: This is so much clearer and easier to follow in the new commit, thank you.

ricellis:
Re: `HttpIamTest`
Given that the assertions `MockWebServerResources.assertMockIamRequests` are available to all the tests it probably makes sense to update the `HttpIamTest` to match. Then at least if we need to change the assertions at any point we'll cover the whole set in one place.
Might be worth updating `assertMockIamRequests` to use `getBody().readUtf8()` as well at the same time.

Issue #322, opened 2016-12-06
Support new view parameters


rhyshort:
CouchDB 2.1.0 will include new view options to replace the existing `stale` parameter. These new parameters are shown in the table below with their mappings to the old API.

|# |	New API |      Old API   | Observed Behaviour (q=8,n=3)|
|--|----------|------------|--------------------------------|
|1|stable=true&update=true|Didn't exist|Blocks; Only 8 ushard indexers started|
|2|stable=true&update=false|stale=ok|Non-blocking; no indexers started; Matches previous API
|3|stable=true&update=lazy|stale=update_after|Non-blocking; 24 indexers started; Matches previous API|
|4|stable=false&update=true|stale not defined|Blocks; 24 indexers started; Matches previous API
|5|stable=false&update=false|Didn't exist|Non-blocking; no indexers started
|6|stable=false&update=lazy|Didn't exist|Non-blocking; 8 indexers started, all same node
------------

See [COUCHDB-3227](https://issues.apache.org/jira/browse/COUCHDB-3227)

sevokn:
Is implementing this on the roadmap?

If yes, when?

If not, is there any workaround way of achieving the functionality provided by the new Cloudant API (`stable` and `update` URL parameters) through the java-cloudant API?

Adding support for this should not be too hard, it's a shame that all the existing Cloudant API is not available to Java developers...

emlaver:
@sevokn We intend to address this in the next major version.
You can call `executeRequest` to use the additional view parameters.  See the [java-cloudant documentation for more details.](https://static.javadoc.io/com.cloudant/cloudant-client/2.11.0/com/cloudant/client/api/CloudantClient.html#executeRequest-com.cloudant.http.HttpConnection-)
e.g.
```
CloudantClient client = ClientBuilder.account("examples").build();
Database db = client.database("animaldb", false);
HttpConnection allDocsGet = Http.GET(new URL(db.getDBUri() + "/_all_docs?stable=true&update=true"));
System.out.println(client.executeRequest(allDocsGet).responseAsString());
```

>{"total_rows":11,"offset":0,"rows":[
{"id":"_design/views101","key":"_design/views101","value":{"rev":"1-a918dd4f11704143b535f0ab3af4bf75"}},
...
{"id":"zebra","key":"zebra","value":{"rev":"3-750dac460a6cc41e6999f8943b8e603e"}}
]}

tomblench:
Closing as PR merged

PR #448, opened 2018-05-29
Support new view parameters

tomblench:
## What

Support new view parameters, adding `stable` and `update`.

## How

Add parameters to `SettableViewParameters`, `CommonViewRequestBuilder`, `ViewQueryParameters`.

## Testing

Re-organise `ViewsTest` by moving mock-only tests to new class `ViewsMockTest`, and simplify logic of `staleParameterValues` test.

## Issues

See #322

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
@@ -53,6 +53,51 @@ public interface SettableViewParameters {
     String STALE_UPDATE_AFTER = "update_after";
 
     /**
+     * Constant for the value "false" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#stable(String)}
+     * <P>
+     * false: the view results should not be returned from a "stable" set of shards
+     * </P>
+     */
+    String STABLE_FALSE = "false";
+
+    /**
+     * Constant for the value "true" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#stable(String)}
+     * <P>
+     * true: the view results should be returned from a "stable" set of shards
+     * </P>
+     */
+    String STABLE_TRUE = "true";
+
+    /**
+     * Constant for the value "false" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#update(String)}
+     * <P>
+     * false: the view in question should not be updated prior to responding to the user
+     * </P>
+     */
+    String UPDATE_FALSE = "false";
+
+    /**
+     * Constant for the value "true" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#update(String)}
+     * <P>
+     * true: the view in question should be updated prior to responding to the user
+     * </P>
+     */
+    String UPDATE_TRUE = "true";
+
+    /**
+     * Constant for the value "lazy" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#update(String)}
+     * <P>
+     * lazy: the view in question should be updated, but after responding to the user
+     * </P>
+     */
+    String UPDATE_LAZY = "lazy";
+
+    /**
      * Setters for parameters that are common to all view requests.
      *
      * @param <K>  the type of the key emitted by the view
@@ -125,6 +170,23 @@ public interface SettableViewParameters {
         RB keys(K... keys);
 
         /**
+         * <p>
+         * Determine whether the view should be returned from a "stable" set of shards.
+         * </p>
+         * <p>
+         * See:
+         * </p>
+         * <ul>
+         * <li>{@link SettableViewParameters#STABLE_FALSE}</li>
+         * <li>{@link SettableViewParameters#STABLE_TRUE}</li>
+         * </ul>
+         *
+         * @param stable string indicating stable view behaviour
+         * @return the builder to compose additional parameters or build the request
+         */
+        RB stable(String stable);
+
+        /**
          * Allow the results from a stale view to be used. This makes the request return
          * immediately, even if the view has not been completely built yet.
          * <P>If this parameter is not given, a response is returned only after the view has been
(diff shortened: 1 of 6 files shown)


cloudant-client/src/test/java/com/cloudant/tests/ViewsMockTest.java
```diff
@@ -0,0 +1,149 @@
...
+            server.enqueue(mockResponse);
+            viewBuilder.stale(s).build().getSingleValue();
+            HttpUrl url = server.takeRequest(1, TimeUnit.SECONDS).getRequestUrl();
+            Assert.assertEquals(s, url.queryParameter("stale"));
+            Assert.assertEquals("/notarealdb/_design/testDDoc/_view/testView", url.encodedPath());
+        }
+    }
+
+    /**
+     * Similar to staleParameterValues, but for `stable` parameter
```
emlaver: I'm wondering if `staleParameterValues` and `staleParameterValues` should have a comment like `Test checks the [stable|stale] parameter in the URL by making a mock request using getSingleValue() ...`

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
```diff
@@ -52,6 +52,53 @@
      */
     String STALE_UPDATE_AFTER = "update_after";
 
+    /**
+     * Constant for the value "false" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#stable(String)}
+     * <P>
+     * false: Do not prefer view results from a 'stable' set of shards.
+     * </P>
+     */
```
emlaver: Add `@since` like in STALE_NO?

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
```diff
@@ -124,6 +171,23 @@
...
+         * </p>
+         * <ul>
+         * <li>{@link SettableViewParameters#STABLE_FALSE}</li>
+         * <li>{@link SettableViewParameters#STABLE_TRUE}</li>
+         * </ul>
+         *
+         * @param stable string indicating stable view behaviour
+         * @return the builder to compose additional parameters or build the request
+         */
+        RB stable(String stable);
```
emlaver: Add `@since` like in `RB stale(String stale);`?

ricellis: The docs say `stable` is a `boolean` I think we should change the type here to match and remove the associated constants.

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
```diff
@@ -52,6 +52,53 @@
...
+    String STABLE_TRUE = "true";
+
+    /**
+     * Constant for the value "false" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#update(String)}
+     * <P>
+     * false: Return results before updating the view.
+     * </P>
+     */
+    String UPDATE_FALSE = "false";
```
ricellis: It's not a big deal, but could use `Boolean.FALSE.toString()` or `Boolean.toString(false)` here and equivalently for the true case.

cloudant-client/src/main/java/com/cloudant/client/internal/views/CommonViewRequestBuilder.java
```diff
@@ -105,6 +105,12 @@ public RB reduce(boolean reduce) {
...
     }
 
+    @Override
+    public RB stable(String stable) {
+        viewQueryParameters.setStable(stable);
+        return returnThis();
+    }
+
     @Override
     public RB stale(String stale) {
```
ricellis: I think we could reasonably deprecate this method pointing to `stable` and `update` instead.

cloudant-client/src/test/java/com/cloudant/tests/SearchTests.java
```diff
@@ -273,5 +273,14 @@ public void escapedAmpersandTest() {
         escapingTest("class:mammal%26test%26escaping", "class:mammal&test&escaping");
     }
 
+    @Test
+    public void searchOptionsStaleTest() {
```
ricellis: I don't understand how this test relates to the view changes. Has the `stale` parameter for `_search` also changed? It didn't appear so from the docs. Is this just filling a test gap? If so perhaps it could move into another commit just for clarity?

tomblench: Thanks for spotting, this was something left over from an earlier change, removed in 49baa12

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
tomblench: done in 1dceeae

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
tomblench: Done in 1dceeae3f4050abe9c6909eb1cb4df9be959f62b

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
tomblench: Done in 1dceeae3f4050abe9c6909eb1cb4df9be959f62b

cloudant-client/src/main/java/com/cloudant/client/internal/views/CommonViewRequestBuilder.java
tomblench: Done in 1dceeae3f4050abe9c6909eb1cb4df9be959f62b

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
tomblench: Done in 8366e5e

CHANGES.md
```diff
@@ -7,6 +7,7 @@
 - [NEW] Add `ComplexKey.addHighSentinel()` to allow matching of all values as part of a complex key
   range.
 - [NEW] Support IAM authentication in replication documents.
+- [NEW] Add support `stable` and `update` parameters in views.
```
ricellis: `Add support for` or `Support...`

ricellis: Also should add a `DEPRECATED` for the `stale`

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
```diff
@@ -52,6 +52,40 @@
      */
     String STALE_UPDATE_AFTER = "update_after";
 
+    /**
+     * Constant for the value "false" for use with
+     * {@link com.cloudant.client.internal.views.CommonViewRequestBuilder#update(String)}
```
ricellis: This should link to the API method in `SettableViewParameters` I think rather than the internal impl.

tomblench: Thanks for spotting, I think they were auto-completes. Fixed in 73bc905

CHANGES.md
tomblench: Done in 54ab529

PR #473, opened 2019-01-29
Partitioned databases

smithsz:
## Checklist

## Description
Adds partitioned database support.

## Approach
Inject `_partition/<partitionKey>` into the request URI when a partition key is specified for the following operations:
- All docs.
- View queries.
- Search queries.
- Find queries.

Also, adds an additional method on the database class to fetch database partition metadata.

## Schema & API Changes

_No breaking API changes_.

New API:

- `CloudantClient.createPartitionedDB(String dbName)`
- `Database.query(String partitionKey, String query, final Class<T> classOfT)`
- `Database.search(String partitionKey, String searchIndexId)`
- `Database.partitionInfo(String partitionKey)`

## Security and Privacy
No change.

## Testing

Added additional tests:

- `PartitionedDatabaseTest.java`: Exercises the new partition querying features against a real Cloudant instance.
- `PartitionInfoMockTests.java`: Verifies mock partition metadata can be correctly deserialised.

## Monitoring and Logging
No change.

### DIFF
--- a/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015, 2018 IBM Corp. All rights reserved.
+ * Copyright © 2015, 2019 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
@@ -248,6 +248,29 @@ public class CloudantClient {
      */
     public void createDB(String dbName) {
         couchDbClient.createDB(dbName);
+    }
+
+    /**
+     * Request to create a new partitioned database with the specified name.
+     *
+     * A partitioned database introduces the ability for a user to create logical groups of
+     * documents called partitions by providing a partition key with each document. Operations on
+     * the database, specifically querying, will be extended to take partition keys as part of the
+     * request in order to constrain the operation (e.g., result set) to a specific partition.
+     *
+     * @param dbName the database name
+     * @throws com.cloudant.client.org.lightcouch.PreconditionFailedException if a database with
+     *                                                                        the same name
+     *                                                                        already exists
+     * @see <a target="_blank"
+     * href="https://console.bluemix.net/docs/services/Cloudant/api/database.html#create">
+     * Databases - create</a>
+     * @see com.cloudant.client.api.Database#query(String, String, Class)
+     * @see com.cloudant.client.api.Database#search(String, String)
+     * @see com.cloudant.client.api.views.SettableViewParameters.Common#partition(String)
+     */
+    public void createPartitionedDB(String dbName) {
+        couchDbClient.createPartitionedDB(dbName);
     }
 
     /**
--- a/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/Database.java
@@ -27,6 +27,7 @@ import com.cloudant.client.api.model.FindByIndexOptions;
 import com.cloudant.client.api.model.Index;
 import com.cloudant.client.api.model.IndexField;
 import com.cloudant.client.api.model.Params;
+import com.cloudant.client.api.model.PartitionInfo;
 import com.cloudant.client.api.model.Permissions;
 import com.cloudant.client.api.query.QueryResult;
 import com.cloudant.client.api.model.Shard;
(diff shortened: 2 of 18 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
```diff
@@ -231,6 +231,16 @@
          */
 
         RB update(String update);
+
+        /**
+         * A partition key can be specified when querying data so that results can be constrained
+         * to a specific database partition.
```
ricellis: Again we should add a comment here to make it clear this is only applicable to databases created with `partitioned:true` and link back.

cloudant-client/src/main/java/com/cloudant/client/api/CloudantClient.java
```diff
@@ -250,6 +250,26 @@ public void createDB(String dbName) {
...
+     * A partitioned database introduces the ability for a user to create logical groups of
+     * documents called partitions by providing a partition key with each document. Operations on
+     * the database, specifically querying, will be extended to take partition keys as part of the
+     * request in order to constrain the operation (e.g., result set) to a specific partition.
+     *
+     * @param dbName the database name
+     * @throws com.cloudant.client.org.lightcouch.PreconditionFailedException if a database with
+     *                                                                        the same name
+     *                                                                        already exists
+     * @see <a target="_blank"
```
ricellis: I think we should `@see ` the associated query/search/view methods that work with partitioned DBs.

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -494,14 +495,48 @@ public void createIndex(String indexDefinition) {
...
+     * @param partitionKey Database partition to query.
+     * @param query        String representation of a JSON object describing criteria used to
+     *                     select documents.
+     * @param classOfT     The class of Java objects to be returned in the {@code docs} field of
+     *                     result.
+     * @param <T>          The type of the Java object to be returned in the {@code docs} field of
+     *                     result.
+     * @return             A {@link QueryResult} object, containing the documents matching the query
+     *                     in the {@code docs} field.
+     * @see com.cloudant.client.api.Database#query(String, Class)
```
ricellis: We should make it clear this is for databases with `partitioned:true` and link back to `createPartitionedDb` or the `DbInfo` method for figuring out if something is partitioned.

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -602,7 +637,31 @@ public void deleteIndex(String indexName, String designDocId, String type) {
...
+     * 	.query("name:puffin", Bird.class);
+     * 	}
+     * </pre>
+     *
+     * @param partitionKey database partition key
+     * @param searchIndexId the design document with the name of the index to search
+     * @return Search object for searching the index
+     * @see <a
+     * href="https://console.bluemix.net/docs/services/Cloudant/api/search.html#search"
+     * target="_blank">Search</a>
```
ricellis: Again we should make it clear this is for `partitioned:true` DBs and link back.

cloudant-client/src/main/java/com/cloudant/client/api/Database.java
```diff
@@ -1334,6 +1393,21 @@ public DbInfo info() {
                 DbInfo.class);
     }
 
+    /**
+     * Get information about this database partition.
```
ricellis: I think this should say `Get information about a partition in this database` since you have to instruct the method which partition and also `this` has meaning in Java that could be misconstrued otherwise.

cloudant-client/src/test/java/com/cloudant/tests/PartitionedDatabaseTest.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@RequiresCloudant
+public class PartitionedDatabaseTest extends TestWithDbPerClass {
+
+    private static final String testDbName = "partitioned_database_" + Utils.generateUUID();
```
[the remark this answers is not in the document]
smithsz: I can remove this now and use the database extension instead.

cloudant-client/src/main/java/com/cloudant/client/api/views/SettableViewParameters.java
smithsz: Is this okay?
```
        /**
         * A partition key can be specified when querying data so that results can be constrained
         * to a specific database partition.
         *
         * Only available in partitioned databases. To ensure a database is partitioned call
         * {@link Database#info()} and check that {@link DbInfo.Props#getPartitioned()} returns
         * {@code true}.
         *
...
```
I'll use the same snippet for search and query too.

ricellis: Sounds good, but I would change `ensure` to  either `check` or `verify`, since the only way to ensure it is partitioned is to create it that way.

PR #495, opened 2019-10-07
Add total row count getter to AllDocsResponse

markewallace:
## Checklist

## Description
I need to be able to make paginated requests to the primary index. This can be done using the `skip` and `limit` parameters and checking the `total_rows` property from the response. However the Java API states `Do not use skip/limit for pagination, instead use ViewRequestBuilder.newPaginatedRequest(Key.Type, Class) .` but it doesn't seem to be possible to use a `ViewRequestBuilder` with the primary index.
I expected the Java API would be a Java wrapper around the underlying REST API and should expose the same data, so I think it is correct to make the `total_rows` property available in the `AllDocsResponse` interface.

## Approach
The total row count is available in the REST API response but the Java API doesn't provide access. This PR adds total row count getter to `AllDocsResponse` interface and implements this new method. 

## Schema & API Changes
Add total row count getter to AllDocsResponse

## Security and Privacy
No change

## Testing
Added new assert

## Monitoring and Logging
No change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [NEW] Add getter for total row count to AllDocsResponse
+
 # 2.18.0 (2019-10-02)
 - [FIXED] Create an array of strings for QueryBuilder.fields() when a single field is provided.
 - [FIXED] Potential NPE creating exception messages.
--- a/cloudant-client/src/main/java/com/cloudant/client/api/views/AllDocsResponse.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/views/AllDocsResponse.java
@@ -93,4 +93,9 @@ public interface AllDocsResponse {
      */
     Map<String, String> getErrors();
 
+    /**
+     * @return the total number of rows in the _all_docs view, before any filtering has taken place
+     * @since 2.19.0
+     */
+    Long getTotalRowCount();
 }
--- a/cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
+++ b/cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
@@ -732,6 +732,7 @@ public class ViewsTest extends TestWithDbPerTest {
         assertThat(response.getDocs().get(1).isDeleted(), is(true));
         assertThat(response.getDocs().get(1).getId(), is(response2.getId()));
         assertThat(response.getDocs().get(1).getRevision(), is(removeResponse.getRev()));
+        assertThat(response.getTotalRowCount(), is(8L));
     }
 
     // all docs with given keys, one document deleted:
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/views/AllDocsRequestResponse.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/views/AllDocsRequestResponse.java
@@ -82,6 +82,11 @@ public class AllDocsRequestResponse implements AllDocsRequest, AllDocsResponse {
         return errors;
     }
 
+    @Override
+    public Long getTotalRowCount() {
+        return response.getTotalRowCount();
+    }
+
     /*
          * Object representation of rev field from a JSON object.
          * <P>


cloudant-client/src/test/java/com/cloudant/tests/ViewsTest.java
```diff
@@ -17,6 +17,7 @@
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThan;
```
bessbd: This import seems unused. Can you please remove it?

cloudant-client/src/main/java/com/cloudant/client/api/views/AllDocsResponse.java
```diff
@@ -93,4 +93,9 @@
      */
     Map<String, String> getErrors();
 
+    /**
+     * @return the total number of rows returned by the _all_docs request.
```
ricellis: As per https://github.com/cloudant/java-cloudant/blob/bdd6d97ab64956b30a939a17eb9359a7734ec8b8/cloudant-client/src/main/java/com/cloudant/client/api/views/ViewResponse.java#L204

>the total number of rows in the view, before any filtering has taken place

Basically `total_rows` is the number of rows that are indexed by the view, not the number of rows returned in the response.

So for `_all_docs` `total_rows` is basically equivalent to the doc count of the database, which you can already get from `db.info().getDocCount()`.

For example for a db with 12 docs if I do _all_docs?key="mydoc" the response has `total_rows: 12` **not** `total_rows: 1` even though only "mydoc" will be in the response _i.e._
`db.getAllDocsRequestBuilder().keys("mydoc").build().getResponse().getDocs().size()` -> `1`
and:
`db.getAllDocsRequestBuilder().keys("mydoc").build().getResponse().getTotalRowCount()` -> `12`

markewallace:
Hi @bessbd 

I think primitives would have been better in `ViewResponse`. The problem with making the change now is that it will be a breaking change e.g. 
```
/Users/mwallace/Talent/git/java-cloudant/cloudant-client/src/test/java/com/cloudant/tests/util/CheckPagination.java:172: error: long cannot be dereferenced
        assertThat(page.getFirstRowCount().intValue(), is(startIndex));
                                          ^
/Users/mwallace/Talent/git/java-cloudant/cloudant-client/src/test/java/com/cloudant/tests/util/CheckPagination.java:173: error: long cannot be dereferenced
        assertThat(page.getLastRowCount().intValue(), is(Math.min(startIndex + docsPerPage - 1,
                                         ^
/Users/mwallace/Talent/git/java-cloudant/cloudant-client/src/test/java/com/cloudant/tests/util/CheckPagination.java:175: error: long cannot be dereferenced
        assertThat(page.getPageNumber().intValue(), is(pageNumber));

```

I can make this change in `AllDocsResponse`

bessbd:
@markewallace : I wasn't aware this would be a breaking change. In that case I'm not sure anymore it's worth doing an extra unboxing. Thank you for looking into that!

cloudant-client/src/main/java/com/cloudant/client/api/views/AllDocsResponse.java
markewallace: Hi @ricellis 

I would like to avoid making an additional API call `db.info().getDocCount()` to get this data and the data is already available so it's convenient to include it in the `AllDocsResponse` interface. It also aligns the Java API with the underlying REST API.

ricellis: that's fine :)
`total_rows` can be  a misleading concept so I wanted to be sure you knew what you were actually going to get from that value.

And on that note we'll need this comment to be updated so as to not mislead other users on what this returns. I'd be happy for it to say the same thing as the view i.e. `the total number of rows in the view, before any filtering has taken place`, but something more all_docs specific would also be OK if you can think of something better.

markewallace: Updated the comment to be `the total number of rows in the _all_docs view, before any filtering has taken place`

markewallace: @ricellis once this PR get's merged how soon will a new release be created and available for use?

ricellis:
>I expected the Java API would be a Java wrapper around the underlying REST API and should expose the same data, so I think it is correct to make the total_rows property available in the AllDocsResponse interface.

I think that is fair and why I don't oppose this change in general. However, I think it is worth addressing a few points about pagination here to ensure you get the best results.

>This can be done using the skip and limit parameters and checking the total_rows property from the response.

Firstly, a word of caution that `total_rows` is subject to change between pages if documents are being added or removed in the database. Secondly, `total_rows` in general is not required for pagination (with `skip`/`limit` or otherwise) - you can terminate the pagination when the returned results are smaller than the `pageSize` e.g.

```java
int pageSize = 100;
AllDocsResponse page;
List<String> docIds;
int skip = 0;
do {
    page = db.getAllDocsRequestBuilder().limit(pageSize).skip(skip).build().getResponse();
    docIds = page.getDocIds();
    if (docIds.isEmpty()) {
        //  Zero size page in the case where n docs is exactly divisible by page size 
        break;
    } else {
        // Increment the skip
        skip+=docIds.size();
    }
    //do something with the page
    System.out.println(page.getDocIds());
} while (docIds.size() == pageSize);
```

We don't recommend using `skip`/`limit` for pagination because the performance is not as good as using a `startkey` and it is possible to get duplicate results in the case where a document is inserted into the DB "behind" the current page. In that case `skip` for the next page will return a document again because there is now one more document to skip to reach that point in the index. You can see more details about this in the [CouchDB pagination recipe](https://docs.couchdb.org/en/stable/ddocs/views/pagination.html#paging-alternate-method). Note that key based pagination is still faster thank skip/limit for larger databases even in recent versions of CouchDB.

>However the Java API states `Do not use skip/limit for pagination, instead use ViewRequestBuilder.newPaginatedRequest(Key.Type, Class)` . but it doesn't seem to be possible to use a ViewRequestBuilder with the primary index.

Correct, the reasons for not recommending `skip`/`limit` are outlined above.
The pagination for `_all_docs` using a `startKey` is actually simpler than for the general view case, because the key is guaranteed to be unique (as it is a doc id). As such one example of how you can paginate is like this:

```java
String UNICODE_START = "/u0000";
String startKey = UNICODE_START;
int pageSize = 100;
AllDocsResponse page;
List<String> docIds;
do {
    page = db.getAllDocsRequestBuilder().limit(pageSize).startKey(startKey).build().getResponse();
    docIds = page.getDocIds();
    if (docIds.isEmpty()) {
        //  Zero size page in the case where n docs is exactly divisible by page size 
        break;
    } else {
        // Get last docId and add unicode start for next page startKey
        startKey = docIds.get(docIds.size()-1) + UNICODE_START;
    }
    //do something with the page
    System.out.println(page.getDocIds());
} while (docIds.size() == pageSize);
```

All that said we can add the total rows getter for completeness sake, it's just probably not the best thing to use for pagination.

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# 2.19.0 (2019-10-07)
+- [IMPROVED] Add getter for total row count to AllDocsResponse
```
ricellis: Please change to
```
# Unreleased
- [NEW] Add getter for total row count to AllDocsResponse
```

We'll fill in the date and version when we cut the release and I think this should be a `NEW` because it is adding API. Thanks!

markewallace: @ricellis I've pushed that changes. Thank you for the explanation of the correct pattern to use for pagination, this topic was causing some confusion for us.

PR #516, opened 2020-09-11
Allow query operators to be used with ONE query expression element

### LINKED ISSUE #515: When using QueryBuilder, the and Operation sometimes doesn't emit an array
## Bug Description
When I instantiate a QueryBuilder with an and Operation containing only a single Selector, the resulting JSON does not have an array for the $and field inside the selector.

I'm an IBM Cloudant customer.

### 1. Steps to reproduce and the simplest code sample possible to demonstrate the issue

1. Create a QueryBuilder: `QueryBuilder builder = new QueryBuilder(and(eq("userId", "myUser")));`
2. Execute the query against an existing database: `database.query(builder.build(), JsonObject.class);`

Alternatively, inspect the `selector` variable on the QueryBuilder after step 1 to see the incorrect JSON. For example, `"$and": {"userId": {"$eq": "myUser"}}`

### 2. What you expected to happen
I expect QueryBuilder to emit JSON which is accepted by CouchDB, meaning $and should always be an array regardless of how many elements are in it.

### 3. What actually happened
CouchDB gives me a 400 error from the _find endpoint because the $and is malformed.

## Environment details
- Version(s) that are affected by this issue.
    > 2.18.0 and 2.19.1 are the only versions I checked, both are affected.
- Java version including vendor and platform
    > OpenJDK 11.0.4 macOS
- Using the optional okhttp dependency.
    > No

I'm executing the query against CouchDB 3 in my local dev environment.

mojito317:
## Checklist

## Description
Fixes #515

## Approach

I passed the operator to the `quoteCurly` method and used the array for all the operators except `not`.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - `basicSelector2WithOneField` in `QueryTests` for an `$and` operator passed with only one `$eq` expression element

## Monitoring and Logging
- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# Unreleased
+- [FIXED] Allow $and, $or, and $nor operator selectors
+  to be used with only one expression selector
+
 # 2.19.1 (2020-07-03)
 - [FIXED] Connection leak regression introduced in 2.18.0 caused by not closing streams from
   successful session response bodies.
@@ -296,7 +300,7 @@
 - [FIXED] `NullPointerException` when parsing `{doc: null}` JSON in search or view results.
 - [FIXED] Fixed issue with pagination numbering when using `queryPage` with
   a clustered DB.
-- [FIXED] Fixed issue where `queryPage` could not handle JSON values emitted from views.  
+- [FIXED] Fixed issue where `queryPage` could not handle JSON values emitted from views.
 - [IMPROVED] Various documentation updates.
 - [DEPRECATED] `com.cloudant.client.api.model.Page` setter methods.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Unreleased
-- [FIXED] Allow $and, $or, and $nor operator selectors
-  to be used with only one expression selector
+- [FIXED] Allow `$and`, `$or`, and `$nor` operator selectors
+  to be used with only one expression selector.
 
 # 2.19.1 (2020-07-03)
 - [FIXED] Connection leak regression introduced in 2.18.0 caused by not closing streams from
--- a/cloudant-client/src/main/java/com/cloudant/client/internal/query/Helpers.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/internal/query/Helpers.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017, 2018, 2020 IBM Corp. All rights reserved.
+ * Copyright © 2017, 2020 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


cloudant-client/src/test/java/com/cloudant/tests/QueryTests.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017, 2018 IBM Corp. All rights reserved.
+ * Copyright © 2017, 2018, 2020 IBM Corp. All rights reserved.
```
bessbd, line 2: Nit: I don't think the "middle" year is necessary.
```suggestion
 * Copyright © 2017, 2020 IBM Corp. All rights reserved.
```

cloudant-client/src/main/java/com/cloudant/client/internal/query/Helpers.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017, 2018 IBM Corp. All rights reserved.
+ * Copyright © 2017, 2018, 2020 IBM Corp. All rights reserved.
```
bessbd, line 2: Nit: 
```suggestion
 * Copyright © 2017, 2020 IBM Corp. All rights reserved.
```

cloudant-client/src/main/java/com/cloudant/client/internal/query/Helpers.java
```diff
@@ -59,8 +59,8 @@ public static String quoteNoSquare(Object[] os) {
         return quoteInternal(os, ", ", "", "", "", "");
     }
 
-    public static String quoteCurly(Object[] os) {
-        if (os.length == 1) {
+    public static String quoteCurly(Object[] os, String op) {
```
bessbd, line 62: As this is a public method, do you think you could add re-add the original `public static String quoteCurly(Object[] os)` so that we don't have to do a major release for this change? (I understand that it says `internal` in the path, but I wonder if there's a way to retain backwards compatibility.)

bessbd (commented):
LGTM in general. Left a couple of questions / nits. Let me know what you think, @mojito317 .

cloudant-client/src/main/java/com/cloudant/client/internal/query/Helpers.java
mojito317: I reverted the original method: https://github.com/cloudant/java-cloudant/pull/516/files#diff-b2d6eda22288fdd2ad1e1819d87181ecR70-R76

bessbd (approved):

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+- [FIXED] Allow $and, $or, and $nor operator selectors
+  to be used with only one expression selector
```
ricellis, line 3: nit: should finish with a `.`

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+- [FIXED] Allow $and, $or, and $nor operator selectors
```
ricellis, line 2: nit: might be nice to backtick the operators for markdown

cloudant-client/src/main/java/com/cloudant/client/internal/query/Helpers.java
ricellis: IMO this was unnecessary it is not API and so would not have required a major release, but I can understand treading carefully and I'm not strongly opposed to the duplication.

ricellis (approved):
+1
Feel free to address (or not) the nits in the `CHANGES.md`.

CHANGES.md
mojito317: Done in 500a660.

CHANGES.md
mojito317: Done in 500a660.

mojito317:
Thanks for the review @bessbd and @ricellis! I'm about to merge this.

PR #527, opened 2021-07-05
Type of `sinceSeq` can be also a String besides an Integer

### LINKED ISSUE #526: SinceSeq should be a string
## Bug Description

The Replicator sinceSeq() call takes an Integer but the changes feed needs the entire sequence as a String to function properly.

### 1. Steps to reproduce and the simplest code sample possible to demonstrate the issue
```
Replicator rep = cloudantClient.replicator()
                    .target(targetUrl + targetDBName)
                    .source(sourceUrl + sourceDBName)
                    .sinceSeq(123)
                    .save();
```

### 2. What you expected to happen
The replicator will start at the sequence provided and then continue to the end of the changes feed.

### 3. What actually happened
The changes reader died.

## Environment details
The current version is affected based on the javadocs

This can be remedied by changing the sinceSeq() to use a String instead of an integer and updating the applicable javadocs to explain what needs to be passed.

This was confirmed by manually changing the `since_seq` option in the replicator document to the entire String of the wanted start point.

mojito317:
## Checklist

## Description

Fixes #526

## Approach

`seqs` could be `Integer`s in CouchDB 1.x and `String`s in 2.x, so I changed all the places where they remained `Integer`s to `JsonElement`. With this, both `Integer` and `String` can be used.

## Security and Privacy

- No change

## Testing

- This behavior can be tested with ITs well, so I just extended ITs with a plus `.sinceSeq(...)` call which is going to use a `seq` from a `{source_db}/_changes` request.

## Monitoring and Logging

- No change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # Unreleased
 - [DEPRECATED] This library is now deprecated and will be EOL on Dec 31 2021.
+- [FIXED] Type of `sinceSeq` can be also a String besides an Integer.
 
 # 2.19.2 (2021-04-07)
 - [NEW] Add migration guide to the newly supported cloudant-java-sdk (coordinates: com.ibm.cloud:cloudant).
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Unreleased
 - [DEPRECATED] This library is now deprecated and will be EOL on Dec 31 2021.
-- [FIXED] Type of `sinceSeq` can be also a String besides an Integer.
+- [FIXED] Type of `sinceSeq` can be also a `String` besides an `Integer`.
 
 # 2.19.2 (2021-04-07)
 - [NEW] Add migration guide to the newly supported cloudant-java-sdk (coordinates: com.ibm.cloud:cloudant).
--- a/cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 IBM Corp. All rights reserved.
+ * Copyright (c) 2015, 2021 IBM Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
--- a/cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
+++ b/cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
@@ -16,6 +16,7 @@ package com.cloudant.client.api.model;
 
 import com.cloudant.client.org.lightcouch.Attachment;
 import com.cloudant.client.org.lightcouch.Replicator;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 import java.util.HashMap;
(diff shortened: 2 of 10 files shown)


cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
```diff
@@ -227,12 +229,16 @@ public void setRetriesPerRequest(Integer retriesPerRequest) {
         replicatorDocument.setRetriesPerRequest(retriesPerRequest);
     }
 
-    public Integer getSinceSeq() {
+    public JsonElement getSinceSeq() {
```
ricellis, line 232: To avoid breaking compatibility with anyone using this still with e.g. Couch 1.x we should not change this API.
I suggest reverting this and adding:
`String getSinceSeqString()` - which can output the JsonElement string (for both string and number seq types), the existing int method can be modified to check if the JsonElement is a number primitive and return the int if it is and throw (I guess an `IllegalStateException`) if it isn't.

ricellis (changes requested):
Nearly there, as discussed offline I think the seq things should preferably be in their own tests.
Otherwise 2 other things:
- don't change the existing API in the ReplicatorDocument model class
- changes entry

cloudant-client/src/main/java/com/cloudant/client/api/model/ReplicatorDocument.java
mojito317: Fixed in 103eb3b and 0c4e3c8

mojito317:
> Nearly there, as discussed offline I think the seq things should preferably be in their own tests.

Done in 79abd33. Lmk if you would like to see more.

> don't change the existing API in the ReplicatorDocument model class

Done in 103eb3b and 0c4e3c8.

> * changes entry

Done in 64770f4.

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 - [DEPRECATED] This library is now deprecated and will be EOL on Dec 31 2021.
+- [FIXED] Type of `sinceSeq` can be also a String besides an Integer.
```
ricellis, line 3: nit formatting
```suggestion
- [FIXED] Type of `sinceSeq` can be also a `String` besides an `Integer`.
```

cloudant-client/src/test/java/com/cloudant/tests/ReplicationTest.java
```diff
@@ -106,4 +106,39 @@ public void replication_conflict() throws Exception {
         //a conflict
         assertConflictsNotZero(db2);
     }
+
```
ricellis, line 109: nit: copyright date

cloudant-client/src/test/java/com/cloudant/tests/ReplicatorTest.java
```diff
@@ -131,4 +131,41 @@ public void replication_conflict() throws Exception {
         //a conflict
         assertConflictsNotZero(db2);
     }
+
```
ricellis, line 134: nit: copyright date

ricellis (approved):
+1 just a couple nits

mojito317:
> +1 just a couple nits

Thanks for the thorough review! I fixed the nits in f3018dd.

emlaver (approved):
Looks good!  We should consider adding a negative test to assert that an error message is thrown when passing in an unexpected type.

mojito317:
> We should consider adding a negative test to assert that an error message is thrown when passing in an unexpected type.

An unexpected sinceSeq (e.g. a `"true"` or a `"asd"`) will cause a crashing state in the replication and the replication will fail and timeout eventually in the `waitForReplicatorToReachStatus` method:
https://github.com/cloudant/java-cloudant/blob/2c53b88a84906a3a47f844ce7c25078e11baee46/cloudant-client/src/test/java/com/cloudant/tests/util/Utils.java#L142

Meanwhile, the replication document will disappear from the `_replicator` database.

Other types cannot be added, because compilation would fail.

mojito317:
Thanks for the reviews @ricellis and @emlaver! I'm about to merge this!

